PR #211, opened 2020-11-23
feat: new repositories install hooks

svatwork:
As described in https://typicode.github.io/husky/#/?id=install, this script will make sure that developers will also receive the necessary hooks for having features like commit linting.

### DIFF
--- a/package.json
+++ b/package.json
@@ -59,6 +59,7 @@
     "husky:pre-commit": "lint-staged && npm run spellcheck",
     "lint": "bin/otomi validate-templates",
     "lint:all": "run-p spellcheck lint",
+    "postinstall": "husky install",
     "release": "standard-version",
     "release:github": "github-release-from-changelog",
     "spellcheck": "cspell \"**/*\" \".**/**\"",


package.json
```diff
@@ -59,6 +59,7 @@
     "husky:pre-commit": "lint-staged && npm run spellcheck",
     "lint": "bin/otomi validate-templates",
     "lint:all": "run-p spellcheck lint",
+    "postinstall": "husky install",
```
Morriz, line 62: I don't think this has any effect after having done "npm install", as that registers the hooks already.

Morriz (commented):
I don't think this has any effect

Morriz:
@svatwork can you keep working on this? I see your suggestions are based on the new v5, where they don't use auto-install We are using v4 configuration (and install I presume). I wouldn't mind if this PR is used to upgrade to v5 and add the exclusion logic to NOT install the hooks in CI.

svatwork:
Yes, sure.

Morriz:
I just tried a fresh npm install husky and see it installs @v4.3.0. So lets forget it for now.

PR #213, opened 2020-11-24
Bug/jq yq not found in first stage ci

svatwork:
There were a couple of warnings and errors in all the CI builds of this repository. Specifically:
- `node-gyp` wouldn't build because it interfered with an dependency of the old repository of `jq`. This has been replaced with `node-jq`, and the reliance on `contextify` 0.1.15 is fixed and no longer present.
- `husky install` wouldn't work if Git isn't installed on the host.

This pull request does increase the image size but now the build will pass. It should be reasonably sized with the `--no-install-recommends` flag of `apt-get`.

_Something to open the discussion on still is whether it's useful to skip CI's (the build should pass anyway) and the || true at the end of the base's layer RUN command in the Dockerfile in the root of the repository._

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ ARG SKIP_TESTS='false'
 ENV EXIT_FAST='true'
 
 # Necessary for node-gyp and some node modules 
-RUN apt-get update && apt-get install --no-install-recommends autoconf libtool build-essential git python -y
+RUN apt-get update && apt-get install --no-install-recommends automake libtool build-essential git python -y
 
 ENV APP_HOME=/home/app/stack
 RUN mkdir -p $APP_HOME
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,9 +1,9 @@
-FROM node:14-slim as build
+FROM node:14-slim as build 
 
 ARG SKIP_TESTS='false'
 ENV EXIT_FAST='true'
 
-RUN apt-get update && apt-get install autoconf build-essential python -y
+RUN apt-get update && apt-get install --no-install-recommends autoconf build-essential python -y
 
 ENV APP_HOME=/home/app/stack
 RUN mkdir -p $APP_HOME
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,7 +3,8 @@ FROM node:14-slim as build
 ARG SKIP_TESTS='false'
 ENV EXIT_FAST='true'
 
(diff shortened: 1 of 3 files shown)


svatwork:
Still some issues. I can't reproduce them on a MacOS host either so instead of debugging it with the web interface of Github Actions I should reproduce these errors on an Ubuntu host.

package.json
```diff
@@ -8,21 +8,22 @@
...
     "@hkdobrev/run-if-changed": "^0.3.1",
     "commitizen": "^4.2.2",
-    "cspell": "^4.1.2",
+    "cspell": "^4.2.2",
     "cz-conventional-changelog": "^3.3.0",
     "github-release-from-changelog": "^2.1.1",
     "husky": "^4.3.0",
-    "lint-staged": "^10.5.0",
+    "lint-staged": "^10.5.2",
+    "node-jq": "^1.11.3",
```
Morriz, line 22: where is node-jq used? I think we only use jq in bash

package.json
```diff
@@ -8,21 +8,22 @@
...
     "npm-run-all": "^4.1.5",
-    "prettier": "^2.1.2",
-    "standard-version": "^9.0.0"
+    "prettier": "^2.2.0",
+    "standard-version": "^9.0.0",
+    "yq": "0.0.1"
```
Morriz, line 26: why is this added as node package? we only use this in bash I believe

Morriz (changes requested):
I think the addition of these 2 node packages can be reverted as they are used nowhere...we can keep the updated packages ;)

Dockerfile
```diff
@@ -1,8 +1,11 @@
-FROM node:14-slim as ci
+FROM node:14-slim as build 
 
 ARG SKIP_TESTS='false'
 ENV EXIT_FAST='true'
 
+# Necessary for node-gyp and some node modules 
+RUN apt-get update && apt-get install --no-install-recommends automake libtool build-essential git python -y
+
```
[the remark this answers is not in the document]
svatwork: OK, I will try to find another image but I didn't find any node images that meet our needs if we want to execute npm install in this layer. From what I gather it's meant to be extended and in this case npm install requires node-gyp and node-gyp needs these tools.

package.json
svatwork: You're right, it should just be available on the host. My bad.

package.json
svatwork: I'll remove it.

Morriz:
Closing as latest master has this rework

PR #225, opened 2020-12-01
feat: support for aws mfa exec

svatwork:
I'm getting a correct `otomi diff` now:

```
 otomi diff
Info: The 'secrets.*.yaml files' are not decrypted, because /Users/sebastiaanverbeek/Projects/redkubes/test/otomi-values/.sops.yaml file is not present
Comparing release=jobs-artifacts, chart=../charts/raw
Comparing release=knative-init, chart=/tmp/charts/knative-init-0.18
Comparing release=loki-artifacts, chart=../charts/raw
monitoring, reverse-proxy-auth-config, Secret (v1) has changed:
  # Source: raw/templates/resources.yaml
  apiVersion: v1
  kind: Secret
  metadata:
    labels:
      app: loki
      app.kubernetes.io/instance: loki-artifacts
```

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -36,6 +36,7 @@ RUN add-apt-repository ppa:rmescandon/yq && apt-get update -qq \
   groff \
   jq \
   yq \
+  awscli \
   && rm -rf /var/lib/apt/lists/*
 
 RUN mkdir -p /home/app
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,23 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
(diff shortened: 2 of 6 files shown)


tools/Dockerfile
```diff
@@ -36,6 +36,7 @@ RUN add-apt-repository ppa:rmescandon/yq && apt-get update -qq \
   groff \
   jq \
   yq \
+  awscli \
```
Morriz, line 39: Is this not included in the aws-authenticator then? I would like to keep our tools image lean and avoid cloud specific tooling as much as possible. We just removed all of that stuff.

Morriz (commented):
I would suggest you install awscli locally as it is not needed as a dep to deploy. The tools image only needs versioned artifacts to talk to the components on the cluster. Credential rotation tooling is not amongst them.

Please close this PR

Morriz (approved):

Morriz:
@svatwork otomi/tools:1.4.10 is not pushed yet. Please do and retrigger this last action

Issue #204, opened 2020-11-16
K8s Jobs for teams


Morriz:
Link to discussion: https://github.com/redkubes/otomi-core/discussions/295

# Tasks 

### Meta-schema proof of concept

- [x] Add proof of concept functionality to `values-schema.yaml` for Kubernetes Jobs.
- [x] Add proof of concept `jobs.demo.yaml`.
- [x] Modified requirements based on initial review.

### Previous configuration refactoring

- [x] Delete previous configuration that allows to specify a job for development reasons. 
- [x] Refactor Job/CronJob chart to enable the `spec.enabled` flag by default.
- [x] Refactor Job/CronJob chart to enable to specify an array of Jobs.

### Team jobs basic functionality

- [x] Teams can enter jobs in `jobs.demo.yaml`.
- [x] Add functionality to Helmfile (helmfile-60.teams.yaml).

### Extra functionality for Jobs (e.g. when the Jobs spec changes -- should be exposed to the end-user)

~~- [ ] Explore Helm hooks.~~
- [x] Explore Kubernetes Job specification (preferred over Helm hooks).
- [x] Add a proof of concept to `values-schema.yaml` for review.
- [x] Add functionality to the relevant Helm Charts.
- [x] Proposal for running a running a job `always` and `onChange` so the default behaviour that a job is always created can be overridden.

### Missing functionality, see #295 

- [x] jobs.demo.yaml
  - [x] Job Spec
    - [x] .init.env
      - [x] values-schema.yaml
      - [x] Job Chart wiring 
    - [x] .init.image
       - [x] values-schema.yaml
       - [x] Job Chart wiring 
     - [x] .init.secretRefs: need to reference secret 
       - [x] values-schema.yaml
       - [x] Job Chart wiring 
     - [x] .secretRefs
       - [x] values-schema.yaml
       - [x] Job Chart wiring 

# Definition Of Done
- [x] Tasks are done
- [x] (Unit) Tests are performed
- [ ] (Architecture) Design Record(s) have been added as `adr/*.md` and appended to list in `adr/_index.md`
- [x] Specs and demo files have been updated to reflect code changes
- [ ] Documentation has been updated
- [x] Functionality, code, and/or documentation has been peer reviewed
- [x] Relevant team members have been notified

svatwork:
I'm awaiting responses to the discussions in #239 and the review of #258 @Morriz @j-zimnowoda.

svatwork:
[...]

PR #239, opened 2020-12-10
Task #204: team jobs mechanism, meta-schema validation

svatwork:
Hello, this is my concept version. @j-zimnowoda suggested to me to first create new entries in `values-schema.yaml` so I can propose it. 

I'm new to schema validation and all the associated technologies so please go easy on me. 

I also started a [discussion on meta-schema validation](https://github.com/redkubes/otomi-core/discussions/235), which I hope to incorporate into this addition. 

## Explanations

- I think `'key': 'value'`should be `key: value` in YAML, is this correct? So to me `'$key':'$value'` doesn't make a lot of sense, it would seem this is evaluated `"'$key':'$value'"` in JSON, correct? 
- According to the JSON specification additionalProperties is evaluated to an empty array, so I don't really understand why the value for this key should be `false`

## Next

Since I just would like some _review on the meta schema_, please be assured that I will continue to work on the checklist mentioned in #204.

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -24,8 +24,7 @@
     "CONTRIBUTING": "markdown"
   },
   "yaml.schemas": {
-    // "http://json-schema.org/draft/2019-09/schema#": "./values-schema.yaml",
-    // "http://json-schema.org/draft/2019-09/schema#": ".vscode/values-schema.yaml",
+    "http://json-schema.org/draft-07/schema#": "values-schema.yaml",
     ".values/values-schema.yaml": ".demo/env/*.yaml"
   },
   "shellformat.flag": "-i 2 -ci"
--- a/charts/jobs/templates/cronjob.yaml
+++ b/charts/jobs/templates/cronjob.yaml
@@ -1,6 +1,6 @@
 {{- $v := .Values }}
 {{- range $name, $spec := $v.tasks }}
-{{- if and $spec.enabled (eq $spec.type "cronjob") }}
+{{- if and $spec.enabled (eq $spec.jobType "cronjob") }}
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -220,9 +220,6 @@ definitions:
     type: object
   job:
     additionalProperties: false
-    description: |
-      You may pass a Job object. You must pass the following properties:
-
     properties:
       enabled:
         type: boolean
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -1525,7 +1525,7 @@ properties:
         patternProperties:
           ^[a-z-]+$:
             $ref: '#/definitions/team'
-        type: object
+        type: array
     type: object
   toolsVersion:
     type: string
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -1,5 +1,5 @@
 ---
-'$schema': 'http://json-schema.org/draft-07/schema'
+$schema: http://json-schema.org/draft-07/schema#
 type: object
 additionalProperties: false
 definitions:
@@ -218,6 +218,40 @@ definitions:
         type: string
     title: Container image
     type: object
+  jobs:
+    description: |
+      You may pass an array of Job objects. The Kubernetes API accepts a Job specification, 
+      which is meant for run-to-completion apps (e.g. batch Jobs).
+    items:
+      description: |
+        You may pass a Job object. You must pass the following properties:
+
+          enabled: (type boolean)
+          env: (type array of strings, pattern-matched)
+          image: (type string, pattern-matched)
+          init: (type object)
+            image: (type string, pattern-matched)
+            script: (type string)
+      properties:
+        enabled:
+          type: boolean
+        env:
+          items:
+            $ref: '#/definitions/env'
+          type: array
+        image:
+          $ref: '#/definitions/image'
+        init:
+          properties:
+            image:
+              $ref: '#/definitions/image'
+            script:
+              description: |
+                You must pass a non-empty string containing an executable script. This script is executed first.
+          type: object
+      type: object
+    minItems: 0
+    type: array
   k8sVersion:
     default: '1.18'
     description: The cluster k8s version. Otomi supports 2 minor versions backwards compatibility from the suggested default.
@@ -377,6 +411,10 @@ definitions:
       id:
         $ref: '#/definitions/idName'
         description: Must be the same as the name.
+      jobs:
+        items:
+          $ref: '#/definitions/jobs'
+        type: array
       oidc:
         additionalProperties: false
         properties:
@@ -602,14 +640,13 @@ properties:
     additionalProperties: false
     properties:
       jobs:
-        type: object
-        additionalProperties: false
+        $ref: '#/definitions/jobs'
         patternProperties:
-          'certs-aws|harbor|ingress-azure|keycloak':
+          certs-aws|harbor|ingress-azure|keycloak:
             type: object
             properties:
               image:
-                '$ref': '#/definitions/image'
+                $ref: '#/definitions/image'
       cert-manager:
         type: object
         additionalProperties: false
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -221,18 +221,8 @@ definitions:
   job:
     additionalProperties: false
     description: |
-      You may pass an array of Job objects. The Kubernetes API accepts a Job specification, 
-      which is meant for run-to-completion apps (e.g. batch Jobs).
-      description: |
-        You may pass a Job object. You must pass the following properties:
+      You may pass a Job object. You must pass the following properties:
 
-          enabled: (type boolean)
-          env: (type strings[], pattern-matched)
-          image: (type string, pattern-matched)
-          init: (type object)
-            image: (type string, pattern-matched)
-            script: (type string)
-            secret: (type string[])
     properties:
       enabled:
         type: boolean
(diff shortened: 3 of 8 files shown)


values-schema.yaml
```diff
@@ -218,6 +218,40 @@ definitions:
         type: string
     title: Container image
     type: object
+  jobs:
```
j-zimnowoda, line 221: Could you define just a schema for a single job?

values-schema.yaml
```diff
@@ -377,6 +411,10 @@ definitions:
       id:
         $ref: '#/definitions/idName'
         description: Must be the same as the name.
+      jobs:
```
j-zimnowoda, line 414: Right now  '#/definitions/jobs` is already an array. But If you make 
a definition for single job the this statement should be fine.

values-schema.yaml
```diff
@@ -218,6 +218,40 @@ definitions:
...
+      description: |
+        You may pass a Job object. You must pass the following properties:
+
+          enabled: (type boolean)
+          env: (type array of strings, pattern-matched)
+          image: (type string, pattern-matched)
+          init: (type object)
+            image: (type string, pattern-matched)
+            script: (type string)
+      properties:
```
j-zimnowoda, line 235: Please add `secret` and `script` properties.

values-schema.yaml
```diff
@@ -218,6 +218,40 @@ definitions:
...
     type: object
+  jobs:
+    description: |
+      You may pass an array of Job objects. The Kubernetes API accepts a Job specification, 
+      which is meant for run-to-completion apps (e.g. batch Jobs).
+    items:
+      description: |
+        You may pass a Job object. You must pass the following properties:
+
+          enabled: (type boolean)
```
j-zimnowoda, line 229: This description is not needed. The `properties` define all types.

j-zimnowoda (changes requested):
Thank @svatwork for making a blueprint.
Regarding the schema itself there are few minor issues.

Regarding the design of the overall feature it is not clear to me when a job is going to be executed. Is it each time an application is redeployed or it is just one time job?

values-schema.yaml
j-zimnowoda, line 235: There is also job type property `type`.
If `type=cronjob` then the `schedule` property is used.
You can find this at `charts/jobs/templates/cronjob.yaml`

values-schema.yaml
svatwork: Yes that makes sense.

values-schema.yaml
svatwork: Yes that makes sense. I think having a job definition with a single object and define arrays for multiple jobs elsewhere is even better, right?

values-schema.yaml
svatwork: `script` already exists right? `script: (type string)`

But that's good, I was just working off the `jobs.demo.yaml` @Morriz gave me.

values-schema.yaml
svatwork: It seems superfluous, but when I hover over the job property I really like having an overview of the capabilities.

svatwork: Screenshot 2020-12-10 at 13 25 11

svatwork: Screenshot 2020-12-10 at 13 27 17

svatwork: It isn't refined, just to give a first impression.

values-schema.yaml
svatwork: Ok, good to know!

values-schema.yaml
```diff
@@ -602,14 +644,14 @@ properties:
...
         additionalProperties: false
         patternProperties:
-          'certs-aws|harbor|ingress-azure|keycloak':
+          certs-aws|harbor|ingress-azure|keycloak:
             type: object
             properties:
               image:
-                '$ref': '#/definitions/image'
+                $ref: '#/definitions/image'
+        type: object
```
svatwork, line 653: I don't feel like breaking this right now

values-schema.yaml
svatwork: I see that I can add a few more like spec.mounts, should I?

Morriz:
In draft v7 I know that `additionalProperties` is polymorphic, and can also be `false`.

To answer the quote question: only use them when needed. Everything is a string unless it is clearly not ;p (google yaml characteristics). We have coding principles like DRY and YAGNI, which together also dictate not to use redundant excess characters.

Now I am typing on my phone and can't see if $ is a special char that needs quoting if intended as literal, but I assume so. Same goes for strings involving hash tags and colons.

Morriz:
Looking at the error in our pipeline, I assume you committed code NOT quoting the reserved yaml key word `$ref`?

charts/jobs/templates/cronjob.yaml
```diff
@@ -1,6 +1,6 @@
 {{- $v := .Values }}
 {{- range $name, $spec := $v.tasks }}
-{{- if and $spec.enabled (eq $spec.type "cronjob") }}
+{{- if and $spec.enabled (eq $spec.jobType "cronjob") }}
```
Morriz, line 3: Why, when we are in the context of jobs, can we not refer to "type" of a job (which to me obviously implies that it refers to a job)?

I actually like this implicitness as it is distinct at the same time. It can ONLY refer to the main subject in question: the job.

I dislike it when I am treated as deemed incapable of NOT making that distinction. The logic and linguistics are clear.

Same goes for giving an "id" prop to an entity. It does not need to duplicate it's shape in its id naming, when the shape itself is known. I actually think it's ugly and confusing.

.vscode/settings.json
```diff
@@ -24,8 +24,7 @@
     "CONTRIBUTING": "markdown"
   },
   "yaml.schemas": {
-    // "http://json-schema.org/draft/2019-09/schema#": "./values-schema.yaml",
-    // "http://json-schema.org/draft/2019-09/schema#": ".vscode/values-schema.yaml",
+    "http://json-schema.org/draft-07/schema#": "values-schema.yaml",
```
Morriz, line 27: I disabled this as it was also applying the same schema suddenly to the values. Can you confirm your change is working as intended and not breaking values validation?

values-schema.yaml
```diff
@@ -1,5 +1,5 @@
 ---
-'$schema': 'http://json-schema.org/draft-07/schema'
+$schema: http://json-schema.org/draft-07/schema#
```
Morriz, line 2: Does this work?

values-schema.yaml
```diff
@@ -377,6 +405,22 @@ definitions:
...
+          You may pass an array of Job objects. The Kubernetes API accepts a Job specification, 
+          which is meant for run-to-completion apps (e.g. batch Jobs).
+
+              enabled: (type boolean)
+                  env: (type strings[], pattern-matched)
+                  image: (type string, pattern-matched)
+                  init: (type object)
+                    image: (type string, pattern-matched)
+                    script: (type string)
+                    secret: (type string[])
```
Morriz, line 419: Please refrain from these using such blocks and let prop descriptions speak for themselves. This is not documentation.

Morriz (changes requested):

values-schema.yaml
Morriz: I think there already exists that option to use mounts in the job chart. (I believe I added it.) Please take a look

svatwork: @Morriz I mean to add it to the `values-schema.yaml`

svatwork:
> Looking at the error in our pipeline, I assume you committed code NOT quoting the reserved yaml key word `$ref`?

I'll have a look. I would like to focus on semantics for now and discuss the structure if that's ok. So I expect to break it a few more times before thoroughly understanding the syntax.

charts/jobs/templates/cronjob.yaml
svatwork: I agree, I just had a syntax problem in this code block (pseudo-code for ease):

```
definitions:
  job:
    type:
      type: string
```

I take it this is valid syntax? If so my modification from type to jobType isn't necessary.

.vscode/settings.json
svatwork: I don't understand, you mean that it would evaluate any `.yaml` file with the `json-schema draft 07`? I didn't have that experience on my computer to confirm.

values-schema.yaml
svatwork: Well, my yaml language server is very inconsistent. However, it seems to be the same as quoted. I've converted both to json, with quotes:

```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "additionalProperties": false,
  "definitions": {
```

without:

```
{
  "$schema": "http://json-schema.org/draft-07/schema#",
  "additionalProperties": false,
  "definitions": {
```

So I don't think we need quotes, unless I'm mistaken.

values-schema.yaml
svatwork: I realize that, I'll delete it in a sec. But maybe you thought this feature would be convenient for a client who is writing a new yaml file in otomi-values. It would look like this:

svatwork: It's what most extensions use of, that's how I got the idea. So for example with the Kubernetes extension you can do something like this:

Screenshot 2020-12-11 at 10 04 26

Screenshot 2020-12-11 at 10 05 21
Screenshot 2020-12-11 at 10 04 26

svatwork: Maybe it's a bit overkill for now.

values-schema.yaml
```diff
@@ -16,14 +14,18 @@ definitions:
         properties:
           critical:
             $ref: '#/definitions/email'
-            description: One or more email addresses (comma separated) for critical events.
+            description: One or more email addresses (comma separated) for critical
```
svatwork, line 17: There is just sorting & formatting here, I don't know if it's useful/outside of the scope but it reads a little easier.

values-schema.yaml
```diff
@@ -1525,7 +1525,7 @@ properties:
         patternProperties:
           ^[a-z-]+$:
             $ref: '#/definitions/team'
-        type: object
+        type: array
```
svatwork, line 1528: This makes more sense to me syntactically, but I don't know what it's gonna break.

Screenshot 2020-12-11 at 13 16 53

values-schema.yaml
svatwork: PS I didn't do this by hand of course

values-schema.yaml
svatwork: Either that or "teams" should just be one team of course.

svatwork:
I feel like I'm going out of scope on this one. I created an issue outlining my idea in #241. I will abandon the tangent for now and just use what's there at the moment.

PR #240, opened 2020-12-11
feat: added documentation for schema validation

svatwork:
As discussed in https://github.com/redkubes/otomi-core/discussions/235, here's a proposal for schema validation documentation.


docs/CONTRIBUTING.md
```diff
@@ -32,3 +32,75 @@ Coming soon!
 After values are pushed the cluster's drone will do a deployment run. Helmfile diff is used targeting `$CLOUD` & `$CLUSTER` before the final deployment of the artifacts.
 
 If you have any meaningful additions to this, please let us know!
+
+## Schema validation
```
Morriz, line 36: I would like this CONTRIBUTING file will stay without detailed instructions, so you will have to add own doc file(s) for this, with a reference from the README. Please keep file names lowercase and with dashes (CONTRIBUTING and LICENSE etc are special files).

docs/CONTRIBUTING.md
```diff
@@ -32,3 +32,75 @@ Coming soon!
...
+
+### Guidelines/best practices
+
+We intend to be very exact in our speech regarding schema validation. Currently we perform meta-schema validation on the `values-schema.yaml` file. By having accurate and consistent descriptions (using the `description` keyword), we can provide the user with a better experience in his configuration management of the `otomi-values` repository.
```
Morriz, line 40: First of all, even though I like your literate style of writing, I do think it is bloated and suggest you write more concise and to the point. Ask yourself questions like "What am I trying to convey?" and "How can I say this as simple and to the point as possible?". Redacting your own writing a couple of times usually works wonders. 

Second, I don't think adding instructions on how to create good schema files should go here. Maybe add a link to the instructions themselves. Again, we will NOT be copying instructions that come with the open source tools and solutions we use. 

So please find where that information lives and add a sentence guiding people there.

docs/CONTRIBUTING.md
```diff
@@ -32,3 +32,75 @@ Coming soon!
...
+
+[Under the Apache V2.0 license agreement we are free to distribute the following content:](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#validation)
```
Morriz, line 42: I don't understand why you bring this in here

docs/CONTRIBUTING.md
```diff
@@ -32,3 +32,75 @@ Coming soon!
...
+
+API objects are validated upon receipt by the apiserver. Validation errors are flagged and returned to the caller in a Failure status with reason set to Invalid. In order to facilitate consistent error messages, we ask that validation logic adheres to the following guidelines whenever possible (though exceptional cases will exist).
```
Morriz, line 44: schema validation is not related to the k8s apiserver, so again I don't understand why this is brought inhere

docs/CONTRIBUTING.md
```diff
@@ -32,3 +32,75 @@ Coming soon!
...
+type: object
+additionalProperties: false
+properties:
+config:
+type: object
+properties:
+cookieSecret:
+type: string
+description: Cookie secret must be 128 bit base64 encoded string.
+pattern: ^(?:[A-Za-z0-9+/]{4})\\\*(?:[A-Za-z0-9+/]{2}==|[A-Za-z0-9+/]{3}=)?\$
```
Morriz, line 72: you are probably not aware that spaces were not copied here, and formatting does not work, so I conclude you have not looked at your copy paste action, which I hope you will in the future ;p

docs/CONTRIBUTING.md
```diff
@@ -32,3 +32,75 @@ Coming soon!
...
+- Telling users what they CAN do is more useful than telling them what they CANNOT do.
+- When asserting a requirement in the positive, use "must". Examples: "must be greater than 0", "must match regex '[a-z]+'". Words like "should" imply that the assertion is optional, and must be avoided.
+- When asserting a formatting requirement in the negative, use "must not". Example: "must not contain '..'". Words like "should not" imply that the assertion is optional, and must be avoided.
+- When asserting a behavioral requirement in the negative, use "may not". Examples: "may not be specified when otherField is empty", "only name may be specified".
+- When referencing a literal string value, indicate the literal in single-quotes. Example: "must not contain '..'".
+- When referencing another field name, indicate the name in back-quotes. Example: "must be greater than request".
+- When specifying inequalities, use words rather than symbols. Examples: "must be less than 256", "must be greater than or equal to 0". Do not use words like "larger than", "bigger than", "more than", "higher than", etc.
+- When specifying numeric ranges, use inclusive ranges when possible.
```
Morriz, line 55: So this kind of info lives somewhere else and should be pointed to

Morriz (changes requested):

docs/CONTRIBUTING.md
svatwork: Yes I agree, it's just my first draft. When I read it again I can usually rewrite it much better.

docs/CONTRIBUTING.md
svatwork: I thought it would be more professional respecting the copyright after copy-and-paste

docs/CONTRIBUTING.md
svatwork: This is just the header for the schema guidelines by the Kubernetes docs. Furthermore, after meta schema validation, they should be validated against the Kubernetes API server. So I thought adding this bit will make people understand the eventual target, don't you agree?

docs/CONTRIBUTING.md
svatwork: Yes, thanks for noticing.

docs/CONTRIBUTING.md
svatwork: That sounds good, but do you propose to do the same then for GO_TEMPLATING.md which has a reference from CONTRIBUTING.md? Or are these a special exemption I'm not aware of?

svatwork: Maybe we could also use the Github Wiki feature?

docs/CONTRIBUTING.md
svatwork: > Second, I don't think adding instructions on how to create good schema files should go here. Maybe add a link to the instructions themselves. Again, we will NOT be copying instructions that come with the open source tools and solutions we use.

Yes that makes sense, will look at it now.

docs/meta-schema-validation.md
```diff
@@ -0,0 +1,59 @@
...
+```
+
+Seeing a small improvement:
+
+```
+
+armAuth:
+   type: object
+   additionalProperties: false
+   description: A service Principal secret ey An explanation about the purpose of this instance.
```
Morriz, line 36: what is going on here in this line? weird language

docs/meta-schema-validation.md
```diff
@@ -0,0 +1,59 @@
...
+   properties:
+   secretJSON:
+      description: **Must be** a service Principal secret JSON key (base64 encoded)
```
Morriz, line 54: Should we use markdown in description fields? I don't know if this is valid, but if not please refrain from using that.

Morriz (changes requested):

docs/meta-schema-validation.md
svatwork: It's like this in `values-schema.yaml`... will modify both.

docs/meta-schema-validation.md
```diff
@@ -0,0 +1,59 @@
+# Meta-schema validation
+
+We validate `values-schema.yaml` using JSON Schema. Refer to `values-schema.yaml` for the supported `$schema`, it should always be on top of the file for easy reference.
+
+## Guidelines for describing strings
+
+By having accurate and consistent descriptions (refer to `description` keys in `values-schema.yaml`), we can provide a better experience in configuring `otomi-values`. Refer to the [Kubernetes API documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#validation) on how to approach API objects.
```
Morriz, line 7:
```suggestion
By having accurate and consistent descriptions (refer to `description` keys in `values-schema.yaml`), we can provide a better experience in configuring `otomi-values`. Refer to the [Kubernetes API documentation](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#validation) on conventions for description writing.
```

Morriz (changes requested):

Morriz (approved):

PR #244, opened 2020-12-11
fix: this will repair generate_loose_schema

svatwork:
As mentioned in #243 (see comments), the generate_loose_schema function is broken because ENV_DIR is not available if it's run as a Docker command.


svatwork:
Never mind. After a few runs it copied `values-schema.yaml` to `otomi-core/.values` for some reason.

PR #247, opened 2020-12-14
fix: modify check empty parameters

svatwork:
Relating to #228. This modification will execute `otomi decrypt` without parameters as `drun bin/crypt.sh decrypt` which will decrypt all files in `otomi-values`.


svatwork:
My apologies, build will fail until #246 is merged.

Morriz (approved):

PR #258, opened 2020-12-21
Team jobs

svatwork:
### Inside the scope

See related issue on the right-hand side.
Task(s) (see issue): 
- Refactor Job/CronJob chart to enable the spec.enabled flag by default.
- Team jobs basic functionality
  - Teams can enter jobs in jobs.demo.yaml.
  - Add functionality to Helmfile (helmfile-60.teams.yaml).

NOTE: the above mentioned tasks still assume that there still exists a "patternProperties object" of jobs. There will be a separate pull request for the refactoring of that task.

### Outside the scope

I went a bit out of scope, I made these changes:

- Modified the naming of 'tasks' to 'jobs'. It didn't make sense to me semantically. They are not 'tasks' by Kubernetes API definition. Note: I changed commands from `npm run tasks:keycloak` to `npm run jobs:keycloak`, did I break anything?
- I changed the ordering of the 'if' of isMultitenant and `hasKey $team services`. The ordering was first like this:

```
{{- range $teamId, $team := $tc.teams }}
{{- if hasKey $team "services" }}
...
{{- if $v.otomi.isMultitenant }}        
...
{{- end }}
<-- new key over here? E.g. {{- if hasKey $team "jobs" }}
{{- end }}
<-- new key over here?
{{- end }}
```
In this ordering, you can't add another $team key because you can't place it outside the scope of `if hasKey $team "services"`  and inside the scope of `if $v.otomi.isMultitenant` simultaneously. Anyhow, I'm not sure if it's supposed to be in the scope of if `$v.otomi.isMultitenant`, please review.


svatwork:
Note: in #248 there were still some open discussions.

svatwork:
Task: Refactor Job/CronJob chart to enable to specify an array of Jobs.

There wasn't any modification necessary to specify an array of jobs, see `jobs.demo.yaml`. It's actually a bit awkward, because if I understand correctly, the Job/CronJob chart was already looping over all the items in the job array (or object) (`{{- range $name, $spec := $v.jobs }}`, but for all present values files there weren't any jobs items/objects. Specifying an array now will template the following:

```
--
# Source: jobs/templates/job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: 0-aaa23533-36ad-45d7-8773-e0921b4893fe
  labels:
    helm.sh/chart: jobs-0.1.0
    app.kubernetes.io/name: jobs
    app.kubernetes.io/instance: jobs-demo
    app.kubernetes.io/version: "0.1.0"
    app.kubernetes.io/managed-by: Helm
spec:
  template:
    metadata:
      labels:
        helm.sh/chart: jobs-0.1.0
        app.kubernetes.io/name: jobs
        app.kubernetes.io/instance: jobs-demo
        app.kubernetes.io/version: "0.1.0"
        app.kubernetes.io/managed-by: Helm
      annotations:
        checksum/secret: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546
    spec:
      serviceAccountName: default
      containers:
      - image: "busybox:latest"
        imagePullPolicy: "IfNotPresent"
        name: "0"
        command: ["sh", "-c"]
        args:
          - |
            echo "Hello World"
        resources:
          limits:
            cpu: 200m
            memory: 256Mi
          requests:
            cpu: 100m
            memory: 128Mi
        envFrom:
        volumeMounts:
          - name: fakeroot
            mountPath: /fakeroot.pem
            subPath: fakeroot.pem
      restartPolicy: Never
      volumes:
        - name: fakeroot
          configMap:
            name: fakeroot
  backoffLimit: 3
  ttlSecondsAfterFinished: 86400
---
# Source: jobs/templates/job.yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: 1-58be380e-9a4a-4767-b93a-df501163ae67
  labels:
    helm.sh/chart: jobs-0.1.0
    app.kubernetes.io/name: jobs
    app.kubernetes.io/instance: jobs-demo
    app.kubernetes.io/version: "0.1.0"
    app.kubernetes.io/managed-by: Helm
spec:
  template:
    metadata:
      labels:
        helm.sh/chart: jobs-0.1.0
        app.kubernetes.io/name: jobs
        app.kubernetes.io/instance: jobs-demo
        app.kubernetes.io/version: "0.1.0"
        app.kubernetes.io/managed-by: Helm
      annotations:
        checksum/secret: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546
        checksum/config: 01ba4719c80b6fe911b091a7c05124b64eeece964e09c058ef8f9805daca546
    spec:
      serviceAccountName: default
      containers:
      - image: "busybox:latest"
        imagePullPolicy: "IfNotPresent"
        name: "1"
        command: ["sh", "-c"]
        args:
          - |
            echo "Hello World too"
        resources:
          limits:
            cpu: 200m
            memory: 256Mi
          requests:
            cpu: 100m
            memory: 128Mi
        envFrom:
        volumeMounts:
          - name: fakeroot
            mountPath: /fakeroot.pem
            subPath: fakeroot.pem
      restartPolicy: Never
      volumes:
        - name: fakeroot
          configMap:
            name: fakeroot
  backoffLimit: 3
  ttlSecondsAfterFinished: 86400
```

.demo/env/clouds/aws/demo/jobs.demo.yaml
```diff
@@ -0,0 +1,16 @@
...
+    teams:
+        demo:
+            jobs:
+                - type: job
+                  image:
+                      repository: busybox
+                      tag: latest
+                      pullPolicy: IfNotPresent
+                  script: echo "Hello World"
+                - type: job
```
Morriz, line 11: can this be a cronjob then?

helmfile.d/helmfile-60.teams.yaml
```diff
@@ -11,6 +11,7 @@ bases:
 
 releases:
   {{- range $teamId, $team := $tc.teams }}
+  {{- if $v.otomi.isMultitenant }}
```
Morriz, line 14: are you sure you know what you are doing here? Why iterate over something that might be false for all iterations? Might be better to put it one line higher.

But on another note: This effectively disables all of the functionality for teams if `isMultitenant`, and on which assumption did you base that decision? Please consult for full understanding in case you don't have clarity on the specs.

Morriz: When `isMultitenant=false` the concept of 'teams' still exist, as it is merely a way of grouping (team/project/whatever) resources, but we don't need every team to get their own telemetry and other heavy stuff.

values-schema.yaml
```diff
@@ -172,8 +172,22 @@ definitions:
...
   env:
-    pattern: '^[a-zA-Z0-9_]*$'
-    type: string
+    items:
+      additionalProperties: false
+      properties:
+        name:
+          pattern: '^[a-zA-Z0-9_]*$'
+          type: string
+        value:
```
Morriz, line 181: as discussed combine both props into patternProperties

Morriz (changes requested):

helmfile.d/helmfile-60.teams.yaml
svatwork: Err ok, I'll just move it back to where it came from and put team jobs outside of that scope.

svatwork:
So there's some extra functionality left over, I've updated the requirements. As far as I am concerned these two branches may be merged and the extra functionality for a job running always or onChange comes in a separate pull request. But that one first needs a proposal/proof of concept.

helmfile.d/helmfile-60.teams.yaml
```diff
@@ -187,7 +186,17 @@ releases:
           - azure{{ end }}
           {{- if $team | get "stack.sitespeed" false }}
           - sitespeed{{ end }}
-
   {{- end }}
   {{- end }}
+  {{- if hasKey $team "jobs" }}
+  - name: jobs-{{ $teamId }}
+    installed: {{ has $v.cluster.id $team.clusters }}
```
Morriz, line 193: You are missing an essential insight here:

Jobs, like services, are ONLY defined for ONE cluster scope, so they run only there. Don't deviate from that approach. You now publish jobs to ALL clusters the team operates on.

Morriz: I would say add your job snippet in the loop that deals with the team services.

values-schema.yaml
```diff
@@ -63,6 +63,7 @@ definitions:
     $ref: '#/definitions/labelsAnnotations'
     description: A set of annotations.
   azureMonitor:
+    readOnly: false
```
Morriz, line 66: Again, I have no clue why this property is added. Maybe it wasn't you and some weird merging is happening, but I never saw it before and it should not be here I think.

Morriz (changes requested):
Look, I don't wish to review redundant code. I want you to start working in one PR to get work done. This is unworkable for reviewers.

So please merge into one PR and ask for another review.

But first address these comments here, as they are important.

values-schema.yaml
svatwork: You're right, I don't know what this is doing here. I'll remove it.

svatwork:
> Look, I don't wish to review redundant code. I want you to start working in one PR to get work done. This is unworkable for reviewers.

Yes, apologies. You told me you didn't like big pr's. I branched out because I was waiting for comments on the previous one and wanted to continue work.

helmfile.d/helmfile-60.teams.yaml
svatwork: > Jobs, like services, are ONLY defined for ONE cluster scope, so they run only there. Don't deviate from that approach. You now publish jobs to ALL clusters the team operates on.

Forgive me if I'm wrong, I don't fully understand everything about this syntax yet. Of course I don't want publish jobs to all clusters. But I don't think that's necessarily true with this condition: 
`{{- if hasKey $team "jobs" }}`

> I would say add your job snippet in the loop that deals with the team services.

Based on your suggestion, `jobs` will only be available if there is a `services` key, which is not necessarily true, correct? Although I reckon the `services` key will always be used.

svatwork:
Apologies for inconvenience, this will be the dedicated pull request. 

Bump for easy reading.

svatwork:
So after discovering that a `uuid` function was creating a new release every time, the issue is actually much simpler imo. Please let me know if my way of thinking is wrong now:

We have three cases: once, always and onSpecChange.
- `once` is supported by defining the job for the first time. 
- Without a unique name, a new release is not created, therefore `onSpecChange` (or equivalent definition) is supported by modifying the Job values, since helmfile apply will successfully `diff` and see there's a change.
- `always` is supported if a new unique name is used.

Before any modification, we had the following cases for "our" Jobs:
- Any cronjob should be deployed `once` and with modification `onSpecChange`
- Any job should always be deployed again (inferred from the fact that every new helmfile apply command would replace metadata.name with a unique name.

The most recent commits in #258 should reflect old behaviour with the addition that any new Jobs (excluding CronJobs) defined by users will support all cases too.

I believe we can just check for a boolean value then and resume helmfile apply behaviour.

svatwork:
I think it can be a bit more polished, but the core idea is present. Thoughts?

values-schema.yaml
```diff
@@ -172,8 +172,14 @@ definitions:
     pattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
     type: string
   env:
-    pattern: '^[a-zA-Z0-9_]*$'
-    type: string
+    items:
+      additionalProperties: false
+      patternProperties:
+        '^[a-zA-Z0-9_]*$':
+          type: string
```
j-zimnowoda, line 179: please add
`maxLength: 131072`

values-schema.yaml
```diff
@@ -172,8 +172,14 @@ definitions:
     pattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
     type: string
   env:
-    pattern: '^[a-zA-Z0-9_]*$'
-    type: string
+    items:
+      additionalProperties: false
+      patternProperties:
+        '^[a-zA-Z0-9_]*$':
```
j-zimnowoda, line 178: env variable name should not start from number as these are resaved names: e.g.: $0, $3
Please use the following regex:
```[a-zA-Z_]{1,}[a-zA-Z0-9_]{0,}```

File: values-schema.yaml [hunk 8]
```diff
@@ -207,6 +214,61 @@ definitions:
...
+    description: You may pass an array of Job objects. The Kubernetes API accepts
+      a Job specification, which is meant for run-to-completion apps (e.g. batch
+      Jobs).
+    additionalProperties: false
+    items:
+      additionalProperties: false
+      properties:
+        deployAlways:
+          description: |
+            May give a boolean value. This is false by default.
```
j-zimnowoda, line 227: ` May give a boolean value.` - unnecessary as there is a type already indicating it
`This is false by default.` - just indicate it as a property: `default: false`

values-schema.yaml
```diff
@@ -207,6 +214,61 @@ definitions:
...
+            If true, this job will run every `helmfile apply`.
```
j-zimnowoda, line 228: There are many helmfile commands that would ultimately trigger deployment. I suggest to rephrase to:
`If true, this job will run every deployment `

values-schema.yaml
```diff
@@ -207,6 +214,61 @@ definitions:
...
+        resources:
+          $ref: '#/definitions/resources'
+        schedule:
+          description: Must give a cron-type expression if the job-type is 'cronjob' (job.type).
+          type: string
+        script:
+          $ref: '#/definitions/script'
+        ttlSecondsAfterFinished:
+          default: 86400
+          pattern: ^[0-9]*$
```
j-zimnowoda, line 256: This pattern is not needed. The type already indicates it.

helmfile.d/helmfile-60.teams.yaml
j-zimnowoda: If jobs are defined only in cluster file then it should be fine. E.g.:
```
env/clouds/aws/demo/jobs.demo.yaml
```
```
env/clouds/aws/dev/jobs.demo.yaml
```

j-zimnowoda (changes requested):
@svatwork thanks for your work.

I put some remarks to existing changes and have some questions regarding missing parts:
1) the jobs values defined for a specific cluster need to be loaded at: helmfile.d/snippets/env.gotmpl
2) Maurice' remark about treating jobs as services may be right (I still need to think about it)
3) I am still missing a good user stories for executing jobs `once` and `with each` deployment. When I am thinking about that most probably I would like to trigger a job after a specific service is redeployed or I would like to be able to trigger job manually on demand.

svatwork:
Thanks for the review @j-zimnowoda! When you are modifying those YAML files you don't notice those things you point out. I'll apply those changes asap. Also, I don't fully understand the structure yet so it's best if you point me in the direction of the files I'm skipping over.

.demo/env/clouds/aws/demo/jobs.demo.yaml
```diff
@@ -0,0 +1,21 @@
...
+            jobs:
+                - deployAlways: true
+                  image:
+                      repository: busybox
+                      tag: latest
+                      pullPolicy: IfNotPresent
+                  name: hello-world
+                  script: echo "Hello World"
+                  type: job
+                - deployAlways: true
```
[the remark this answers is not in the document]
Morriz: indeed...and can we also add the 3rd use case for `deployAlways: false`?

helmfile.d/helmfile-60.teams.yaml
Morriz: Of course, I am too much out of the game it seems ;p

Morriz (changes requested):
last thing

j-zimnowoda:
> Thanks for the review @j-zimnowoda! 
You welcome, please apply changes soon.

Morriz:
Almost there @svatwork

svatwork:
Yes apologies, I usually wait for @Morriz's review. I'll do it first thing tomorrow.

values-schema.yaml
```diff
@@ -172,8 +172,14 @@ definitions:
     pattern: '^(([^<>()\[\]\\.,;:\s@"]+(\.[^<>()\[\]\\.,;:\s@"]+)*)|(".+"))@((\[[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}\.[0-9]{1,3}])|(([a-zA-Z\-0-9]+\.)+[a-zA-Z]{2,}))$'
     type: string
   env:
```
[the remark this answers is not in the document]
svatwork: What are you referring to?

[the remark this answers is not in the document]
j-zimnowoda: I guess my typo ;)

values-schema.yaml
j-zimnowoda: It is not solved buddy

values-schema.yaml
j-zimnowoda: I don't see any response or code change. I do not understand why it is marked as resolved

values-schema.yaml
j-zimnowoda: I don't see any response or code change. I do not understand why it is marked as resolved

values-schema.yaml
j-zimnowoda: I don't see any response or code change. I do not understand why it is marked as resolved

values-schema.yaml
j-zimnowoda: I don't see any response or code change. I do not understand why it is marked as resolved

.demo/env/clouds/aws/demo/jobs.demo.yaml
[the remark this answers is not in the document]
j-zimnowoda: I don't see any response or code change. I do not understand why it is marked as resolved

j-zimnowoda (changes requested):
@svatwork please do not resolve conversations without any comment or code change.
It does not mean that you need to change each line that reviewer points out but it would be nice to know why you make certain decisions ;)

values-schema.yaml
svatwork: I do see the code change? 

```
patternProperties:
         '[a-zA-Z_]{1,}[a-zA-Z0-9_]{0,}':
           maxLength: 131072
           type: string
```

svatwork:
> please do not resolve conversations without any comment or code change.
It does not mean that you need to change each line that reviewer points out but it would be nice to know why you make certain decisions ;)

I'm sorry but I had some difficulty with the merge, apparently the last commit merge back into an older version. I did make the changes as requested.

values/jobs/keycloak.gotmpl
```diff
@@ -12,10 +12,10 @@
 {{- $cm := $c | get "cert-manager" -}}
 {{- $k := $c | get "keycloak" dict }}
 {{- $skipVerify := eq ($cm | get "stage") "staging" }}
-tasks:
-
+jobs:
   keycloak:
     type: job
+    deployAlways: true
```
Morriz, line 18: Btw, this only needs to run "OnChange".

Morriz (changes requested):
Small potatoes

.demo/env/clouds/aws/demo/jobs.demo.yaml
```diff
@@ -0,0 +1,28 @@
+teamConfig:
+    teams:
+        demo:
+            jobs:
+                - deployAlways: true
```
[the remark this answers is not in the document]
j-zimnowoda: Maurice, it seems some clarification is needed.
In case of teams job how does a user define "things" to calculate hash from.  And what are "things"?
Please be more specific.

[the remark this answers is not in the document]
Morriz: Like I said, please look at that chart

values-schema.yaml
```diff
@@ -528,11 +597,23 @@ definitions:
...
+      additionalProperties: false
+      properties:
+        name:
+          description: Name must match mount name.
+          type: string
+        configMap:
+          properties:
+            name:
+              type: string
+    type: array
```
Morriz, line 611: so this whole part can go as it will come from `mounts`

Morriz (changes requested):

.demo/env/clouds/aws/demo/jobs.demo.yaml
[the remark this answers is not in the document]
svatwork: Err, why not stick with simple if else logic and check for equality for the presence of this key?

values/jobs/keycloak.gotmpl
svatwork: That could be, but the original logic was "Always" (or equivalent).

values-schema.yaml
```diff
@@ -225,12 +287,27 @@ definitions:
...
+      name:
+        description: Must match relevant spec.volumes[].name.
+        type: string
+      subPath:
+        description: May specify path to mount the file(s) in the Pod's volume.
+        type: string
+  mounts:
+    items:
+      $ref: '#/definitions/mount'
+    type: array
```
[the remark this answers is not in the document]
svatwork: > The vision is that we offer abstractions that deal with what is published via otomi only.

I agree with the premise, but "the" vision could have been documented somewhere, which would be useful for reference right now. Now I stand accused of not following the vision. 

> We will allow the mounting of configmaps and secrets only, so we have to offer abstractions for that. So please come up with requirements for that first.

Yes, we could've come up with requirements, which I've been asking for.

> Suggestion

The suggestion makes sense to me.

svatwork:
Well we don't have guidelines for code reviews yet, so this is quite arbitrary of course. But I feel like this violating the idea/principle/whatever of "[Find an end](https://chromium.googlesource.com/chromium/src/+/HEAD/docs/cr_respect.md#find-an-end)"

values-schema.yaml
```diff
@@ -207,6 +215,60 @@ definitions:
...
+    additionalProperties: false
+    items:
+      additionalProperties: false
+      properties:
+        deployAlways:
+          default: false
+          description: |
+            If true, this job will run every deployment.
+          type: boolean
+        enabled:
```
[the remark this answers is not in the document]
svatwork: So I infer that you want to leave this code as is and end the discussion about the flag?

values-schema.yaml
j-zimnowoda: ok

j-zimnowoda:
Tests performed on cluster revealed few issues.
@svatwork is working on them.

.demo/env/clouds/aws/demo/jobs.demo.yaml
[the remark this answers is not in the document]
j-zimnowoda: Computing hash is needed for implementing deployment onChange.
The onChange flag means that the job will be redeployed only if any ConfigMap/Secret that is referenced in Job/CronJob changes.

However, the goal of this PR has been reached - teams can define jobs/cronJobs. We just wait for final tests.

The onChange functionality is something else and should NOT be part of this PR. 
This PR is already too big and there is too much feedback that is hard to maintain and makes reviewer and author overwhelmed.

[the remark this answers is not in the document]
svatwork: Alright, thanks again.

values-schema.yaml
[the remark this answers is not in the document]
j-zimnowoda: @Morriz 
What is `subpath`?
The `key` is optional, isn't it?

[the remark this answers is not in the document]
svatwork: I think `key` is equivalent to `name`?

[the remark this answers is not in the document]
svatwork: I split this story in #286.

.demo/env/clouds/aws/demo/jobs.demo.yaml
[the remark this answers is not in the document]
svatwork: I split this in #287.

Morriz (changes requested):
looking good, now only waiting for mounts as we will never ship unintended code...so no more of your "splits" please

svatwork:
> so no more of your "splits" please

Can you please be more respectful? I thought we agreed that we wouldn't bikeshed. To quote: 

"Always ask yourself if this decision really matters in the long run, or if you‘re enforcing a subjective preference. It feels good to be right, but only one of the two participants can win that game. If it’s not important, agree to disagree, and move on."

I can agree to disagree on this, but to me it was like that the goal posts were moving. Yes, this is an important feature. But we agreed on the issue, I asked for your validation and now a new thing is added. Furthermore, it took about 3 intermediate reviews that were good at first glance before this matter came forward, which dragged this pull request on for a month. 

Like you said, sometimes there's no democratic decision possible, you imply to mean this is one of those cases.

Morriz:
@svatwork I am thankful to have had a discussion with you face to face about this. I agree I can be more respectful, and am glad we could agree on a new strategy to only start on a Story after reviewing its requirements. (PRs and Tasks are different, so please only take Stories into account when you consolidate our discussion into process.)

svatwork:
Yes me too, thank you for being so understanding.

svatwork:
I added a feature to support environment variables and secrets as direct environment variables in the pods. If this is not enough, then please let me know and I will add mounting volumes.

.demo/env/clouds/aws/demo/jobs.demo.yaml
```diff
@@ -0,0 +1,44 @@
...
+                  type: job
+                - image:
+                      repository: busybox
+                      tag: latest
+                      pullPolicy: IfNotPresent
+                  name: regular-cronjob
+                  schedule: '0 1 * * *'
+                  script: echo "Hello World every day"
+                  type: cronjob
+                - env:
```
Morriz, line 33: env is fine, as we will still allow users to add env vars directly.

_But_ we will also let them create configmaps, just like secrets, so we will create `teams/config.*.yaml` which will contain configuration to deploy configmaps. Then these can be referenced by name as well (just like secrets), and array of keys selected.

Morriz: For both secrets and configmaps, for each key an optional path and subpath can be given: to allow users to give a path in the container where the key will be mounted (see example snippet I gave for schema before).

charts/jobs/templates/cronjob.yaml
```diff
@@ -21,7 +21,7 @@ spec:
           containers:
             - image: "{{ $spec.image.repository }}:{{ $spec.image.tag }}"
               imagePullPolicy: "{{ $spec.image.pullPolicy }}"
-              name: {{ $name }}
+              name: "{{ $spec.name }}"
```
Morriz, line 24: why the quotes if we know we don't need them (looking at the schema for container names)?

Morriz: I think you are trying to cover for empty string, but those are invalid, and will be caught by validator. Please understand these distinctions and don't solve in one domain what should be solved in another.

charts/jobs/templates/job.yaml
```diff
@@ -29,36 +33,42 @@ spec:
       containers:
       - image: "{{ $spec.image.repository }}:{{ $spec.image.tag }}"
         imagePullPolicy: "{{ $spec.image.pullPolicy }}"
-        name: "{{ $name }}"
+        name: "{{ $spec.name }}"
```
Morriz, line 36: same, no quotes

charts/jobs/templates/job.yaml
```diff
@@ -29,36 +33,42 @@ spec:
...
+        {{- if $spec.mounts }}
         volumeMounts:
+          {{- if $spec.fakeroot }}
           - name: fakeroot
             mountPath: /fakeroot.pem
             subPath: fakeroot.pem
-          {{- if $spec.mounts }}
-          {{- $spec.mounts | toYaml | nindent 10 }}
           {{- end }}
+          {{- $spec.mounts | toYaml | nindent 10 }}
```
Morriz, line 60: what happened to the condition? I would put it back to avoid issues.

And where is the logic to create the mounts? That whole part is missing.

values/jobs/harbor.gotmpl
```diff
@@ -31,6 +32,7 @@ tasks:
       repository: {{ $c | get "jobs.harbor.image.repository" "otomi/tasks" }}
       tag: {{ $c | get "jobs.harbor.image.tag" "v0.2.1" }}
       pullPolicy: {{ $c | get "jobs.harbor.image.pullPolicy" "IfNotPresent" }}
+    name: harbor
```
Morriz, line 35: why is this redundant property put there? It is implicitly defined as key in line 18.

values/jobs/harbor.gotmpl
```diff
@@ -44,5 +46,6 @@ tasks:
       OIDC_NAME: 'keycloak'
       OIDC_SCOPE: 'openid'
       OIDC_VERIFY_CERT: '{{ not $skipVerify }}'
+    fakeroot: true
```
Morriz, line 49: what is this doing? are you hardcoding something which should be conditional?

values/jobs/keycloak.gotmpl
```diff
@@ -62,4 +63,5 @@ tasks:
           {{- end }}
           "https://harbor.{{ $v.cluster.domain }}/*"
       ]'
+    fakeroot: true
```
Morriz, line 66: again, understand first what is the result

values/jobs/harbor.gotmpl
```diff
@@ -15,9 +15,10 @@
 {{- $teamNames = print "team-" . | append $teamNames -}}
 {{- end -}}
 
-tasks:
+jobs:
   harbor:
```
Morriz, line 19: now that we don't allow external overrides of core jobs anymore (we never should have in the first place), we can make this an array of jobs, as opposed to a hash with the name as key. That way we conform to the rest of the work

values/jobs/keycloak.gotmpl
```diff
@@ -12,10 +12,10 @@
 {{- $cm := $c | get "cert-manager" -}}
 {{- $k := $c | get "keycloak" dict }}
 {{- $skipVerify := eq ($cm | get "stage") "staging" }}
-tasks:
-
+jobs:
   keycloak:
```
Morriz, line 16: here too, please make into an array

Morriz (changes requested):
Hmmm...not good progress, even though I thought the requirements are clear now. Maybe we should revisit them?

Key issue missing still: secret & configmap mounts and keys selection.

.demo/env/clouds/aws/demo/jobs.demo.yaml
```diff
@@ -0,0 +1,44 @@
...
+                      this-is: an env
+                  image:
+                      repository: busybox
+                      tag: latest
+                      pullPolicy: IfNotPresent
+                  name: job-with-env-and-secret
+                  runPolicy: OnSpecChange
+                  secret:
+                      this-is: a secret
```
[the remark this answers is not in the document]
svatwork: Yes I missed that insight. I'll continue working on it.

charts/jobs/templates/cronjob.yaml
svatwork: I think I just copied them from `job.yaml` for consistency.

.demo/env/clouds/aws/demo/jobs.demo.yaml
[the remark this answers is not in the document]
svatwork: Can we move forward on this and do I have permission to add some documentation?

.demo/env/clouds/aws/demo/jobs.demo.yaml
svatwork: Ok sounds good.

charts/jobs/templates/cronjob.yaml
svatwork: Sorry, I don't remember adding these quotes but it looks like I did. I didn't have any special reason for it.

charts/jobs/templates/job.yaml
svatwork: Do you mean the removal of line 51? It's moved to line 53 because otherwise in some cases there is an empty key `volumeMounts`. Jeho and I moved it when we were testing deployment.

values/jobs/harbor.gotmpl
svatwork: I think this is a bit contradictory if we change it to an array of jobs? But yes it was definitely redundant.

charts/jobs/templates/job.yaml
svatwork: Actually, I understand the problem now, apologies.

svatwork:
I just started a discussion (#295) about the final requirements.

charts/jobs/templates/job.yaml
```diff
@@ -1,11 +1,15 @@
...
 {{- $ := . }}
-{{- range $name, $spec := $v.tasks }}
-{{- if and $spec.enabled (eq $spec.type "job") }}
+{{- range $spec := $v.jobs }}
+{{- if and (default true $spec.enabled) (eq $spec.type "job") }}
 apiVersion: batch/v1
 kind: Job
 metadata:
-  name: {{ join "-" (list $name  uuidv4) | trunc 63 }}
+  {{- if eq $spec.runPolicy "OnSpecChange" }}
```
j-zimnowoda, line 8: A spec for an existing job cannot be redeployed. The only way to deploy a job is to ensure that it has a new name.


My proposal for `OnSpecChange` scenario is the following:
To create a job with a name that contains hash from spec. 
In this way we ensure that if there is no change the job won't be run, as it will be deployed with unchanged status, e.g.:
```
job.batch/<job-name> unchanged
```

j-zimnowoda (changes requested):

Morriz (commented):
I don't see the need for both `$ref.name` and `$ref.secretKeyRefName`. After that we should ship this if it meets requirements. We will see this during the demo I guess ;)

charts/jobs/templates/job.yaml
```diff
@@ -39,35 +54,39 @@ spec:
...
+        {{- if $spec.secretKeyRefs }}
+        env:
+          {{- range $ref := $spec.secretKeyRefs }}
+          - name: {{ $ref.name }}
+            valueFrom:
+              secretKeyRef:
+                name: {{ $ref.secretKeyRefName }}
+                key: {{ $ref.key }}
+          {{- end }}
+        {{- end }}
```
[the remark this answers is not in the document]
svatwork: I have considered this while I was going through the documentation (https://kubernetes.io/docs/concepts/configuration/secret/#using-secrets-as-environment-variables). 

It would seem to me that in the case the user wants to reference multiple keys from the same secret, she cannot do it unless she can specify a separate variable name for the referenced key.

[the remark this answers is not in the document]
Morriz: I still count only the secret name and var holding the list of keys: 2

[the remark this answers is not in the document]
Morriz: And can we refrain from using he/she? Just "one" has always done the job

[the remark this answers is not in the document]
svatwork: > I still count only the secret name and var holding the list of keys: 2

I'm sorry, I don't really understand what you imply with this statement. The documentation has an example I wanted to stick to:

```yaml
env: 
      - name: SECRET_USERNAME
        valueFrom:
          secretKeyRef:
            name: mysecret
            key: username
      - name: SECRET_PASSWORD
        valueFrom:
          secretKeyRef:
            name: mysecret
            key: password
```
`mysecret` is referenced multiple times with 2 different keys. If the name and key were the same, it would look like this in this specification:

```yaml
env: 
      - name: mysecret
        valueFrom:
          secretKeyRef:
            name: mysecret
            key: username
      - name: mysecret
        valueFrom:
          secretKeyRef:
            name: mysecret
            key: password
```

I would assume this would throw a reference error or something similar.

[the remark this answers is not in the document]
Morriz: Ok I get it, I forgot about being able to name the var name differently than the key name. Go ahead!

svatwork:
Ok, put my heart and soul into the last 1%. Go easy on me 💯 .

charts/jobs/values.yaml
```diff
@@ -6,15 +35,15 @@ resources:
...
+# schedule: '1 * * * *'
+
+script: echo "bla"
+secrets: []
+  # - name: secret-name
+  #   entries:
+  #       - foo
+  #       - bar
+
+type: job # cronjob
```
Morriz, line 48: and coming to think of it, can we stick with CamelCasing and have enum with `Job` and `CronJob`?

helmfile.d/helmfile-60.teams.yaml
```diff
@@ -187,7 +186,29 @@ releases:
           - azure{{ end }}
           {{- if $team | get "stack.sitespeed" false }}
           - sitespeed{{ end }}
-
   {{- end }}
   {{- end }}
+  {{- if hasKey $team "jobs" }}
+  {{- range $job := $team.jobs }}
+  - name: job-{{ $teamId }}-{{ $job.name }}
+    installed: {{ has $v.cluster.id $team.clusters }}
```
Morriz, line 194: the installed flag will come from the spec (users can temporarily disable them), so this is not correct

Morriz (changes requested):
2 small things left

.demo/env/clouds/aws/demo/jobs.demo.yaml
```diff
@@ -0,0 +1,109 @@
...
+                      secrets:
+                          - name: secret-name
+                            entries:
+                                - foo
+                                - bar
+                  name: contains-everything
+                  runPolicy: Always
+                  script: echo "hello $receiver"
+                  secrets:
+                      - name: secret-name
```
Morriz, line 35: this secret does not exist in team secrets ;/

bin/bootstrap.sh
```diff
@@ -16,7 +16,7 @@ has_otomi='false'
 function generate_loose_schema() {
   local targetPath="$ENV_DIR/.vscode/values-schema.yaml"
   local sourcePath="$PWD/values-schema.yaml"
-  yq r -j "${sourcePath}" | jq "del(.. | .required?)" | yq r --prettyPrint - > "${targetPath}"
+  yq r -j "${sourcePath}" | jq "del(.. | .required?)" | yq r --prettyPrint - >"${targetPath}"
```
Morriz, line 19:
```suggestion
  yq r -j "${sourcePath}" | jq "del(.. | .required?)" | yq r --prettyPrint - >"$targetPath"
```

Morriz (changes requested):
2 tiny thingsz

svatwork:
I performed the following things since we last spoke @Morriz:
- Made sure on the cluster that the secrets would be selected
- Made sure on the cluster that the files will be available if the user specifies a path. It was wise to remove the `key` and `path` properties because the projected `mountPath` will be the full path that the user will see the files at if one looks into the container (together with @j-zimnowoda)
- Merged the master branch with the conf test stuff, it caused some conflicts (together with @rawc0der)

I'm gonna take it easy and upgrade MacOS now. If you give any other comments I will do it on Monday with a clear head.

helmfile.d/helmfile-60.teams.yaml
```diff
@@ -187,7 +186,31 @@ releases:
           - azure{{ end }}
           {{- if $team | get "stack.sitespeed" false }}
           - sitespeed{{ end }}
-
   {{- end }}
   {{- end }}
+  {{- if hasKey $team "jobs" }}
```
svatwork, line 191: Before you start ranting about this check @Morriz, it introduces an error without it if you run `otomi template -l group=jobs`. If there is another solution feel free to modify it but this works for now.

charts/jobs/templates/cronjob.yaml
```diff
@@ -1,13 +1,17 @@
...
+{{- $teamSecrets := (include "itemsByName" ($v.teamSecrets | default list) | fromYaml) }}
+{{- if eq $v.type "CronJob" }}
 apiVersion: batch/v1beta1
 kind: CronJob
 metadata:
-  name: {{ $name }}
+  name: job-{{ $v.name }}
   labels: {{- include "jobs.labels" $ | nindent 4 }}
+  annotations:
+    policy.otomi.io/ignore: psp-allowed-users
```
svatwork, line 10: So this is ignoring the check right now, for the purpose of finishing the story can we please improve it later?

Morriz (approved):

Issue #259, opened 2020-12-22
Create values-schema mechanism to update values to conform to schema changes


0-sv:
# User story

As a developer making breaking changes to the values-schema, I want to have a mechanism that will copy the corresponding values to their new locations.

## Problem

We break the schema when:

1. We move properties around (more frequent)
2. We delete props (more frequent)
3. We mutate (change property type or value shape, less frequent)

`otomi-core` should have a mechanism that migrates values to conform to the new spec.

## Proposed solution

A schema will get a `version` (semver, patches are additions, minors are breaking changes) and a `changes` property holding information about breaking changes:

1. The new location (if changed) by mapping old to new
2. The list of deleted props
3. The new type (if changed) by giving a go template for transformation (see schema change below)

**Workflow:**

Every time a developer changes the schema which includes these changes as DOD, the `otomi migrate-values` command should be performed that will massage the values to migrate to the new structure.

**Schema change:**

Sample part:
```yaml
changes:
  - version: 0.23.7
    locations:
      charts.bla.someProp: someNewRootProp.someProp
    deletions: 
      - charts.bla.someOtherProp
    mutations:
      # image tag went from semver to glob
      charts.bla.image.tag: 'printf "v%s"'
  - version: 0.23.8
```

## Tasks:

core:
- [ ] Enrich values-schema.yaml
  - [x] changes property
  - [ ] a lint function to ensure that making a breaking change will be checked before commit
- [x] `otomi migrate-values-rev` to show all history from previous versions from version control.
- [ ] Create `otomi migrate-values` script that migrates otomi-values forward based on changes defined, and sets latest version in values.
  - [x] Connect `otomi-tasks` to `bin/migrate-values.sh` and `bin/otomi migrates-values`
  - [x] Load otomi-values file as json with fileName as argument to start modification per file
  - [ ] Create a mechanism that will be able to pin an `otomi-values` repository to a certain version so the modification can take place
    - [x] `getNewVersion()` mock
    - [ ] `getNewVersion()`
    - [x] `getOldVersion()` mock
    - [ ] `getOldVersion()`
  - [ ] `otomi-values` modification
    - [x] `displacements ()`
    - [ ] `deletions ()`
    - [ ] `mutations ()`
- [ ] CLI functionality

api:
- [ ] Add extra call to tools server just before deploy: migrate-values

# Definition Of Done
- [ ] Tasks are done
- [ ] (Unit) Tests are added to code
- [ ] Refactoring:
[...]

PR #260, opened 2020-12-22
Schema versioning proposal

svatwork:
Related to #259. 

Please say if this is simplistic, lacks features or for any other reason.

To me it seems obvious to use the diff between versions to migrate values. 

Furthermore, ["the "$id" keyword identifies a schema resource with its canonical URI." ](https://json-schema.org/draft/2019-09/json-schema-core.html#rfc.section.8.2.2) So there's already some room to identify the schema inside and outside of itself.


svatwork:
Build is failing for obvious reasons. I will spend time on that if we move forward.

PR #271, opened 2021-01-07
Refactor: Error echo's to STDERR

### LINKED ISSUE #270: Print error echos into SDTERR
**Describe the bug:**
Currently error messages are printed to STDOUT, they should be printed to SDTERR

**To Reproduce**
In `bin/common.sh` function `check_sops_file` echo error is not printed to STDERR
Other locations as well.

**Expected behavior:** 
Print error message to SDTERR

Dunky13:
Change echo's printing ERROR messages to print to SDTERR instead of STDOUT

resolves #270


Dunky13:
There are some commits from the other PR #269 as I branched from off from the branch related to that issue/PR

bin/common.sh
```diff
@@ -35,11 +48,11 @@ hf() {
 hf_values() {
   [ "${VERBOSE-'false'}" = 'false' ] && quiet='--quiet'
   helmfile ${quiet-} -e "$CLOUD-$CLUSTER" -f helmfile.tpl/helmfile-dump.yaml build | grep -Ev $helmfileOutputHide | sed -e $replacePathsPattern |
-    yq read -P - 'releases[0].values[0]'
+    dyq read -P - 'releases[0].values[0]'
 }
 
 prepare_crypt() {
-  [[ -z "$GCLOUD_SERVICE_KEY" ]] && echo "Error: The GCLOUD_SERVICE_KEY environment variable is not set" && exit 2
+  [[ -z "$GCLOUD_SERVICE_KEY" ]] && echo "Error: The GCLOUD_SERVICE_KEY environment variable is not set" >&2 && exit 2
```
j-zimnowoda, line 55: Can we define and use `print_error` function?
```
function print_error() {
echo "Error: $1" >&2
}
```

bin/common.sh
```diff
@@ -6,23 +6,36 @@ readonly clustersFile="$ENV_DIR/env/clusters.yaml"
 readonly helmfileOutputHide="(^\W+$|skipping|basePath=|Decrypting)"
 readonly helmfileOutputHideTpl="(^[\W^-]+$|skipping|basePath=|Decrypting)"
 readonly replacePathsPattern="s@../env@${ENV_DIR}@g"
+readonly yqDockerImage="mikefarah/yq:3"
+
+function dyq() {
+  local yqcommand=$@
```
j-zimnowoda, line 12: yq_args

bin/common.sh
```diff
@@ -6,23 +6,36 @@ readonly clustersFile="$ENV_DIR/env/clusters.yaml"
 readonly helmfileOutputHide="(^\W+$|skipping|basePath=|Decrypting)"
 readonly helmfileOutputHideTpl="(^[\W^-]+$|skipping|basePath=|Decrypting)"
 readonly replacePathsPattern="s@../env@${ENV_DIR}@g"
+readonly yqDockerImage="mikefarah/yq:3"
+
+function dyq() {
```
j-zimnowoda, line 11: You can shadow yq and use command inside, e.g.:
```
function yq {
 # use command to ignore functions/aliases
  echo "yq wrapper"
  command yq "$1"
}
```

In this way, you don't need to change any `yq` call

bin/common.sh
```diff
@@ -6,23 +6,36 @@ readonly clustersFile="$ENV_DIR/env/clusters.yaml"
...
+  local yqcommand=$@
+  if [ -x "$(command -v yq)" ] && [[ "$(yq --version 2>&1 | sed 's/yq version //g')" == "3."* ]] ; then
```
j-zimnowoda, line 13: This conditional statement may need some changes. As far I understand the requirements, we just need to check if we are running in docker context or not.
1) If docker context then `yq` should already exist as a part of the `otomi-core` docker image. 
       if yq does not exist then return error
2) If otomi command is run on a regular host then run dockerized version of `yq`.

j-zimnowoda (changes requested):
Great! Just few remarks. Please write, if you disagree.

bin/common.sh
Dunky13: In `bin/otomi` STDERR is also printed like this. So to keep the same code style I think >&2 would be most solid. That way to keep the same code style we don't need to source `bin/common.sh` for the print_error function.

bin/common.sh
Dunky13: Neat! Will change

bin/common.sh
Dunky13: That might be true, but lets assume:

1. in docker context yq exists, so first check should pass and the "yq" should run
2. On regular host
    a. yq is installed on version 3: Why not use local yq, it's more performant and no harm's done
    b.  yq is not installed (on version 3): Try to run dockerized version of yq
    
The second part of the conditional statement just makes sure that yq is running in V3. In the otomi-core image that's true, so will pass. On local machine fail if yq@4 is installed, then try dockerized yq@3

Or is it an incorrect assumption?

j-zimnowoda: 2) On host yq can have different versions that the one we are using in docker file
Maybe some have yq@3.0.0 which has some bugs and we did not tested it.
As you can see we are not able to control user binaries. Moreover we do not want to enforce a user to install any binary except docker. So I think is wise to use always the same `yq`.

Dunky13: Changed implementation, so marking this as resolved for now

Morriz (approved):

bin/common.sh
Morriz: Here I agree with @j-zimnowoda.

Morriz (changes requested):
print_error function would be nice. Could you then do that wherever we print errors?

j-zimnowoda (changes requested):
LGTM,
just implement `print_error` function as there are two votes for that solution

bin/common.sh
Dunky13: Implemented

Dunky13:
@Morriz @j-zimnowoda the `print_error` has been implemented

bin/common.sh
```diff
@@ -6,49 +6,87 @@ readonly clustersFile="$ENV_DIR/env/clusters.yaml"
...
+
+function print_error() {
+    echo "Error: " $@ >&2
+}
+
+function _shadow() {
+  local cmd="$1"
+  shift
+  local args="$@"
+  if [[ $has_docker -eq 1 && ("${IN_DOCKER}" != "1" || $dind -eq 1) ]]; then
```
j-zimnowoda, line 23: The value of `dind` is always 0. I think we can remove this flag.

j-zimnowoda (changes requested):

bin/common.sh
```diff
@@ -6,49 +6,87 @@ readonly clustersFile="$ENV_DIR/env/clusters.yaml"
 readonly helmfileOutputHide="(^\W+$|skipping|basePath=|Decrypting)"
 readonly helmfileOutputHideTpl="(^[\W^-]+$|skipping|basePath=|Decrypting)"
 readonly replacePathsPattern="s@../env@${ENV_DIR}@g"
+readonly otomiToolsImage="otomi/tools:1.4.10"
 
-get_k8s_version() {
+has_docker=$( docker --version &>/dev/null && echo "1" || echo "0")
+dind=0
```
Morriz, line 12: Can we not use 0/1 for booleans? We do have a style guide

bin/common.sh
Morriz: Yeah. Not gonna be used in this context I think.

bin/otomi
```diff
@@ -101,21 +99,16 @@ function show_usage() {
...
 
 function check_docker_running() {
-  set +e
-  docker --version &>/dev/null
-  set -e
-  if [ "$?" = "0" ]; then
-    has_docker=1
-  else
-    echo "Otomi CLI expects a running docker daemon!" >&2 && exit 1
+  if [ "${has_docker}" = "0" ]; then
```
Morriz, line 102: Please remove this pointless function, and print the message where this is invoked

Morriz (changes requested):
Couple small things left

j-zimnowoda (changes requested):
Thanks for changes. There is one question and remark about dead code

bin/common.sh
```diff
@@ -12,20 +12,39 @@ has_docker=$( docker --version &>/dev/null && echo "1" || echo "0")
...
+    exit 1
   fi
 }
 
+function yq() {
+  local yq_args="$@"
+  _shadow yq "${yq_args}"
+}
+
+# function jq() {
```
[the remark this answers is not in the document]
Dunky13: I think it's useful to shadow jq as well as yq. Since it uses otomi-tools, we can dictate which version needs to be run.

bin/common.sh
```diff
@@ -6,49 +6,87 @@ readonly clustersFile="$ENV_DIR/env/clusters.yaml"
...
+    command ${cmd} ${args}
+  else
+    print_error "Docker is not available and ${cmd} is not installed locally"
+    exit 1
+  fi
+}
+
+function yq() {
+  local yq_args="$@"
+  _shadow yq "${yq_args}"
```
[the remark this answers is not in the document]
Dunky13: When tested with quoting, it worked well. But I can change it to `_rind yq $@` if preferred.

bin/common.sh
```diff
@@ -6,49 +6,87 @@ readonly clustersFile="$ENV_DIR/env/clusters.yaml"
...
+      ${otomiToolsImage} ${cmd} ${args}
+  elif command -v ${cmd} &>/dev/null; then
+    command ${cmd} ${args}
+  else
+    print_error "Docker is not available and ${cmd} is not installed locally"
+    exit 1
+  fi
+}
+
+function yq() {
```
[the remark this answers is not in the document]
Dunky13: I did have a `dyq` initially, but @j-zimnowoda suggested to use a shadow function: https://github.com/redkubes/otomi-core/pull/271#discussion_r553363560
To run docker, if docker is not installed, run native.

bin/common.sh
Dunky13: I copied the style from the existing codebase:
https://github.com/redkubes/otomi-core/blob/master/bin/otomi#L43-L45
Where can I find the style guide?

bin/otomi
Dunky13: https://github.com/redkubes/otomi-core/blob/master/bin/otomi#L103
https://github.com/redkubes/otomi-core/blob/master/bin/otomi#L403

Again, keeping codestyle close to original. Will change

bin/common.sh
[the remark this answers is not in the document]
Morriz: Yes, let's be precise since we know exactly where we want to invoke dockerized version and also where not (in the `bin/*.sh` scripts). So we don't have to use a "smart" function that has to figure this out anymore.

bin/common.sh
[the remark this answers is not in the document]
Morriz: See remark above about being specific

bin/common.sh
Morriz: In the redkubes/docs repo

bin/common.sh
Morriz: @Dunky13 you probably CAN comment inline here, so please follow us here.

But dind might come later, so please discard.

bin/common.sh
Dunky13: I read through the repo again, and only found: https://github.com/redkubes/docs/blob/master/development/principles.md Which does not mention boolean usage.

But I'll rewrite to true/false, if that has preference.

bin/common.sh
j-zimnowoda: @Morriz your initial idea was to create a `yq` alias. We made a function instead. It shadows `yq` command in the same way. Thanks to that we do not need to change anything else. I am surprised that suddenly you changed your mind.

bin/common.sh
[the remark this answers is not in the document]
j-zimnowoda: @Morriz your initial idea was to create a yq alias. We made a function instead. It shadows yq command in the same way. Thanks to that we do not need to change anything else. I am surprised that suddenly you changed your mind.

bin/common.sh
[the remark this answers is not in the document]
j-zimnowoda: If jq is not used outside container then it is not needed.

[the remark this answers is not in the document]
Dunky13: It's used in otomi
https://github.com/redkubes/otomi-core/blob/master/bin/otomi#L177

And in common.sh as well, which is sourced by `otomi`

bin/common.sh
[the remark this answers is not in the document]
Dunky13: Discussed with @Morriz in person. yq & jq are used both in a docker environment, as well as native. Maintaining duplicate code is not useful, and having this function makes it work regardless of environment

j-zimnowoda:
Thanks for changes. Could you also take a look at build that is failing?

Dunky13:
> Thanks for changes. Could you also take a look at build that is failing?

Build fails, because the otomi/tools version it uses in `bin/common.sh` is `latest` which is (still) linked to otomi/tools:1.4.9(and 1.4.8)
Getting docker credentials from Maurice to so I can update the `otomi/tools:latest` tag to Alin's update to 1.4.11, that should resolve the issue

Dunky13:
I updated to 1.4.11, but the build is still failing... |:-/

j-zimnowoda:
> I updated to 1.4.11, but the build is still failing... |:-/

The output from build provides you some hints ;)
```
jq: error: Could not open file |: No such file or directory
jq: error: Could not open file keys[]: No such file or directory
The command '/bin/sh -c if [ "$SKIP_TESTS" = 'false' ]; then   cp -r .demo/ env/ &&   bin/validate-values.sh &&   bin/validate-templates.sh &&   bats bin/tests; fi' returned a non-zero code: 2
```

Dunky13:
Turns out I didn't handle the `$@` properly in all the places. Fixed now :)

j-zimnowoda (approved):
Thank you for your hard work.
I merged with latest master.

svatwork:
Would you please check if this also resolves #261?

Dunky13:
> Would you please check if this also resolves #261?

No, but I left a comment that uses `jq`, hopefully that helps

bin/common.sh
```diff
@@ -24,16 +24,17 @@ function err() {
 function _rind() {
   local cmd="$1"
   shift
-  local args="$@"
   if [[ "${has_docker}" == 'true' && "${IN_DOCKER}" != '1' ]]; then
     docker run --rm \
       -v ${ENV_DIR}:${ENV_DIR} \
       -e CLOUD="$CLOUD" \
-      -e IN_DOCKER="1" \\
+      -e IN_DOCKER="1" \
```
Morriz, line 31: can we progressively (or agressively) start using single quotes for non-expandable strings?

Morriz (changes requested):
nitpick then we are done

bin/common.sh
```diff
@@ -1,66 +1,72 @@
...
+readonly helmfile_output_hide_tpl="(^[\W^-]+$|skipping|basePath=|Decrypting)"
+readonly replace_paths_pattern="s@../env@${ENV_DIR}@g"
+readonly otomi_tools_image="otomi/tools:latest"
 
-has_docker=$( docker --version &>/dev/null && echo "1" || echo "0")
-dind=0
+has_docker='false'
+if docker --version &>/dev/null ; then
+  has_docker='true'
+fi
```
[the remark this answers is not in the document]
Dunky13: I agree, but @Morriz requested true/false: https://github.com/redkubes/otomi-core/pull/271#discussion_r559456999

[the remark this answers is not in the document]
Dunky13: And although true/false is not explicitly stated in our style guide, not in Google's styleguide, it is used there (second code block): https://github.com/google/styleguide/blob/gh-pages/shellguide.md#case-statement

bin/common.sh
```diff
@@ -27,9 +29,9 @@ function _rind() {
   if [[ "${has_docker}" == 'true' && "${IN_DOCKER}" != '1' ]]; then
     docker run --rm \
       -v ${ENV_DIR}:${ENV_DIR} \
-      -e CLOUD="$CLOUD" \
-      -e IN_DOCKER="1" \
-      -e CLUSTER="$CLUSTER" \
+      -e CLOUD="${CLOUD}" \
```
Morriz, line 32: Now you add accolades that have no effect and are unneeded. I would love to be DRY and aim for least effort coding.

Morriz (changes requested):
DRYing up all the way

bin/common.sh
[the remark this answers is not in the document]
svatwork: Yes, I wanted to mention this. I reviewed the style guide, which was merged today, and after that it doesn't contain any statement about not using 0 and 1. But I don't need to have the last word on it, @Morriz probably has a good reason to avoid 0 and 1.

bin/common.sh
Dunky13: @Morriz, I am not aware that I'm repeating myself, I do agree with least effort, but also staying consistent with the style guide.
https://github.com/google/styleguide/blob/gh-pages/shellguide.md#variable-expansion
The styleguide says: In order of precedence: Stay consistent with what you find; quote your variables; prefer "${var}" over "$var".
If you insist, I can change it back to $CLOUD, but then it would be inconsistent with the styleguide.

Morriz: Yeah, I sometimes literally see DRY standing for shrunken bare-minimum code, and the letter R reminds me of Redundant, which is in line with it's philosophy to not repeat yourself typing unnecessary code. Are you still with me?

As discussed, we should add this to the execeptions/additions part of our styleguide.

Morriz: Using good linters syntax errors will be caught. Devs should be schooled in the format of bash vars to differentiate what is var name and what is not.

Morriz (approved):
thanks

PR #273, opened 2021-01-08
refactor: don't require ppa, but get yq from developer's docker image

### LINKED ISSUE #272: Move yq ppa installation to docker copy
To become less reliant on 3rd party developer to maintain een ppa library from which `yq` is installed, a `COPY --from:...` should works as well.

Because less dependencies are needed, the build image is also slightly smaller

Dunky13:
Copy the yq binary directly from the developers own docker image, this way ppa is no longer required
and other dependencies can be removed, and the two run apt install layers can be merged into one.

fix #272

### DIFF
--- a/tools/Dockerfile
+++ b/tools/Dockerfile
@@ -27,9 +27,10 @@ ARG BATS_FILE_VERSION=0.2.0
 WORKDIR /
 
 RUN apt-get update -qq \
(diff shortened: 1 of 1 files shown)


j-zimnowoda (approved):
For golang application that are self contained binaries that approach should work.
We have already proceed similar way with other 3rd party tools.

Morriz (approved):

PR #307, opened 2021-02-03
fix: setting ttl to 0 to remove jobs on completion

Morriz:


j-zimnowoda (changes requested):
Have you tested it?
I tried to to create jobs with TTL `0` and `1` on AWS cluster and I do not see that controller removes them :(

Morriz:
I am expecting them too be removed according to k8s docs, and I am not in the business of testing their code. We should file a bug if what we follow is not correct.

What does your test file look like?

Morriz:
Also, that was added in 1.12. Can you investigate wether AWS has that enabled? I expect that to be GA by now, so I assumed it to be available. It must be, otherwise the controller would bork.

Morriz:
You are right, can take a lot of time before this passes as beta or ga. Closing this PR.

Issue #290, opened 2021-01-25
Propagate helmfile parameters to otomi validate-templates 


svatwork:
We want to limit `otomi validate-templates` to select charts for developing. 

Tasks:
- [ ]: Rewrite $1 part to instead take named parameter `--all` which implies to run the routine for all clusters (else the target cluster)
- [ ]: Add propagation from all helmfile parameters to templating commands (`hf template` -> `hf "$@" template`)

Morriz:
@svatwork can this be pulled forward?

svatwork:
I'm not certain what you mean, but if you mean focus on this, then yes, this is my priority.

Morriz:
closing in favor of #330

PR #330, opened 2021-02-23
Propagate helmfile parameters --all and --label

svatwork:
Solves #290.

Defensively, I added comments to the changes I made. Let's have a **respectful** discussion about it.

_Note that I didn't like to go outside of the scope, but without the changes I couldn't bring the `validate-templates` function under test coverage._


bin/bootstrap.sh
```diff
@@ -46,7 +46,7 @@ done
 for f in '.gitignore' '.prettierrc.yml' 'README.md'; do
   cp $PWD/.values/$f $ENV_DIR/
 done
-if [ -z "$(ls -A $ENV_DIR/env)" ]; then
+if [ ! -d "$ENV_DIR/env" ]; then
```
svatwork, line 49: I originally created the check but there's `-d` for that so `ls -A` will fail in edge cases (e.g. $ENV_DIR not set).

bin/common.sh
```diff
@@ -38,17 +38,17 @@ function abort() {
 }
 trap abort SIGINT
 
-#######################################
-# https://github.com/google/styleguide/blob/gh-pages/shellguide.md#stdout-vs-stderr
-#######################################
+#####################################################################################
+# https://github.com/google/styleguide/blob/gh-pages/shellguide.md#stdout-vs-stderr #
+#####################################################################################
```
svatwork, line 43: Broad interpretation of https://github.com/google/styleguide/blob/gh-pages/shellguide.md#s4.2-function-comments.

Don't mind to change it, my personal opinion is that this looks better.

bin/common.sh
```diff
@@ -38,17 +38,17 @@ function abort() {
...
 function err() {
   echo "[$(date +'%Y-%m-%dT %T.%3N')] ERROR: $*" >&2
 }
 
 function _rind() {
   local cmd="$1"
   shift
-  if [ $has_docker = 'true' ] && [ -n "$IN_DOCKER" ]; then
+  if [ $has_docker = 'true' ] && [[ ! $IN_DOCKER ]]; then
```
svatwork, line 51: `[[` is more powerful (https://stackoverflow.com/a/49825114).

Also discovered that the `[` check doesn't always work properly on Linux hosts.

bin/common.sh
```diff
@@ -128,9 +127,70 @@ function for_each_cluster() {
   done
 }
 
-hf_templates_init() {
+function hf_templates_init() {
```
svatwork, line 130: Using `function` is more consistent with all the other functions we have.

bin/common.sh
```diff
@@ -128,9 +127,70 @@ function for_each_cluster() {
...
+        *)
+          echo "Programming error: expected '--' but got $1"
+          exit 3
+          ;;
+      esac
+    done
+  else
+    echo "Error: --all or --label not specified"
+    exit 5
+  fi
```
svatwork, line 195: Implementation of https://stackoverflow.com/questions/192249/how-do-i-parse-command-line-arguments-in-bash.

My vision for this function is to extend it for all the other options we add and add them to other functions as well. This should become the source of truth for passing options. Only have to add extra options and update the function comment. 

My criticism would be that the variables are going to pollute the global namespace. We can pass them around but there isn't a function return call so I don't know the nice way of doing this.

bin/otomi
```diff
@@ -14,7 +14,7 @@
 set -e
 
 # normalize $CI to be set to 'true' if set and not 'false'
-[ "$CI" != 'false' ] && [ "$CI" != '' ] && CI='true'
+{ [ "$CI" = 'true' ] || [ "$CI" != 'false' ]; } && CI='true'
```
svatwork, line 17: This change broke my test suite. 

I would like for us to start running the `bats` tests as the amount of tests are growing. 

**Whenever someone makes a change to our bash plumbing it should be immediately visible whether the change introduces a bug.**

bin/otomi
```diff
@@ -68,7 +68,9 @@ uname -a | grep -i darwin >/dev/null && helm_config="$HOME/Library/Preferences/h
 function show_usage() {
   echo "otomi usage:
     bash                run interactive bash shell in otomi-core container
+    bats                runs BATS on bin/tests (see docs/BATS.md)
```
svatwork, line 71: You can agree whether we want this in our CLI or not. But this is a damn easy way to execute tests. The command is accessible to everyone and the whole setup is available through `otomi`.

bin/otomi
```diff
@@ -68,7 +68,9 @@ uname -a | grep -i darwin >/dev/null && helm_config="$HOME/Library/Preferences/h
...
     bootstrap           bootstrap values repo with artifacts corresponding to the cluster's stack version
+    check-policies      check if generated manifests adhere to defined OPA policies
```
svatwork, line 73: Alphabetical sorting, it was annoying the crap out of me. Can put it back.

bin/otomi
```diff
@@ -98,7 +99,7 @@ function show_usage() {
 function evaluate_k8s_context() {
   local env_path="$ENV_DIR/env/clouds/$CLOUD/$CLUSTER/.env"
   [ ! -f $env_path ] && err "The file '$env_path' does not exist" && exit 1
-  source $env_path
+  . $env_path
```
svatwork, line 102: More consistent: make a choice. `source` everywhere or `.` everywhere.

bin/otomi
```diff
@@ -130,11 +131,10 @@ function evaluate_secrets() {
...
   }
-  if [ -z "$CI" ]; then
-    local secrets_path="$ENV_DIR/.secrets"
-    [ ! -r "$secrets_path" ] && err "Unable to find the '$secrets_path' file.\nPlease follow to documentation: $readme_url" && exit 2
-    source $secrets_path
-  fi
+  local secrets_path="$ENV_DIR/.secrets"
+  [ ! -r "$secrets_path" ] && err "Unable to find the '$secrets_path' file.\nPlease follow to documentation: $readme_url" && exit 2
+  # shellcheck source=/dev/null
+  . $secrets_path
```
svatwork, line 137: The logic introduced here wasn't tested. I can understand that it's difficult to make changes to our bash plumbing without a test suite but **this logic puts the carriage before the horse.** 

I don't understand why it would only source secrets when in CI, because it is also a dependency for some commands that think `.secrets` is there. An enhancement would be to check if `.secrets` are present but I just put some bandages on the problem for now.

bin/otomi
```diff
@@ -244,6 +244,9 @@ function execute() {
       docker_terminal_params='-ti'
       drun bash
       ;;
+    bats)
+      drun bats -T "$@" bin/tests
```
svatwork, line 248: `-T` also outputs the time it took for the script to run.

bin/otomi
```diff
@@ -324,6 +331,11 @@ function execute() {
       git -C $ENV_DIR pull
       drun bin/bootstrap.sh 1
       ;;
+    rotate-keys)
+      check_sops_file
+      evaluate_secrets
+      drun bin/crypt.sh rotate
+      ;;
     sync)
```
svatwork, line 339: Alphabetical sorting.

bin/otomi
```diff
@@ -336,36 +348,27 @@ function execute() {
...
-    validate-templates)
+    values)
       evaluate_secrets
-      if [ "$*" != '' ]; then
-        validate_cluster_env
-      fi
-      drun bin/validate-templates.sh "$@"
+      validate_cluster_env
+      set -o pipefail
+      drun helmfile -e $CLOUD-$CLUSTER -f helmfile.tpl/helmfile-dump.yaml build | grep -Ev $helmfile_output_hide | sed -e $replace_paths_pattern
```
svatwork, line 371: Alphabetical sorting.

bin/tests/bats-common.sh
```diff
@@ -0,0 +1,7 @@
+#!/usr/local/env bash
+
+lib_dir="/usr/local/lib"
+
+load "$lib_dir/bats-support/load.bash"
+load "$lib_dir/bats-assert/load.bash"
+load "$lib_dir/bats-file/load.bash"
```
svatwork, line 7: Removes duplication

bin/tests/bootstrap.bats
```diff
@@ -1,24 +1,24 @@
...
-teardown () {
+function teardown () {
     temp_del "$test_temp_dir"
     unset ENV_DIR
     unset env_path
 }
+#######################
+# env folder creation #
+#######################
+env_folder="env folder"
```
svatwork, line 19: Removes duplication.

bin/tests/otomi.bats
```diff
@@ -0,0 +1,64 @@
...
+    assert_failure 2
+}
+
+##########################
+# bin/validate-templates #
+##########################
+generating_text="Generating k8s v1.18 manifests for cluster 'aws-dev'"
+assert_output_partial_generating_text="assert_output --partial $generating_text"
+validate_templates_name="validate-templates"
+run_otomi_validate_templates="run bin/otomi $validate_templates_name"
```
svatwork, line 33: Removes duplication.

bin/tests/bootstrap.bats
```diff
@@ -38,26 +38,26 @@ teardown () {
...
 }
 
-@test "executing bootstrap.sh should pass with new ENV_DIR (otomi-values) folder" {
+################
+# bootstrap.sh #
+################
+bootstrap_sh="bootstrap.sh"
+env_dir_str="with new ENV_DIR (otomi-values)"
+
+@test "$bootstrap_sh should pass $env_dir_str folder" {
```
svatwork, line 47: Removes duplication.

bin/validate-templates.sh
```diff
@@ -3,7 +3,7 @@
 [ "$CI" = 'true' ] && set -e
 set -o pipefail
 
-source bin/common.sh
+. bin/common.sh
```
svatwork, line 6: For consistency (see top comment)

bin/validate-templates.sh
```diff
@@ -109,14 +119,18 @@ function validate_templates() {
...
+    for_each_cluster validate_templates
+    exit 0
+  fi
+}
+
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
+  main "$@"
+  if [ $? -gt 0 ]; then
+    exit 1
+  fi
```
svatwork, line 135: Main plumbing.

There are 127 exit codes and we only use 2. I deliberately used some to `assert_failure $exit_code`. Correct me if I'm wrong.

docs/BATS.md
```diff
@@ -8,7 +8,7 @@ The otomi/tools:1.4.10 or newer includes the [bats framework](https://github.com
 
 Example of calling the binary:
 
-`docker run --rm otomi/core:latest bats bin/tests`
+`otomi bats`
```
svatwork, line 11: Updated documentation. This should make it much simpler.

bin/tests/common.bats
```diff
@@ -0,0 +1,40 @@
...
+@test "parse_args with ONLY -l should fail" {
+    run parse_args -l
+    assert_output --partial "option requires an argument -- 'l'"
+    assert_failure 2
+}
+
+@test "parse_args with --label 'arbitrary value' should pass" {
+    parse_args --label this=should_work
+    assert [ ${label} = 'this=should_work' ]
+}
```
svatwork, line 40: Brings the `parse_args` function under coverage.

svatwork (commented):
Some comments after review cleaning up a bit.

Morriz:
I will review this after getting an answer to this question:

Does this incorporate the requirements from the discussion? I just added to that discussion so I think this work is preliminary.

svatwork:
> Does this incorporate the requirements from the discussion?

No, not since your reply. But before that, I did my best to do so.

bin/otomi
svatwork: See if `otomi x bats` could be easier to not expose developer tooling.

Morriz: can we just use `otomi x bats ...` to achieve the same? if so then don't expose dev stuff please

bin/otomi
svatwork: Apparently there's a difference, will revert: https://unix.stackexchange.com/questions/309768/source-vs-why-different-behaviour.

svatwork (commented):
After stand-up conversation 25-2-2021

bin/otomi
Morriz: Why did you remove the check for CI?? It is there for a reason: if not in CI then source the secrets

Morriz: Please add it back but check for 'true'

bin/otomi
Morriz: nice!

bin/tests/bootstrap.bats
```diff
@@ -1,24 +1,24 @@
...
     unset ENV_DIR
     unset env_path
 }
+#######################
+# env folder creation #
+#######################
+env_folder="env folder"
 
-@test "the env folder should not be overwritten" {
+@test "$env_folder should not be overwritten" {
```
Morriz, line 21: why use a variable here? Please just use a meaningful word like `\$ENV_DIR/env`

bin/tests/common.bats
Morriz: awesome

bin/validate-templates.sh
```diff
@@ -109,14 +119,18 @@ function validate_templates() {
...
-  # hf lint
-else
-  [ -n "$VERBOSE" ] && echo "Running validate-templates for all clusters"
-  for_each_cluster validate_templates
-  # re-enable next line after helm does not throw error any more: https://github.com/helm/helm/issues/8596
-  # for_each_cluster hf lint
+function main() {
+  parse_args "$@"
+  [[ $all && $label ]] && echo "Error: cannot specify --all and --label simultaneously" && exit 6
+  if [[ $all = 'y' || $label ]]; then
```
Morriz, line 125: its just `--all` right? so why check for `y`?

bin/validate-templates.sh
Morriz: ok, but is that consistent? Confuses me to assign numbers without meaning. Where are these codes documented?

docs/BATS.md
Morriz: `otomi x bats` ?

Morriz (changes requested):
Nice work. Please revert the bracket testing stuff and address the comments.

bin/tests/bootstrap.bats
svatwork: It is to avoid duplication, this test starts with the same name as the other one -- easy to change.

bin/validate-templates.sh
svatwork: `parse_args` will set `all='y'` if `--all` is passed. This will provide a "module"-like functionality in `common.sh`. But I could also set it to `'true'` or `1`, what do you prefer?

svatwork: I thought we agreed on `'true'` and `'false'` but can't find it easily in the style guide..

Dunky13: In some places we use `"true"` and in other places it is `true`
So even that style is not consistent

bin/validate-templates.sh
svatwork: I don't want to suggest it is a panacea or perfect solution, no.

I provided exit codes documentation in the function comment for `parse_args`. I can imagine that if someone checks for the exit code of the script it might be confusing. So I don't know what a good solution there would be.

But for `bats` tests it is quite nice, as one can `assert_failure $exit_code`.

svatwork: Here is another idea for standardisation: https://stackoverflow.com/a/24121322.

bin/common.sh
```diff
@@ -128,9 +127,70 @@ function for_each_cluster() {
...
+#     STDERR 3: no '--' passed with options
+#     STDERR 4: special ERR reserved for checking if enhanced getopt is present on the host
+#     STDERR 5: no options specified
+# Returns:
+#     all -> if passed, sets to 'y' and can be used globally in conditional statements
+#     label -> if passed (e.g. label init=true), sets to label (e.g. 'init=true') and
+#              can be used globally in conditional statements
+# Resources:
+# - https://stackoverflow.com/a/29754866
+##############################################################################################
```
svatwork, line 151: This may look inconsistent but it is a function comment and I followed the style guide for that (https://github.com/google/styleguide/blob/gh-pages/shellguide.md#function-comments).

bin/validate-templates.sh
Morriz: If we want a consistent testing approach that can handle both booleans as well as strings, and looking at your below exmple `[[ $all && $label ]]` that uses that, I would stick with the strings.

Morriz: so use 'true' and 'false'

bin/common.sh
Morriz: Read that and I favor single brackets as they are more DRY and should just work

svatwork: Let's override the style guide in that case. Do you agree with the style guide then that in the following cases `[[` is preferred:
- pathname expansion
- word splitting
- regex matching, e.g: `[ ab = a? ]: a? glob expands. So may be true or false depending on the files in the current directory.` -> is unreliable, so have to use `[[ ab = a? ]]: true, because it does pattern matching (* ? [ are magic). Does not glob expand to files in current directory.` 
- `[[ ab =~ ab? ]]` because `[ a =~ a ]: syntax error. No bash equivalent.`
- logical AND-s, e.g. `[[ a = a && b = b ]]` (or do we prefer `[ a=a ] && [ b = b]`?)
- `x='a b'; [ $x = 'a b' ]: syntax error, expands to [ a b = 'a b' ]` so we have to use `x='a b'; [[ $x = 'a b' ]]: true, quotes not needed`

`[` in all other cases.

svatwork: Also in our conversation today we wanted to keep the Bash version consistent, here is mine:

```
bash --version
GNU bash, version 5.0.17(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2019 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

If you can send me yours then I will install that one and we check for something like `[ $(bash -c --version) = vX.Y.Z.`

Dunky13:
```sh
bash --version
GNU bash, version 5.1.4(1)-release (x86_64-apple-darwin20.2.0)
Copyright (C) 2020 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>
```

docs/BATS.md
Dunky13: https://github.com/redkubes/otomi-core/pull/330/files#r581026346

By adding `otomi bats` as shown in the above comment (that you liked) Sebastiaan is able to specify what & how it's tested and outputted. If we go `otomi x bats` then we miss the `-T` and `bin/tests` parameters, and only the `$@` is passed to bats

Dunky13 (approved):

docs/BATS.md
svatwork: Yes @Dunky13 that's right. I want developers to have immediate feedback at the tip of the fingers.

I also see the possibility for a --dev flag (that's easy to add now) as a compromise (e.g. in `~/.zshrc`: `alias otomi="otomi --dev"`

bin/validate-templates.sh
svatwork: I think I have a better compromise. I improved the `err` function to be a bit more verbose.

bin/common.sh
Morriz: Good idea. One thing: can we give this function a meaningful name and NOT put it in `common.sh`? I suggest we instead start another file `bin/functions.sh` that holds all the shared functions that are NOT used by `bin/otomi`.

docs/BATS.md
Morriz: What is wrong with passing extra flags to `otomi x bats`? Why would this:

> then we miss the -T and bin/tests parameters, and only the $@ is passed to bats

be the case?

bin/common.sh
Morriz: We just settle for the one that is most prevalent. If that is 5 or 4 I don't care. No need to show you mine ;p

Morriz (changes requested):
@svatwork please finish the last comments next workday, with focus on:
- brackets being DRY
- BASH v4 or 5
- moving shared functions that are invoked in a container from `bin/common.sh` to `bin/functions.sh`

so we can merge this and benefit :)

svatwork:
> please finish the last comments next workday, with focus on:

I don't know why you think I wasn't aware of the issues that need addressing...

Still need to add the extra features from the discussion. I will push what I have now, if you want to make an enhancement that's fine too (not sure what you want from me now).

bin/common.sh
```diff
@@ -38,11 +48,20 @@ function abort() {
...
 trap abort SIGINT
 
-#######################################
+#####
 # https://github.com/google/styleguide/blob/gh-pages/shellguide.md#stdout-vs-stderr
-#######################################
+#####
 function err() {
-  echo "[$(date +'%Y-%m-%dT %T.%3N')] ERROR: $*" >&2
+  local tab=$'\t'
```
svatwork, line 55: I propose to extend this error message. I think it could really help with debugging (thanks @Dunky13 for the set-up). It will look like this e.g.:

```
otomi bats
Error: unknown command "r" for "yq"
Run 'yq --help' for usage.
Error: unknown command "r" for "yq"
Run 'yq --help' for usage.
bin/otomi: line 235: bats: command not found
---------------------------                        --------------------------- 
- Time                                             [2021-03-02T 08:27:28.456]
- Faulty script                                    bin/otomi
- Last function                                    drun
- Last arguments                                   $command
- Script message (if any)                          common.sh FAILED
```

bin/common.sh
```diff
@@ -65,9 +84,9 @@ function _rind() {
   fi
 }
 
-#######################################
-# https://github.com/google/styleguide/blob/gh-pages/shellguide.md#quoting
-#######################################
+#####
```
svatwork, line 87: I propose we should agree on the size of the function comments and add it to the style guide. 5 seems enough so one doesn't have to think about it a lot and it's still visible when skimming a script.

docs/BATS.md
svatwork:
```
    bats)
      drun bats -T "$@" bin/tests
      ;;
```

I could be wrong, but running the load in a container would give us a more consistent environment?

svatwork: Yes, actually, I don't think that we would miss functionality. You're right about that.

It's just a difference in naming the command. I think it's more convenient to have the shortcut across all dev environments. I think `otomi bats` is easier to remember than a second-class citizen like `otomi x bats -T bin/tests`

bin/common.sh
Morriz: 3 is also fine with me then...but I think ppl will just copy paste

bin/otomi
```diff
@@ -24,7 +24,7 @@ ENV_DIR=${ENV_DIR:-$PWD}
 
 . bin/common.sh
 
-[ $has_docker = 'false' ] && [[ ! "$CI" ]] && err "Otomi CLI expects a running docker daemon!" && exit 1
+[ $has_docker = 'false' ] && [ -n "$CI" ] && err "Otomi CLI expects a running docker daemon!" && exit 1
```
Morriz, line 27: now that I think of this, I think this second check should test for IN_DOCKER to be valuable, because even in CI we need docker to run

bin/otomi
```diff
@@ -131,19 +132,21 @@ function evaluate_secrets() {
     [ -n "$VERBOSE" ] && echo "Info: The 'secrets.*.yaml files' are not decrypted, because $ENV_DIR/.sops.yaml file is not present"
     return 0
   }
-  local secrets_path="$ENV_DIR/.secrets"
-  [ ! -r "$secrets_path" ] && err "Unable to find the '$secrets_path' file.\nPlease follow to documentation: $readme_url" && exit 2
-  # shellcheck source=/dev/null
-  . $secrets_path
+  if [ "$CI" != 'true' ]; then
```
Morriz, line 135: did we not go for `-z "CI"`?

bin/tests/otomi.bats
```diff
@@ -21,7 +21,8 @@ function teardown () {
 #############
 @test "otomi without arguments prints help message" {
     run bin/otomi
-    assert_failure 2
+    # Exit code is different depending on global $CI ...
```
Morriz, line 24: may I ask why?

Morriz (changes requested):
small potatoes left...sorry for the messy review...

bin/otomi
svatwork: Sorry, I don't fully understand what you mean by that...

bin/tests/otomi.bats
svatwork: At the moment, this line is caught by the CI runner and exits 1:

```
[ $has_docker = 'false' ] && [ -z "$CI" ] && err "Otomi CLI expects a running docker daemon!" && exit 1
```

bin/common.sh
```diff
@@ -128,9 +133,9 @@ function for_each_cluster() {
...
 }
 
-hf_templates_init() {
+function hf_templates_init() {
   local out_dir="$1"
   shift
-  [ -z "$*" ] && hf -f helmfile.tpl/helmfile-init.yaml template --skip-deps --output-dir="$out_dir" >/dev/null 2>&1
-  hf "$@" template --skip-deps --output-dir="$out_dir" >/dev/null 2>&1
-}
+  [ -n "$all" ] && hf -f helmfile.tpl/helmfile-init.yaml template --skip-deps --output-dir="$out_dir" >/dev/null 2>&1
```
svatwork, line 139: Hmm.. actually not too sure whether we should just leave `[ -z "$*" ]` here. Any input?

svatwork:
I responded to the discussion with a question about `configure`: https://github.com/redkubes/otomi-core/discussions/310#discussioncomment-424371.

bin/otomi
```diff
@@ -133,9 +140,10 @@ function evaluate_secrets() {
   if [ -z "$CI" ]; then
     local secrets_path="$ENV_DIR/.secrets"
     [ ! -r "$secrets_path" ] && err "Unable to find the '$secrets_path' file.\nPlease follow to documentation: $readme_url" && exit 2
-    source $secrets_path
+    . $secrets_path
   fi
   prepare_crypt
+  set +x
```
j-zimnowoda, line 146: I guess not needed

j-zimnowoda (commented):
Please merge with master

bin/common.sh
Morriz: I believe Alin added that to exclude that line when labels are given, but it never really worked I think. Anyways, that should be solved by this PR.

bin/otomi
```diff
@@ -68,46 +68,53 @@ uname -a | grep -i darwin >/dev/null && helm_config="$HOME/Library/Preferences/h
...
 
   Env flags:
-    VERBOSE=1;          Run otomi CLI in verbose mode
     CI=1                Exit the script after first error
+    DEBUG=1             Run otomi CLI in debug mode
+    VERBOSE=1;          Run otomi CLI in verbose mode
+
+  Options:
+    -A|--all            Target all charts (not simultaneously with -l|--label)     
+    -c|--cluster        Target clusters (format: CLOUD-CLUSTER), e.g. '-c aws-dev' or '--cluster azure-demo'
```
Morriz, line 100:
```suggestion
    -c|--cluster        Override a set target cluster (format: CLOUD-CLUSTER), e.g. '-c aws-dev' or '--cluster azure-demo'
```

bin/otomi
```diff
@@ -68,46 +68,53 @@ uname -a | grep -i darwin >/dev/null && helm_config="$HOME/Library/Preferences/h
...
     x                   execute command in container
 
   Env flags:
-    VERBOSE=1;          Run otomi CLI in verbose mode
     CI=1                Exit the script after first error
+    DEBUG=1             Run otomi CLI in debug mode
+    VERBOSE=1;          Run otomi CLI in verbose mode
+
+  Options:
+    -A|--all            Target all charts (not simultaneously with -l|--label)
```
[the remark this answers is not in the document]
Morriz: Ok, I think we should go back to the drawing board.

bin/common.sh
svatwork: `parse_args` is not meaningful? It parses args? Or parses opts (options)?

bin/otomi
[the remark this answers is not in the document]
Morriz:
```suggestion
    -A|--all            Target all clusters
```

bin/common.sh
svatwork: If this is the use-case, I will retrofit it to exclude that line when labels are given.

bin/otomi
svatwork: Oops.

svatwork:
To do after conversation with @Morriz:
- Refactor `validate-templates`, `check-policies`, `validate-values` to use `getopt`
- Add `otomi configure --cluster`, this sets a config file in `otomi-values` and exports the variable to set target context of cluster.

Morriz:
> * Add `otomi configure --cluster`, this sets a config file in `otomi-values` and exports the variable to set target context of cluster.

Great. Can we call that file `otomi.cfg` and make it use ini format?

svatwork:
> Great. Can we call that file otomi.cfg

Yes

> ini format

I think `ini` parsing is a bit more nuanced than that: https://stackoverflow.com/questions/16987648/update-var-in-ini-file-using-bash.

Let's stick with exporting vars for now?

Morriz:
> > ini format
> 
> I think `ini` parsing is a bit more nuanced than that: https://stackoverflow.com/questions/16987648/update-var-in-ini-file-using-bash.
> 
> Let's stick with exporting vars for now?

Ok, so no `otomi configure` for now then? (As export does not put vars into parent process.) That is fine.

Morriz:
I see conflicts reported here

Morriz:
I think this needs merging with master

svatwork:
> Ok, so no otomi configure for now then? (As export does not put vars into parent process.) That is fine.

I think we can still do it but not in the `ini` format. I've proposed a solution.

bin/check-policies.sh
```diff
@@ -34,20 +37,21 @@ check_policies() {
...
 
-if [ "${1-}" != "" ]; then
-  echo "Checking policies for cluster '$(cluster_env)'"
-  shift
-  check_policies "$@"
-else
-  echo "Checking policies for all clusters"
-  for_each_cluster check_policies
+function main() {
+  validate_resources check_policies "$@"
```
svatwork, line 49: This is the main refactor of `check_policies`, `validate-templates` and `validate-values`. I am actually quite satisfied with the solution. They all have the same pattern actually so this removes a lot of duplication and creates opportunity for re-use.

bin/common-modules.sh
```diff
@@ -0,0 +1,72 @@
...
+    # - pass arguments only via   -- "$@"   to separate them correctly
+    ! PARSED=$(getopt --options=$OPTIONS --longoptions=$LONGOPTS --name "$0" -- "$@")
+    if [[ ${PIPESTATUS[0]} -ne 0 ]]; then
+      exit 1
+    fi
+    eval set -- "$PARSED"
+    while true; do
+      case "$1" in
+        -A | --all)
+          ALL_OPT='true'
```
svatwork, line 32: I chose to make more explicit that it's an OPT(ION) and they are of course global.

j-zimnowoda (changes requested):
Few observations from running CLI:
1. `otomi validate-values -l name=test` should not be allowed
2. Performing command for not existing cluster shall rise error. Instead:
```
otomi validate-templates -c demooooo                                                                                                           
Generating k8s v1.18 manifests for cluster 'demooooo'...
Processing CRD files...
find: '/tmp/otomi/generated-manifests/v1.18': No such file or directory
Validating resources for cluster 'demooooo'...
Templates validation SUCCEEDED
common.sh SUCCEEDED
```
3. The `otomi values` appends `common.sh SUCCEEDED` string to rendered values, what make it hard to parse output as YAML

bin/common-modules.sh
```diff
@@ -0,0 +1,72 @@
...
+#####
+# Use OPTIONS/LONGOPTS(LONGOPTIONS) to set additional parameters.
+# Resources:
+# - https://github.com/google/styleguide/blob/gh-pages/shellguide.md#s4.2-function-comments
+# - https://stackoverflow.com/a/29754866
+#####
+function parse_args() {
+  if [[ "$*" != "" ]]; then
+    ! getopt --test >/dev/null
+    if [[ ${PIPESTATUS[0]} -ne 4 ]]; then
```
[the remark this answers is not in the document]
svatwork: I understand your concern, but let's keep that out of this pull request... 

It's just checking whether enhanced `getopt` is available on the machine.

svatwork:
> otomi validate-values -l name=test should not be allowed

Why not? It is a valid name for a manifest. 

> Performing command for not existing cluster shall rise error. Instead:

This is a good catch. I see `find` should've exited, but by some weird logic there is no `set -e` for this part.... If we do it, the whole build will break I think. Opinions?

bin/common-modules.sh
```diff
@@ -0,0 +1,72 @@
...
+  parse_args "$@"
+  [ -n "$ALL_OPT" ] && [ -n "$CLUSTER_OPT" ] && err "cannot specify --all and --cluster simultaneously" && exit 1
+  if [ -n "$ALL_OPT" ]; then
+    for_each_cluster $cmd
+    exit 0
+  else
+    $cmd
+    exit 0
+  fi
+  exit 1
```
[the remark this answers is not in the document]
svatwork: You never know with `bash`...

bin/common-modules.sh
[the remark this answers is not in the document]
svatwork: I see your point, but let's leave that discussion out.

svatwork:
> otomi validate-values -l name=test should not be allowed

I agree, see latest commit.

> Performing command for not existing cluster shall rise error. Instead:

I think that this code shouldn't be coupled to a list of clusters. I agree we should improve error handling, see next reply:

> The otomi values appends common.sh SUCCEEDED string to rendered values, what make it hard to parse output as YAML

The error handling isn't a one-size-fits-all solution at the moment, I agree we can improve it but I don't see it as part of the scope of this pr.

docs/BATS.md
Morriz: Maybe we should expose dev commands only when one has that role? Or put them not in the docs if that role is not set? Should we introduce a flag `DEV`?

Then we have a straightforward approach to add dev commands and only show them when `DEV` is set

Morriz (changes requested):
Quite some things to look into I think. One addition:

`[ -n "$CI" ] && set -e` is not used in master anymore I believe, as we want to force errors on exit with `set -e` after sourcing `common.sh`

bin/check-policies.sh
```diff
@@ -1,7 +1,9 @@
 #!/usr/bin/env bash
 
+[ -n "$CI" ] && set -e
```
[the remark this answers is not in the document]
svatwork: > yesterdays change that favors set -e always

What do you mean? I am a bit confused by this statement.

bin/common-modules.sh
```diff
@@ -0,0 +1,71 @@
...
+function parse_args() {
+  if [[ "$*" != "" ]]; then
+    ! getopt --test >/dev/null
+    if [[ ${PIPESTATUS[0]} -ne 4 ]]; then
+      err '`getopt --test` failed in this environment.'
+      exit 1
+    fi
+
+    OPTIONS=Al:c:
+    LONGOPTS=all,label:,cluster:
```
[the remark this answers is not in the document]
svatwork: You mean to override the style guide thusly? https://github.com/google/styleguide/blob/gh-pages/shellguide.md#constants-and-environment-variable-names

bin/common.sh
```diff
@@ -18,13 +20,17 @@ fi
 
 # some exit handling for scripts to clean up
 exitcode=0
-script_message=''
+script_message='common.sh'
```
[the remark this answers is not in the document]
svatwork: My intention was to throw in this function as well... But now that I think about it and follow your reasoning, that might be unnecessary.

bin/common.sh
```diff
@@ -18,13 +20,17 @@ fi
...
   [ $x -ne 0 ] && exitcode=$x
-  [ "$script_message" != '' ] && ([ $exitcode -eq 0 ] && echo "$script_message SUCCESS" || err "$script_message FAILED")
+  if [ $exitcode -eq 0 ]; then
+    echo "$script_message SUCCEEDED"
+  else
+    err "$script_message FAILED"
+  fi
   cleanup
-  trap "exit $exitcode" EXIT ERR
+  trap 'exit $exitcode' EXIT ERR
```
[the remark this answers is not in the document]
svatwork: That is true, my intention was to solve https://github.com/koalaman/shellcheck/wiki/SC2064.

bin/otomi
```diff
@@ -68,46 +68,54 @@ uname -a | grep -i darwin >/dev/null && helm_config="$HOME/Library/Preferences/h
...
+    check-policies      check if generated manifests adhere to defined OPA policies
+    commit              execute wrapper for encrypt -> generate pipelines -> git commit changed files
+    configure           set environment variables in otomi-values/otomi.cfg (ini format). Current location: $ENV_DIR/otomi.cfg
+    console             start Otomi Console (add any arg to quit) on http://127.0.0.1:3000 (licensed, so pull secret required for api)
     decrypt             decrypts file(s) given as arguments (relative to env folder), or all env/*.secrets.yaml to env/*.secrets.yaml.dec files
     deploy              execute otomi-core deploy script
-    destroy             destroy all or some charts (use labels like '-l name=prometheus-operator')
-    diff                diff k8s resources (use labels like '-l name=prometheus-operator')
-    encrypt             encrypts file(s) given as arguments (relative to env folder), or all env/*.secrets.yaml files
+    destroy             destroy all or some charts
```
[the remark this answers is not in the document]
svatwork: I don't know this function very well, but I see it does take $CLOUD-$CLUSTER. However, I find this comment ambiguous with the options also defined in the help message... Do you have a suggestion for a compromise?

bin/tests/validate-templates.bats
```diff
@@ -0,0 +1,76 @@
...
+
+@test "$validate_templates_name --cluster aws-demo starts generating 'aws-demo'" {
+    eval "$run_otomi_validate_templates --cluster $aws_demo_str"
+    eval "$assert_output_partial_generating_text $generating_text $aws_demo_str"
+}
+
+@test "$validate_templates_name -c aws-demo starts generating 'aws-demo'" {
+    eval "$run_otomi_validate_templates -c $aws_demo_str"
+    eval "$assert_output_partial_generating_text $generating_text $aws_demo_str"
+}
```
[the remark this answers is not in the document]
svatwork: That's not true? Every test tests `bin/validate-templates.sh` with zero or more options. There is a test for `parse_args`.

However, I agree that testing whether a script outputs a string is very fragile. I have been pondering that over the weekend, but I think the code to test is poorly designed. We would have to refactor the methods to make them testable. Do you want to spend more time on this now?

bin/tests/validate-values.bats
```diff
@@ -0,0 +1,30 @@
...
+@test "$validate_values_str starts generating with --cluster azure-demo" {
+    eval "$run_otomi_validate_values --cluster azure-demo"
+    assert_output --partial "/tmp/otomi/values/azure-demo.yaml passed test"
+}
+
+@test "$validate_values_str exit with label" {
+    eval "$run_otomi_validate_values --label something=something"
+    assert_output --partial "Cannot pass option something=something: please specify --all|-A or --cluster|-c"
+    assert_failure 1
+}
```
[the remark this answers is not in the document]
svatwork: Same

svatwork:
>  Should we introduce a flag DEV

Yes, let's do that?

> Quite some things to look into I think.

Yes, that's what happens with scope creep... It started with: add a one-liner to enforce --label and --all...

bin/validate-values.sh
```diff
@@ -17,15 +18,21 @@ function cleanup() {
...
   return 0
 }
 
-if [ -n "$1" ]; then
-  validate_values
-else
-  for_each_cluster validate_values
+function main() {
+  validate_resources validate_values "$@"
+}
```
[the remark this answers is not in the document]
redkubesbot: ah, ok...is `validate_resources` really  validating resources? Can we call it something like `process_clusters`?

bin/tests/validate-templates.bats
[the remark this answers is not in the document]
redkubesbot: Look, it seems to me you are testing them here all over again. These super expensive tests are not needed. Only test these 2 to get coverage of this scripts code path
- one target (with one chart selector to limit scope and duration)
- all targets (with one chart selector to limit scope and duration)

I think that will cover most.

docs/BATS.md
svatwork: Yes, that sounds like a great idea.

I would suggest to add a task as an enhancement, a "dev" mode, because otherwise we will keep this pr open for too long.

bin/common.sh
```diff
@@ -18,13 +20,17 @@ fi
...
+script_message='common.sh'
 function exit_handler() {
   local x=$?
   [ $x -ne 0 ] && exitcode=$x
-  [ "$script_message" != '' ] && ([ $exitcode -eq 0 ] && echo "$script_message SUCCESS" || err "$script_message FAILED")
+  if [ $exitcode -eq 0 ]; then
+    echo "$script_message SUCCEEDED"
+  else
+    err "$script_message FAILED"
+  fi
```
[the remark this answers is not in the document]
svatwork: I just reverted it. I will just assume it works correctly.

bin/check-policies.sh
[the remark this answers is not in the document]
Morriz: if you check master you can see that the `bin/*` scripts now have another setting on top, to be sure that they exit on error immediately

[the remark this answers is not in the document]
svatwork: That's good.

bin/gen-schemas.sh
```diff
@@ -1,5 +1,7 @@
 #!/bin/bash -xe
 
+set -e
```
Morriz, line 3: Why are you now blindly setting this flag everywhere? Maybe not all scripts are ok with that. Did you test them all?

Please be aware that some scripts might invoke failing commands to test something.

bin/prepare-gke-cluster.sh
```diff
@@ -1,5 +1,7 @@
 #!/usr/bin/env bash
 
+set -e
```
Morriz, line 3: this script may be removed with this PR

bin/tests/otomi.bats
Morriz: Ah, ok. But this test title suggests a message is to be seen. Is that tested for? It seems it only asserts an exit with a code of 1 or 2

bin/validate-templates.sh
```diff
@@ -1,8 +1,11 @@
 #!/usr/bin/env bash
 
-source bin/common.sh
+set -e
 set -o pipefail
```
Morriz, line 4: we combine setting flags in one line

Morriz (changes requested):
were still movinggg

bin/gen-schemas.sh
svatwork: Sorry, I misunderstood your earlier comment.

svatwork:
#290 is the issue that will be closed by this pr, I don't understand why it's been closed already.

bin/tests/otomi.bats
Morriz: no answer here? lets move on then...

bin/validate-values.sh
```diff
@@ -17,15 +17,19 @@ function cleanup() {
...
-else
-  for_each_cluster validate_values
+function main() {
+  process_clusters validate_values "$@"
+}
+
+main "$@"
+if [ $? -gt 0 ]; then
+  exit 1
 fi
```
Morriz, line 35: with set -e this will never be needed, can we remove those?

Morriz (changes requested):
last thing...

plus: can you rename `hf_templates_init` back to `hf_templates`? It is like that now in master.

bin/check-policies.sh
```diff
@@ -34,20 +37,21 @@ check_policies() {
...
-  shift
-  check_policies "$@"
-else
-  echo "Checking policies for all clusters"
-  for_each_cluster check_policies
+function main() {
+  process_clusters check_policies "$@"
+}
+
+if [[ "${BASH_SOURCE[0]}" == "${0}" ]]; then
```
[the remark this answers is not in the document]
Dunky13: https://stackoverflow.com/a/2684300/14982291

What this check does it check whether the script is being `source`d or being ran directly.
The difference impacts which variables are available and which are not.
So yes, this check is false when the script is sourced.

Morriz:
> #290 is the issue that will be closed by this pr, I don't understand why it's been closed already.

Sorry, thought it was a duplicate item. Have to get used to the outstanding Tasks. Reopenend

Morriz (approved):

j-zimnowoda (changes requested):
Unfortunately I cannot prove it, because `otomi validate-template` with `-l` argument fails whereas without this argument it succeed.

Would you be so kind a take a look at this issue?

```
➜  otomi validate-templates              
Generating k8s v1.18 manifests for cluster 'aws-dev'...
Processing CRD files...
Validating resources for cluster 'aws-dev'...
WARN - knative-init/templates/all.yaml containing a CustomResourceDefinition (knativeeventings.operator.knative.dev) was not validated against a schema
WARN - knative-init/templates/all.yaml containing a CustomResourceDefinition (knativeservings.operator.knative.dev) was not validated against a schema
WARN - kubeapps/templates/apprepositories.yaml containing a AppRepository (bitnami) was not validated against a schema
WARN - team-ns/templates/external-secrets.yaml contains an invalid ExternalSecret (otomi-welcome) - spec.data.0.key: Invalid type. Expected: string, given: null
WARN - team-ns/templates/external-secrets.yaml contains an invalid ExternalSecret (otomi-welcome) - spec.data.0.key: Invalid type. Expected: string, given: null
WARN - team-ns/templates/external-secrets.yaml contains an invalid ExternalSecret (otomi-welcome) - spec.data.0.key: Invalid type. Expected: string, given: null
Templates validation SUCCESS
➜  otomi validate-templates -l=name=keycloak                                                      (⎈ otomi-eks-dev:default)
Generating k8s v1.18 manifests for cluster 'aws-dev'...
[2021-03-15T 11:34:13.738] ERROR: Templates validation FAILED
```

Morriz:
@j-zimnowoda you are passing the wrong params to helmfile ;p

j-zimnowoda:
> @j-zimnowoda you are passing the wrong params to helmfile ;p
> I see you asked for changes, but I can't see which. Can you remove your changes request if you have nothing left?

See comment above. 
> Unfortunately I cannot prove it, because ...

It am not asking to change a specific line but I am asking about functionality that does not seem to work.

Morriz:
@j-zimnowoda I deleted your comment by accident. I did not think I can. Anyway, please check my answer to your perceived "not working" above.

PR #336, opened 2021-03-05
Add slack notify workflow

svatwork:
Let's add some visibility into our builds. What do you think team?

If so, please add the secret for `SLACK_WEBHOOK: ${{ secrets.SLACK_WEBHOOK }}`: https://github.com/marketplace/actions/slack-notify.

I've set it to a newly created `github-ci` channel to not pollute other channels.

### DIFF
--- a/.github/workflows/slack-notify.yml
+++ b/.github/workflows/slack-notify.yml
@@ -1,0 +1,13 @@
+on: push
+name: Slack Notification
+jobs:
+  slackNotification:
+    name: Slack Notification
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


svatwork:
Let's see if we can do it for the whole organization: https://docs.github.com/en/actions/learn-github-actions/sharing-workflows-with-your-organization.

Morriz (approved):
awesome

svatwork:
The secret doesn't isn't set... And I was waiting for your credentials to make the workflow for all repositories. So just to avoid errors, I will close this now.

Morriz:
Why close this when it is finished? SLACK_WEBHOOK was just added.

Issue #383, opened 2021-03-29
otomi cli calls run twice


Morriz:
**Describe the bug:** (a clear and concise description of what the bug is)

When running otomi cli calls they seem to get called twice. When running the following I see diff output twice:

`DEBUG=1 otomi diff -l name=loki`

svatwork:
> DEBUG=1 otomi diff -l name=loki

Looks like `DEBUG`s meaning in the context of `bin/otomi` has changed to TRACE.

Dunky13:
I checked out the master branch, and pulled the latest code. 
For testing purposes, I changed the Gitea version number from `1.13.4` to `1.13.6` 
Then I both: `DEBUG=1 otomi diff -l name=gitea` and `TRACE=1 otomi diff -l name=gitea`

In both cases the output was the same and I did not find the code running, or seeming to run twice.

If this is still a bug, that I am not aware of, please amend the description, so it can be more easily reproduced. As for now it seems this issue does no longer exist in `master`

Morriz:
closing then...tnx

PR #392, opened 2021-04-07
bin/otomi and bin/common.sh conditionals

svatwork:
#382 broke an execution path I rely on for my own Docker image.

I also saw `$has_docker` two times, curious why that's necessary?


Dunky13 (approved):

svatwork:
This might fix #383, what do you think?

Morriz (approved):

PR #422, opened 2021-05-07
fix: in azureMonitor you need to provide either both of dlientId and …

mojtabaimani:
In value-schema, in azureMonitor you need to provide either both of dlientId and clientSecret together or non of them.


j-zimnowoda (changes requested):
It is conform with json schema bit may not be with OAS 3.0
Also Please ensure that this can be rendered by https://rjsf-team.github.io/react-jsonschema-form/

j-zimnowoda (changes requested):
Please ensure that this schema can be renderd as react form: https://rjsf-team.github.io/react-jsonschema-form/

Also check if 'dependencies' are conform with OAS 3.0

values-schema.yaml
```diff
@@ -64,14 +64,20 @@ definitions:
...
         type: string
       appInsightsAppId:
         description: An Azure AppInsights client id (defaults to clientId).
         type: string
+      clientId:
+        description: An Azure client id.
+        type: string
+      clientSecret:
+        description: An Azure client secret.
+        type: string
```
Morriz, line 80: where is `tenantId`? How did that get lost @j-zimnowoda (I am asking you first after the single-repo refactor)? This was coming from the `azure.tenantId` prop. Where did the root `azure` prop go? And how was the refactoring done of the templates that relied on those?

Morriz (changes requested):

Morriz (approved):

PR #423, opened 2021-05-07
fix(bin/otomi): command is echo'd instead of executed

bartusz01:
environment variable is set with functionname instead of executing the function


Morriz:
closed this as it was a oneliner and decision was made to do it differently

PR #428, opened 2021-05-20
KMS schema is not valid JSON Schema/OpenAPI schema

svatwork:
I created this pull request because the `otomi-api` schema will rely on this schema. @j-zimnowoda made it clear that we don't want divergence in the schemas and I agree of course.

The story: The `openapi-generator` was complaining about this schema. At first I assumed that this schema is valid JSON schema but not OpenAPI schema, but I can't find anything about the previous modification being valid any schema.

An `oneOf` takes an array of subschemas, and not an array of objects. It is not possible to use `$ref` anywhere one likes. "`$ref` can only be used in place of a subschema."

If this breaks anything, how can we integrate it?

### DIFF
--- a/profiles/aws/env/secrets.settings.yaml
+++ b/profiles/aws/env/secrets.settings.yaml
@@ -12,7 +12,7 @@ home:
 kms:
     sops:
         aws:
-            clientKey: somesecretvalue
+            accessKey: somesecretvalue
             secretKey: somesecretvalue
             region: somesecretvalue
 oidc:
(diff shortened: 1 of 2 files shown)


Morriz (changes requested):
I think we are missing a discriminator prop now. If you add "provider" to each creds definition we can determine what is given.

svatwork:
It seems discriminator is an OpenAPI concept but not in JSON Schema (at least not draft-07).

Morriz (approved):

Issue #353, opened 2021-03-10
update otomi with workaround for docker run in linux


Morriz:
See last comments

Dunky13:
Will research this on my linux machine to see if this is an issue or a non-issue

svatwork:
I bake my own container image and don't use the `app` user, works fine now.

Morriz:
Can you provide more details for others here to help them?

svatwork:
For development purposes:

1. Remove the following from the tools Docker:

```
RUN chown -R app:app /home/app
USER app
```

2. Docker build the tools image, e.g.

```bash
 docker build -t otomi/tools:my-tools -e NPM_TOKEN=$NPM_TOKEN . 
```

3. Replace the `otomi-core` Dockerfile's referred image to the built tools image:

```
FROM otomi/tools:my-tools as test
FROM otomi/tools:my-tools as prod
```

4. Replace every occurrence of `--chown=app` from the root Dockerfile's source code:

```bash
COPY --chown=app . .
```

with 

```bash
COPY . .
```

5. Build the `otomi-core` image, e.g.

```bash
docker build -t otomi/core:my-core .
``` 

6. Replace `otomiVersion` in `otomi-values` with core image, e.g.:

```yaml
clouds:
    aws:
        domain: eks.otomi.cloud
        clusters:
            dev:
                enabled: true
                apiName: eks_otomi-cloud_eu-central-1_otomi-eks-dev
                apiServer: BFE1BC8655F2C73189DCE38F5E40D0AC.gr7.eu-central-1.eks.amazonaws.com
                k8sVersion: '1.19'
                otomiVersion: my-core
```

Morriz:
This effectively makes it use the base image user, which is root. Not wanted.

svatwork:
Then why ask for it?

Morriz:
assigning to Sebastiaan as he is into this. Don't reassign!

svatwork:
There is a project that changes ownership on the fly, basically what OSXFS does for Docker for Mac: https://github.com/tianon/gosu. 

For reference, this post: https://stackoverflow.com/a/54787364/8357826.

<details>
<summary>Summary</summary>

For others that see this issue with containers running as a different user, you need to ensure the uid/gid of the user inside the container has permissions to the file on the host. On production servers, this is often done by controlling the uid/gid in the image build process to match a uid/gid on the host that has access to the files (or even better, do not use host mounts in production).

[...]

PR #430, opened 2021-05-26
Define option to set custom image

svatwork:
I was working on #353 (because I have to test building my own image) and this would help me with development, by not having to set the image in `cluster.yaml` but able to override it with an option. 

BTW, line 55-97 auto-formatted for some reason to the left...


Dunky13 (approved):
Looks good

Morriz (changes requested):
I think you don't need this as you can locally build the image name with a custom tag, and just reference that from the values.

Can you close this PR yourself?

PR #431, opened 2021-05-26
feat: otomi install chart skeleton 

mojtabaimani:
This is not ready just for getting your comments.


otomi-install/Chart.yaml
```diff
@@ -0,0 +1,24 @@
+apiVersion: v2
+name: otomi-install
```
svatwork, line 2: I don't know about the others but I would just name it otomi. I think it's standard if I look at the naming of other open source charts (e.g. https://github.com/prometheus-community/helm-charts/tree/main/charts: alertmanager is just called alertmanager, prometheus just prometheus).

Morriz: yes please

Morriz: and can this folder be renamed "chart" so it is clear it is the chart that comes with this software/

otomi-install/templates/serviceaccount.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: ServiceAccount
+metadata:
+  name: '{{ .Release.Name }}'
+---
+apiVersion: rbac.authorization.k8s.io/v1
```
Morriz, line 6: please put this in `rbac.yaml`

j-zimnowoda (commented):
LGTM
An idea for improvement:
We could enrich this chart by moving values.schema.json inside the chart.
> values.schema.json  # OPTIONAL: A JSON Schema for imposing a structure on the values.yaml file
Read more: https://helm.sh/docs/topics/charts/

mojtabaimani:
> LGTM
> An idea for improvement:
> We could enrich this chart by moving values.schema.json inside the chart.
> 
> > values.schema.json  # OPTIONAL: A JSON Schema for imposing a structure on the values.yaml file
> > Read more: https://helm.sh/docs/topics/charts/

the otomi-core value-schema.json is outside of the chart folder, so the helm command do not have access to it. Should we make a copy of the value schema in the chart folder?

Morriz:
> An idea for improvement:
> We could enrich this chart by moving values.schema.json inside the chart.

we can generate that in there yes.

j-zimnowoda (approved):

PR #456, opened 2021-06-30
set maxlength of dns

bartusz01:
fix for [otomi-api #182](https://github.com/redkubes/otomi-api/issues/182)

### DIFF
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -644,7 +644,8 @@ definitions:
           workload:
             $ref: '#/definitions/labelsAnnotations'
       domain:
-        description: A custom service domain name.
+        description: A custom service domain name (max 64 bytes).
+        maxLength: 64
         type: string
       forwardPath:
         default: false


Morriz (changes requested):
not gonna be functional because it is not used in the forms. we need functional checking on the length in the api which constructs the url from the "Host" and "Dns zone". This needs to have code that checks that length and throws an http 400 error if it exceeds.

bartusz01:
@Morriz Yes, I figured, thats why I posted my comment on the related issue today. Thanks for the explanation on how to do it, but its too much for me to look into atm. So perhaps I can bother you or Jeho to do some peer programming soon.

Morriz (approved):
tnx

bartusz01:
**For later reference** 
In case we want to enable domain names of 64+ bytes, we need to make sure the lets-encrypt certs can be issued with cert-manager, as this is whats blocking us.
This can be done when using SANs (subject alternative names), of which at least one should be within the 64 bytes limit. I played around with the simple ingress manifest below, using cert-manager and external-dns, resulting in a correctly issued certificate and the 64+  domain could be reached. 
The only minor issue I ran into is that external-dns will also create a record for the shorter domain name (SAN) in ```spec.tls[].hosts```. This record could be removed or perhaps there is a better solution if we would ever look into this.

~~Another solution might be to use wildcard certificates but I guess we want to avoid this~~
```yaml
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  annotations:
    cert-manager.io/cluster-issuer: letsencrypt-clusterissuer
    kubernetes.io/ingress.class: xxx
  name: my-ingress
spec:
  rules:
  - host: i-am-toooooooooooooooooooooooooooooooooooooooooooooooo-long.xxx.nl
    http:
      paths:
      - backend:
          service:
            name: my-svc
            port:
              number: 80
        path: /
        pathType: Prefix
  tls:
  - hosts:
    - i-am-toooooooooooooooooooooooooooooooooooooooooooooooo-long.xxx.nl
    - long.xxx.nl
    secretName: my-svc-tls
```

Morriz:
If you would please look at the cert-manager's cert template in otomi-core/helmfile.d/snippets, then you might be able to create a PR for this.

PR #470, opened 2021-07-16
fix: gitea initcontainer ca volumeMount name fixed

mojtabaimani:


Dunky13 (approved):
Nasty little bugger, lgtm

Morriz (commented):
The setup is working in multiple locations successfully (I can recall Drone as well). If it only fails for gitea I start to wonder if this is the right fix

Dunky13:
It was mounting a wrong volume. Not sure how it could've worked in previous runs? Maybe not spotted due to production? But it tried to mount a secret, not the certs. This change actually mounts the certs

Morriz:
Did something change in the setup since then?

mojtabaimani:
when you add docker imagePullSecret directly in the pod they will not take the gcr imagePullSecret form the default service account anymore. I added the gcr imagePullSecret to the pods as well.

values/drone/drone-admit-members.gotmpl
```diff
@@ -29,4 +29,5 @@ podAnnotations:
 image:  
   pullSecrets:
     - name: otomi-pullsecret-global
+    - name: gcr-json-key
```
Morriz, line 32: Why are you adding a pullsecret here for otomi api?

values/drone/drone.gotmpl
```diff
@@ -112,4 +112,5 @@ extraRootCA: |
 {{- with .Values.otomi | get "globalPullSecret" nil }}
 imagePullSecrets:
   - name: otomi-pullsecret-global
+  - name: gcr-json-key
```
Morriz, line 115: Same? Are you just trying things and not confirming the effects?

Morriz (changes requested):
This looks like another PR with changes that make no sense to me, and I have no clue what you did to come to this change, or why and what it would solve.
I would like to ask you to write out what you intend to do, what you decided, how you think to solve something, and how you tested your work, and wether you can conclude it solves the problem at hand.

Sorry to be so anal but we can't afford devs to suggest changes that introduce instability or bad code

values/drone/drone-admit-members.gotmpl
svatwork: I had to add this Secret in `otomi-eks-dev` or else it threw a pull error, it doesn't make sense to you? I am curious what the alternative solution is (also for my learning experience)?

svatwork: I didn't understand @Morriz comment, but he means like "why is it added in this namespace", looking at it now I wonder if it is relevant in this context? Did you verify it @mojtabaimani?

svatwork (changes requested):

Morriz (changes requested):

Morriz (approved):

PR #473, opened 2021-07-21
Linux workaround

### LINKED ISSUE #353: update otomi with workaround for docker run in linux

svatwork:
Closes #353.


Dunky13:
Can we please not work/make changes to the old CLI? Gives me a lot of overhead porting it.

svatwork:
One line of code is overhead? The "old" CLI is still in use so how else can I solve this issue?

svatwork:
Just add it and I will do it if the new cli is delivered...

Morriz (approved):

Dunky13 (approved):
Don't see the added value, as the new CLI will already have this, and can already be used. But for the sake of continuation, let's merge it.

PR #477, opened 2021-07-23
fix: alertmanager will be enabled only if alerts block is provided in values.yaml in chart

mojtabaimani:


chart/otomi/Chart.yaml
```diff
@@ -4,7 +4,7 @@ description: A Helm chart for installing otomi in Kubernetes
 home: https://otomi.io/
 icon: https://otomi.io/img/otomi-logo.svg
 type: application
-version: '0.1.5'
+version: '0.1.6'
```
Morriz, line 7: why is this bumped when the chart has no changes??

Morriz (changes requested):

chart/otomi/Chart.yaml
mojtabaimani: the current release is 0.1.6 already. when I had pushed to my branch it is released. Now it is just merging with master. It will not create a new release I guess.

Dunky13 (approved):
lgtm

Morriz (approved):

PR #482, opened 2021-07-26
fix: only release the chart if it is pushed to master

mojtabaimani:


.github/workflows/main.yml
```diff
@@ -74,7 +74,13 @@ jobs:
...
           release_name: Release ${{ steps.git_tag.outputs.tag }}
           body: ${{ steps.git_tag.outputs.changes }}
-      - name: Prepare chart for publishing
+      - if: github.ref == 'refs/heads/master'
+        name: Check if the branch is master 
+        id: branch_name
+        run: |
+          echo ::set-output name=master::true
+      - if: '${{ steps.branch_name.outputs.master }}'
+        name: Prepare chart for publishing
```
svatwork, line 83:
```suggestion
      - if: github.ref == 'refs/heads/master'
        name: Prepare chart for publishing
```

This is the same, right? Why set the variable?

But only publishing on master seems like a good option to me, now the releases page of `otomi-core` is producing a lot of noise...

svatwork (changes requested):

svatwork (approved):

PR #506, opened 2021-08-09
fix: extract secrets path bug fixed

mojtabaimani:
There was an error extracting secret path after converting bash to typescript in otomi chart. There was always an extra 'properties.' at the beginning of the paths. Now it is fixed.


j-zimnowoda (approved):
Looks ok but why unites did not discover that bug?

j-zimnowoda (approved):

PR #526, opened 2021-08-20
fix: 🐛 Imply decrypt step in apply,sync,destroy

Dunky13:
Imply that decrypt needs to happen on apply.


Morriz:
closing as already fixed

Dunky13:
Where is this already fixed? Not in master?

PR #597, opened 2021-09-20
fix: drone resources

Morriz:


tpl/.drone.yml.gotmpl
```diff
@@ -32,6 +39,13 @@ steps:
 {{- else if eq .provider "msteams" }}
   - name: notify-start
     image: diyfr/drone_plugin_teams
+    resources:
+      limits:
+        cpu: 200
+        memory: 200Mi
+      requests:
+        cpu: 100
+        memory: 100Mi
```
Dunky13, line 48: Can we make this a repeatable block? That we can call once - in a variable, function or something similar? 
Since all the resources are the same. Would make maintaining in the long run easier, I think

Dunky13 (commented):
Looks good, one question

tpl/.drone.yml.gotmpl
Morriz: I tend to stray from abstracting similar resource blocks as those should usually be customized to the workload, so only the similar notification calls remain as suspect for your suggestion. Let me just do those

tpl/.drone.yml.gotmpl
```diff
@@ -22,6 +22,13 @@ steps:
 {{- if eq .provider "slack" }}
   - name: slack-start
     image: plugins/slack
+    resources:
+      limits:
+        cpu: 200
```
svatwork, line 27: Is e.g. `cpu: 200` valid quantity syntax? (https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-units-in-kubernetes) 

I can't find anything about it in the documentation and never have seen it before (it seems like you allocate 200 cpu).

svatwork (changes requested):

tpl/.drone.yml.gotmpl
Morriz: it's how it is in the drone docs...tested and works

PR #598, opened 2021-09-23
fix: team alerts, prom monitors for core apps

Morriz:


Morriz:
This also includes an upgrade to cert-manager chart as that includes the servicemonitor fix...deploying on a fresh cluster to check

svatwork (approved):

PR #600, opened 2021-09-27
fix: 🐛 Docker if image exists check only on non-CI

Dunky13:


binzx/otomi
```diff
@@ -300,8 +302,11 @@ fi
 
 mkdir -p /tmp/otomi
 container_name='otomi-core-dev-container'
-count=$(docker ps -a | grep $container_name | wc -l) >/dev/null
-count=$(($count))
+count=0
+if [ -n "$has_docker" ]; then
+  count=$(docker ps -a | grep $container_name | wc -l) >/dev/null
+  count=$(($count))
```
Morriz, line 308:
```suggestion
  count=$((docker ps -a | grep $container_name | wc -l)) >/dev/null
```

Morriz (changes requested):

binzx/otomi
Dunky13: Nope, doesn't work:
```
➜ count=$((docker ps -a | grep $container_name | wc -l)) >/dev/null
zsh: bad math expression: operator expected at `ps -a | gr...'
```

binzx/otomi
```diff
@@ -43,6 +43,7 @@ silent() {
 }
 
 docker_tag_exists() {
+  [ -n "$CI" ] && return 0
```
Morriz, line 46: why is this added?

Morriz (changes requested):

binzx/otomi
Dunky13: To skip when running in CI mode

Morriz: is that needed? and here in this function? out of place and confusing

Morriz: please create conditions where they apply, so it stays maintainable

binzx/otomi
```diff
@@ -300,8 +302,10 @@ fi
 
 mkdir -p /tmp/otomi
 container_name='otomi-core-dev-container'
-count=$(docker ps -a | grep $container_name | wc -l) >/dev/null
-count=$(($count))
+count=0
+if [ -n "$has_docker" ]; then
+  count=$(($(docker ps -a | grep -c $container_name))) >/dev/null
+fi
```
Morriz, line 308:
```suggestion
[ -n "$has_docker" ] &&  count=$(($(docker ps -a | grep -c $container_name))) >/dev/null
```

PR #676, opened 2021-12-02
Moz/test/bootstrap

Morriz:
## Checklist


kind/Dockerfile
```diff
@@ -18,4 +18,4 @@ WORKDIR kind
 COPY chart chart/
 COPY helmfile.yaml kind.gotmpl kind.yaml start-kind.sh ./
 
-CMD "./start-kind.sh"
\ No newline at end of file
+CMD "./kind/start-kind.sh"
```
0-sv, line 21: I don't think this is going to work because start-kind.sh is copied into the container, but not in the `kind` path.

tools/Dockerfile
```diff
@@ -131,15 +130,17 @@ RUN tar -zxvf /tmp/conftest_${CONFTEST_VERSION}_Linux_x86_64.tar.gz -C /tmp && m
...
-
 # node
 RUN curl -sL https://deb.nodesource.com/setup_${NODE_VERSION}.x | bash -
 RUN apt-get install -y nodejs && \
   npm install -g ajv-cli@v3.3.0 json-dereference-cli@0.1.2 zx
 
+# docker
+RUN curl -sSL https://get.docker.com/ | sh
+
+# kind
```
0-sv, line 141: I see this is copied over from `kind/Dockerfile`. That file still has these statements.

0-sv (changes requested):
If you want, we can discuss the refactoring of kind (if you think it has business value to do so right now), but I think this is a bit too hasted and I would suggest to remove it from this PR because it will break the build. 

The mock tests looks good, thanks for adding unit tests.

kind/Dockerfile
Morriz: everything is copied as-is,  and that is the location it is found from CWD in the container

tools/Dockerfile
Morriz: that other Dockerfile will be removed

Morriz: but I reverted the kind changes like you suggested...and also because I don't wish to add dev tools to the final core container

PR #706, opened 2022-01-04
Moz/refactor/dockerfiles

Morriz:
splits up dockerfiles and compresses their builds to minimize size:

otomi/core:
1.7Gi before
1.2Gi after


j-zimnowoda (commented):
Thanks 
How should I test it. Do you have any scenarios?

Morriz:
you may inspect the Dockerfiles? It is not complex. As the description mentions: it saves 500MB size when leaving out unnecesary tools, but it adds this setup to maintain. Will be beneficial when we get traction

j-zimnowoda:
yes, we will do it in this week.

binzx/otomi
```diff
@@ -69,7 +74,25 @@ else
...
 otomi_version_used=${OTOMI_TAG:-$otomi_version}
-readonly otomi_tools_image="otomi/core:${otomi_version_used}"
+otomi_image="cli"
+[ -n "$IN_DOCKER" ] && otomi_image="core"
+otomi_tools_image="otomi/$otomi_image:$otomi_version_used"
+# backwards compatibility with pre core/cli split era
+if [ $otomi_image = 'cli' ] && ! docker_tag_exists otomi/cli $otomi_version_used; then
+  # check if master version exists
+  if docker_tag_exists otomi/cli master; then
+    otomi_version_used='master'
```
j-zimnowoda, line 84: Since we moved to main branch, the tag should be `main`, right ?

binzx/otomi
```diff
@@ -69,7 +74,25 @@ else
...
+  else
+    otomi_image='core'
+  fi
+  otomi_tools_image="otomi/$otomi_image:$otomi_version_used"
+fi
+# finally fall back to master tag if image not found
+if ! docker_tag_exists $otomi_image $otomi_version_used; then
+  otomi_version_used=master
```
j-zimnowoda, line 92: `main`

binzx/otomi
```diff
@@ -288,7 +312,7 @@ elif [[ $calling_args == *'--get-yargs-completions'* ]]; then
   docker run --name $container_name --rm -e SHELL=$SHELL "$otomi_tools_image" bash -c "$cmd"
   status=$?
 else
-  docker run --name $container_name --rm --init $it \
+  docker run --name $container_name --rm --init $it -u 0 \
```
j-zimnowoda, line 315: Why is it necessary to run as root

tools/Dockerfile.cli
```diff
@@ -0,0 +1,14 @@
...
+FROM otomi/core:$TAG
+
+COPY --from=tools $TOOLS_HOME/awscli $TOOLS_HOME/awscli
+COPY --from=tools $TOOLS_HOME/aws-iam-authenticator $TOOLS_HOME/aws-iam-authenticator
+COPY --from=tools $TOOLS_HOME/gsdk $TOOLS_HOME/gsdk
+COPY --from=tools $TOOLS_HOME/opa $TOOLS_HOME/opa
+
+ENV PATH $PATH:$APP_HOME/awscli/bin:$APP_HOME/gsdk/bin
+
+USER root
```
j-zimnowoda, line 12: I do not really see the necessity to run us root. Perhaps you could clarify it.

diabhey:
We should consider using this tool: https://github.com/docker-slim/docker-slim

tools/Dockerfile.cli
```diff
@@ -0,0 +1,14 @@
...
+RUN chown -R app:app $TOOLS_HOME/awscli $TOOLS_HOME/aws-iam-authenticator $TOOLS_HOME/gsdk $TOOLS_HOME/opa
```
diabhey, line 13: https://github.com/koalaman/shellcheck/wiki/SC2086

tools/Dockerfile.tools
```diff
@@ -0,0 +1,42 @@
...
+
+# opa
+ADD https://github.com/open-policy-agent/opa/releases/download/v${OPA_VERSION}/opa_linux_amd64 opa
+RUN chmod +x opa
+
+# kubescore
+ADD https://github.com/zegl/kube-score/releases/download/v${KUBESCORE_VERSION}/kube-score_${KUBESCORE_VERSION}_linux_amd64 kube-score
+RUN chmod +x kube-score
+
+RUN chown -R app:app $TOOLS_HOME
```
diabhey, line 38: Double quote to prevent globbing and word splitting: https://github.com/koalaman/shellcheck/wiki/SC2086

tools/Dockerfile.tools
```diff
@@ -0,0 +1,42 @@
...
+
+USER app
+
+CMD "/bin/bash"
```
diabhey, line 42: https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#cmd

tools/build.sh
```diff
@@ -0,0 +1,11 @@
...
+TAG=${TAG:-"v1.4.22"}
+
+cd tools >/dev/null 2>&1 || true
+
+docker build -t otomi/base:$TAG -f Dockerfile.base --squash .
+docker build -t otomi/tools:$TAG -f Dockerfile.tools --squash .
+# Dockerfile.cli is built from github workflow
+
+docker push otomi/base:$TAG
+docker push otomi/tools:$TAG
```
diabhey, line 11: Fixed  https://www.shellcheck.net/wiki/SC2148
```bash
#!/usr/bin/env bash
TAG=${TAG:-"v1.4.22"}

cd tools >/dev/null 2>&1 || true

docker build -t otomi/base:"$TAG" -f Dockerfile.base --squash .
docker build -t otomi/tools:"$TAG" -f Dockerfile.tools --squash .
# Dockerfile.cli is built from github workflow

docker push otomi/base:"$TAG"
docker push otomi/tools:"$TAG"
```

diabhey (commented):
- Some minor best practices changes

tools/Dockerfile.tools
Morriz: Not needed as no spaces in path. We do DRY

tools/build.sh
Morriz: Again, tags never contain spaces. So not needed. You might add exclusions for shellcheck which is a horrible tool

Morriz:
Tnx @diabhey for the suggestions. I rejected them in favor of shellcheck exclusions for our Dockerfile(s). String split should only warn.

Morriz:
I prefer not to. I think we should use minimal base images that are secure.
We are not in need of creating base images ourselves

binzx/otomi
Morriz: I don't remember...try it without?

binzx/otomi
Morriz: you may commit inline with batch

j-zimnowoda:
> you may inspect the Dockerfiles? It is not complex. As the description mentions: it saves 500MB size when leaving out unnecesary tools, but it adds this setup to maintain. Will be beneficial when we get traction

I inspected them but it is a bit convoluted, because they are released in many places, so it is difficult to grasp an idea and check if order of build pipeline is correct.
So far I see that there are four container images now:
- core
- base
- tools
- cli 
How do they depend on each other?

Btw, it does not sound like high priority now, doesn't it?

Morriz:
> > you may inspect the Dockerfiles? It is not complex. As the description mentions: it saves 500MB size when leaving out unnecesary tools, but it adds this setup to maintain. Will be beneficial when we get traction
> 
> I inspected them but it is a bit convoluted, because they are released in many places, so it is difficult to grasp an idea and check if order of build pipeline is correct. So far I see that there are four container images now:
> 
> * core
> * base
> * tools
> * cli
>   How do they depend on each other?

C'mon, those dockerfiles are easy to read. What do you not get? Only core is split to also deliver cli

> Btw, it does not sound like high priority now, doesn't it?

Not really, but saving 500Mb is quite a lot, so it would be beneficial for many reasons. Having all nodes pulling the otomi container when assigned a drone runner is costly.

PR #804, opened 2022-05-24
feat(component): add kubeclarity helm chart

staticvoid255:
Please merge this before #800 !

add the kubeclarity helm chart only, means we'll get a better diff when merging feature/kubeclarity branch

feat #409


j-zimnowoda (approved):

staticvoid255:
It wasn't me that suggested this, it was @j-zimnowoda and the purpose of which was to create a shared diff that he could see without being beside me looking at my screen. Wouldn't be an issue if we had a develop branch...

Morriz:
I don't care who suggests what, as long as devs understand the importance of convening with the lead dev/architect (me) about important decisions such as fidgeting with the main branch.

And checking diffs should be done before committing in order to get the right stuff in/done. We call this putting the carriage before the horse.

If devs are interested during review about the exact patches applied onto a 3rd party chart, then in that PR it is very well possible to do what you guys propose, but I see no purpose in keeping that diff around and after review that will be squashed.

Morriz:
What is more important is understanding what always needs to be done to/exist in charts to meet our requirements. That knowledge is now not concentrated in one location, so please update our docs to reflect that if you are serious about improving our work.

staticvoid255:
Do we have a commit policy written? If so I would appreciate you directing me towards it as I certainly didn't see anything like that in the developer onboarding guide. If the documentation does not exist then I would recommend that it be written by the person enforcing said policy to prevent any future stress.

staticvoid255:
We should also include a contribution guide in the repo as most open source projects do.

Morriz:
Commit policy is enforced with our tooling, so that is not an issue. Again, if you feel in need of documentation (you can infer a lot from our tooling and release history) and that is not in there, I hope you can update it.

Morriz:
About a contribution guide: we have a central place for most documentation, which is otomi.io. We try to keep our readme files limited and use refs to the central documentation. If you are interested in everything we do, and want to "fix" things you see as problematic (might be due to missing context knowledge?) then I suggest adding an issue to zenhub.

PR #836, opened 2022-07-04
fix(config): change polling_interval to poller_interval_milliseconds

staticvoid255:
Fix for KES - polling_interval should be poller_interval_milliseconds

### DIFF
--- a/helmfile.d/snippets/defaults.yaml
+++ b/helmfile.d/snippets/defaults.yaml
@@ -29,7 +29,7 @@ environments:
             domainFilters: []
           external-secrets:
             logLevel: info
-            pollingInterval: 60000
+            pollerIntervalMilliseconds: 60000
           gatekeeper:
             enabled: false
             auditFromCache: false
--- a/values-changes.yaml
+++ b/values-changes.yaml
@@ -19,6 +19,7 @@ changes:
       - apps.prometheus.resources.grafana: apps.grafana.resources
       - apps.prometheus.alertmanager: apps.alertmanager
       - apps.prometheus.resources.alertmanager: apps.alertmanager.resources
+      - apps.external-secrets.pollingInterval: apps.external-secrets.pollerIntervalMilliseconds
     mutations:
       - alerts.drone: 'list .prev'
       - teamConfig.{team}.services[].networkPolicy.ingressPrivate.mode: '.prev | title'
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -1992,7 +1992,7 @@ properties:
             type: boolean
           logLevel:
             $ref: '#/definitions/logLevel'
-          pollingInterval:
+          pollerIntervalMilliseconds:
             type: integer
             default: 60000
       kube-descheduler:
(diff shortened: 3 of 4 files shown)


j-zimnowoda (commented):
Checked with documentation. Looking good.
Since schema has changed I recommend adding a new entry in values-changes.yaml that indicates this `relocation`

Morriz:
I don't think we should put units in the names of configuration, as the description/examples/regex already show what is expected. This goes the same for variables in code.

Morriz:
no, please understand, I talked to the core dev and he wants to change the interval...

Morriz:
That is why it is BAD PRACTICE to be overly specific in naming

Morriz:
I sincerely hope you make good decisions in such matters in my absence. You may think it is nitpicking, but it is all about guarding the guide so that such poor decisions do not proliferate and incur unneeded work

staticvoid255:
I would personally lean towards the name being as close to the actual name in the chart as possible so as not to insert ourselves too much between the user and underlying Helm chart

Morriz:
This is about coding in general. Please get some courses on best practices and read coding style guides about variable naming etc. We also have the added limitation that bad choices are more impactful than variable naming: we expose configuration to users.

staticvoid255:
Ok well your opinion on naming doesn't change the fact that you've used `POLLING_INTERVAL`  instead of `POLLER_INTERVAL_MILLISECONDS` and therefore aren't even targeting the actual value in the Helm chart.

Morriz:
moot ;p

staticvoid255:
I also think a unit change scenario doesn't apply here. There's no way the developers are going to change to say, seconds, without changing the name of the value - why would  they ever do so without changing `POLLER_INTERVAL_MILLISECONDS` to `POLLER_INTERVAL_SECONDS`? It's better in my opinion that the correspondence between our definitions and theirs be as close possible. In the case of a unit change It would be better practice to fall back to Helm defaults than applying the existing value under the new unit regime.

Morriz:
nevermind, this is my mistake as I see a PR out of the blue not explaining this is actually a bug fix. I just see a PR that I reply to as code gatekeeper. It was introduced without a message what it is for, and used wording I did not recognize. I think I would have recognized "`POLLING_INTERVAL` should be `POLLER_INTERVAL_MILLISECONDS`".

My mistake to assume it was a suggestion.

Morriz:
> I also think a unit change scenario doesn't apply here. There's no way the developers are going to change to say, seconds, without changing the name of the value - why would they ever do so without changing POLLER_INTERVAL_MILLISECONDS to POLLER_INTERVAL_SECONDS? It's better in my opinion that the correspondence between our definitions and theirs be as close possible. In the case of a unit change It would be better practice to fall back to Helm defaults than applying the existing value under the new unit regime.

I incorrectly assumed their choice was `POLLING_INTERVAL` (which, like prometheus' `duration`, could have been expressed like `1s` or even `1m`)

Morriz:
FYI: I adopted this fix into main for next release (coming up)

Issue #845, opened 2022-07-21
Vault unseal backup to Azure KeyVault not working


staticvoid255:
Getting the following error after configuring Azure KV backup for seal keys:
`Error parsing Seal configuration: key name is required`

Looking at the Vault docs it should be as easy as adding key_name to the values schema in Core and API,
I have branches created for this and will test (fix/azure-vault-unseal)

Possible that other KMS providers don't work, but this is high priority for the client and they use Azure KV so I'm tackling this first.

PR #846, opened 2022-07-21
fix: azure vault unseal

staticvoid255:
This is a fix for #845 

Tested and it works but one thing to be VERY careful about is that when you add an external provider for storing unseal secrets, that it will delete the kubernetes secret containing the keys **without** pushing them to keyvault, so you absolutely need to take a backup of that key before enabling this configuration.

I have included a warning in the field description in the values schema to help people avoid losing their keys (in both Core and API).

[Respective PR in API](https://github.com/redkubes/otomi-api/pull/316)

### DIFF
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -2292,11 +2292,14 @@ properties:
                       client_secret:
                         type: string
                         x-secret: ''
+                      key_name:
+                        $ref: '#/definitions/wordCharacterPattern'
                     required:
                       - vault_name
                       - tenant_id
                       - client_id
                       - client_secret
+                      - key_name
                 required:
                   - azurekeyvault
           storage:
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -2238,8 +2238,8 @@ properties:
           seal:
             description: |
               Use a KMS to encrypt and decrypt the master key.\n
-              WARNING - enabling this feature will delete the unseal key from the cluster without pushing to external kms!\n
-              You must first copy the keys from the "vault-unseal-keys" secret, and import them to your KMS.
+              WARNING - You must first copy the keys from the "vault-unseal-keys" secret, and import them to your KMS.\n
+              (enabling this feature will delete the unseal key from the cluster without pushing to external kms!)
             nullable: true
             type: object
             oneOf:
--- a/values-schema.yaml
+++ b/values-schema.yaml
@@ -2237,9 +2237,8 @@ properties:
             $ref: '#/definitions/logLevel'
           seal:
             description: |
-              Use a KMS to encrypt and decrypt the master key.
-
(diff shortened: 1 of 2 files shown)


values-schema.yaml
```diff
@@ -2292,11 +2295,14 @@ properties:
...
                         type: string
                         x-secret: ''
+                      key_name:
+                        $ref: '#/definitions/wordCharacterPattern'
                     required:
                       - vault_name
                       - tenant_id
                       - client_id
                       - client_secret
+                      - key_name
```
Morriz, line 2305: any reason to make this required? I think it should be optional

Morriz (changes requested):
one thing

values-schema.yaml
staticvoid255: because it throws an error if you don't include a key name and is required by the Vault schema
from the issue:
`Error parsing Seal configuration: key name is required`

j-zimnowoda (approved):

values-schema.yaml
Morriz: But I mean it is optional IF you don't want to opt into your new unseal setup. So it is ONLY required then. If you look into other areas where we have opional required fields (look for anyOf) you understand. We try to keep the api schema in core well validated for `otomi validate-values`.
@j-zimnowoda please also check these situations in reviews.

Morriz:
@staticvoid255 please don't resolve issues others have started and have to approve/agree to

staticvoid255:
> @staticvoid255 please don't resolve issues others have started and have to approve/agree to

I don't know what you mean as I did no such thing, and was myself the creator of the issue. Since you are on sabbatical I went through Jeho for this, so please take it up with him.

values-schema.yaml
staticvoid255: Please review the code and note that all fields are required for this feature to work, and that this PR is only in the context of Azure KeyVault.

Morriz: I meant this comment that was created by me, that was set to resolved by you. I get that you thought I would not respond in time, but I still don't see why this breaking change was introduced for Azure, when the signing key should have been made optional. @j-zimnowoda why was this not caught?

staticvoid255: Again, the key name is required. Please read #845 as linked in the PR:
`Error parsing Seal configuration: key name is required`
No breaking change was introduced as the feature was broken to begin with.

staticvoid255: Also, this was all in consultation with Jeho while you were on holidays. Just because I set it to resolved doesn't mean I made that decision on my own.

Morriz: That is not true, as we were using already Azure signing for a long time. Also on GemU ONT and ACC. I use it all the time. It is simply not mandatory to determine a key, and Azure will generate one for you.

PR #877, opened 2022-08-16
feat: add vultr provider

srodenhuis:
This PR adds a new cluster.provider to Otomi for Vultr managed kubernetes.
The VKE requires minimal persistent volumes size of 10GB


j-zimnowoda (approved):
I approve as it is good enough. After reviewing this PR I noticed that we should allow users to adjust storage size.

PR #937, opened 2022-11-08
feat: Thanos and Minio with updated Velero config

srodenhuis:
This PR brings:

- Thanos and Minio in an out-of-the-box remote write setup
- Configure Prometheus for remote write
- Install Minio with provisioning (creating buckets and policies for all Otomi apps to consume object storage
- Splitting up cloud and storage provider for Velero to support using in-cluster Minio on other clouds

## Checklist

- [ x ] The OpenApi Schema from redkubes/otomi-api project is compatible with definitions from `values-schema.yaml` file, if applicable.
- [ x ] Helm releases are meeting otomi's baseline security policies, if applicable.


staticvoid255:
Hi Sander - I noticed that if you try to install Minio and Thanos at the same time, that Core will try to install Thanos first. This then causes the Drone pipeline to hang as Thanos is stuck in error trying to reach a Minio instance that doesn't yet exist.
Screenshot 2022-11-10 at 15 13 16

Morriz:
> Hi Sander - I noticed that if you try to install Minio and Thanos at the same time, that Core will try to install Thanos first. This then causes the Drone pipeline to hang as Thanos is stuck in error trying to reach a Minio instance that doesn't yet exist.
> Screenshot 2022-11-10 at 15 13 16

Any app that has a dependency on another service should wait for that service in a pre-install job. Easiest way to do that is to add a `needs:` dependency to the helmfile release by deploying another `wait-for` chart. Both releases need to be in the same helmfile. Example:
```yaml
  - name: thanos
    namespace: thanos
    needs: [maintenance/wait-for-minio]
    ...
  - name: wait-for-minio
    installed: {{ $a | get "minio.enabled" }}
    <<: *wait
    values:
      - url: {{ http://minio.minio }}
        host: minio.{{ $v.cluster | get "domainSuffix" nil }}
        tasksVersion: {{ $v.versions.tasks }}
        {{- if $v._derived.untrustedCA }}
        customCa: |
          {{- $v._derived.caCert | nindent 10 }}
        {{- end }}
        runPolicy: OnSpecChange    

```

helmfile.d/snippets/defaults.yaml
```diff
@@ -143,24 +150,113 @@ environments:
...
+                  memory: 256Mi
+            compactor:
+              resources:
+                requests:
+                  cpu: 200m
+                  memory: 128Mi
+                limits:
+                  cpu: 500m
+                  memory: 256Mi
+              retentionResolutionRaw: 30d
```
j-zimnowoda, line 220: > As a rule of thumb retention for each downsampling level should be the same, and should be greater than the maximum date range (10 days for 5m to 1h downsampling).
https://thanos.io/tip/components/compact.md/#-downsampling-note-about-resolution-and-retention-

Could you explain why do you chose `retentionResolution1h: 10y`

helmfile.d/snippets/defaults.yaml
```diff
@@ -143,24 +150,113 @@ environments:
...
+              retentionResolution5m: 30d
+              retentionResolution1h: 10y
+              persistence:
+                size: 8Gi
```
j-zimnowoda, line 224: > The compactor needs local disk space to store intermediate data for its processing as well as bucket state cache. Generally, for medium sized bucket about 100GB should be enough to keep working as the compacted time ranges grow over time.
https://thanos.io/tip/components/compact.md/#disk

8 GB does not seem to be a good default for compactor

j-zimnowoda (changes requested):
So far few comments. Overall way too many changes in one PR.

values/minio/minio.gotmpl
```diff
@@ -0,0 +1,127 @@
...
+
+{{- if $v._derived.untrustedCA }}
+extraVolumes:       
+  - name: root-ca-1
+    secret:
+      secretName: root-ca
+  - name: root-ca-2
+    secret:
+      secretName: root-ca
+extraVolumeMounts:
```
j-zimnowoda, line 37: Could you add a comment in the code explaining why the same cert needs to be mounted twice?

values/minio/minio.gotmpl
```diff
@@ -0,0 +1,127 @@
+{{- $v := .Values }}
+{{- $vl:= $v.apps.minio }}
```
j-zimnowoda, line 2: I believe this variable should be called `$m`

srodenhuis:
Will be split up into 4 separate PRs

PR #945, opened 2022-11-19
feat: upgrade harbor

srodenhuis:
depends-on: redkubes/otomi-core/pull/952
This PR delivers:

- Upgrade of Harbor from v2.4.0 to v2.5.4

TODO:
- [ ] test otomi upgrade with enabled harbor


j-zimnowoda (changes requested):
Thanks Sander,
As we discussed it this morning, please move changes to related to DB backup to a spare PR. 
Then harbor upgrade PR will not have a dependency on minio, which is not delivered yet.

j-zimnowoda (approved):

PR #956, opened 2022-11-29
fix: prometheus resources

staticvoid255:
fixes the issue where resource quota modifications aren't applied


j-zimnowoda (approved):
LGTM
Do not worry about the team prometheus as it has its own resources already hardcoded
`./values/prometheus-operator/prometheus-operator-team.gotmpl` and those will be leading for prometheus-<team> Helm release.

Morriz (approved):

Morriz:
We should collect team specific config to inject into these team versions of kube-prometheus-stack tho, so teams have control over it's settings

staticvoid255:
yes I have suggested this to Jeho last week and created unassigned-issues/#494

PR #998, opened 2023-01-05
fix: kiali deps

srodenhuis:
Makes Kiali require Prometheus


Morriz:
closing as there is (or should be) no dep on prometheus. We will fix the gotmpl that is now spaghetti

j-zimnowoda (approved):

Issue #1000, opened 2023-01-09
Monitoring of "podMonitor" and  "serviceMonitor" from team namespaces not working if multitenant disabled mode


staticvoid255:
Hi there,

I noticed that our Prometheus XXNamespaceSelectors in our templates does not contain logic allowing the global instance of Prometheus to target team namespaces even if multitenancy is disabled.

Prometheus-operator.gotmpl:
```

prometheus:
  enabled: {{ $p.enabled }}
  prometheusSpec:
    {{- range $selType := list "podMonitor" "probe" "rule" "serviceMonitor" }}
    {{ $selType }}NamespaceSelector:
      matchExpressions:
        - key: name
          operator: NotIn
          values:
            {{- range $teamId, $_ := $v.teamConfig }}
            - team-{{ $teamId }}
            {{- end }}

```

Suggested fix is to extend this logic to allow the global instance to target team namespaces if multitenancy is disabled

Cheers

Morriz:
> Hi there,
> 
> 
> 
> I noticed that our Prometheus XXNamespaceSelectors in our templates does not contain logic allowing the global instance of Prometheus to target team namespaces even if multitenancy is disabled.
> 

Thanks. Just some facts about current state of things (which can be improved upon for sure):

1. Only when multitenancy is enabled do we deploy team prometheuses.
2. Team prometheuses are not governing what platform prometheus is already governing, and are not given any rulesets therefor. In effect this allows teams to publish custom metrics which team prometheuses act on. Nothing else (yet).

Morriz:
So I believe your initial observation about global prom not governing team namespaces is incorrect as it is not namespaced.

staticvoid255:
Hi there,

I have updated the title for clarity.
Can you please point to me in code where the NotIn team namespace constraint is lifted in the event that multitenancy is disabled, as from what I can see this constraint is `always ` applied regardless of disabled or not.

staticvoid255:
Drone pipeline logs - multitenancy disabled:
<img width="339" alt="Screenshot 2023-01-11 at 20 05 58" src="https://user-images.githubusercontent.com/59235486/211895830-b0b713d4-01c7-4d6f-aa52-5fc9a55c6b0d.png">

@Morriz please advise

Morriz:
If that is the case then that is clearly not the way it should be.

PR #1005, opened 2023-01-12
fix: enable Prometheus detection of XXMonitor objects in team namespace if MT=false

staticvoid255:
fixes: #1000 

If MT=false global Prometheus instance can pick up XXMonitors with `prometheus: system` label in any namespace.


staticvoid255:
behaving properly when MT is enabled, but !MT selector needs work - Prometheus docs aren't very clear on this and trying `any: true` didn't work. I'll come back to this next week.

Morriz (approved):
tnx

values/prometheus-operator/prometheus-operator.gotmpl
```diff
@@ -53,6 +54,7 @@ prometheus:
             {{- range $teamId, $_ := $v.teamConfig }}
             - team-{{ $teamId }}
             {{- end }}
+    {{- end }}
```
Morriz, line 57:
```suggestion
    {{- else }}
    {{ $selType }}NamespaceSelector:
      matchLabels: {}
    {{- end }}
```

Morriz (approved):
@staticvoid255 with my committed suggestion I believe it covers !MT

values/prometheus-operator/prometheus-operator.gotmpl
staticvoid255: To my surprise I found that with an empty value it targets only the monitoring namespace, so all-namespace targeting must be explicitly declared I'm afraid. With this commit it would resolve to the same as the chart default. Not ready for merging yet I'm afraid.

Morriz: `matchLabels: {}` is not an empty value though. The logic in the chart:
```yaml
{{- if .Values.prometheus.prometheusSpec.serviceMonitorNamespaceSelector }}
  serviceMonitorNamespaceSelector:
{{ toYaml .Values.prometheus.prometheusSpec.serviceMonitorNamespaceSelector | indent 4 }}
{{ else }}
  serviceMonitorNamespaceSelector: {}
{{- end }}
```
Did you try my suggestion?

staticvoid255: This will not work, `XXMonitorNamespaceSelector: {}` is the same as the chart default, and will result in Prometheus only picking up monitor objects in the monitoring namespace (the namespace into which Prometheus is deployed). Including your commit or not, it still resolves to `XXMonitorNamespaceSelector: {}`.
Like I said, we must explicitly declare all-namespace targeting, and I will come back to this Monday.

Morriz: Please read carefully:
```yaml
XXMonitorNamespaceSelector: {}
```
is NOT the same as:
```yaml
XXMonitorNamespaceSelector:
  matchLabels: {}
```

staticvoid255:
Tested and working, ready for review 
Screenshot 2023-01-16 at 14 13 07

j-zimnowoda (approved):

PR #1038, opened 2023-02-20
fix: scaling defaults

staticvoid255:
Increase default replica counts so as not to block nodes draining during k8s upgrade.

Notes:
- Gatekeeper controller manager does not support HPA, so the only practical option is to set a static default of 2
- Since istio ingress-gateway is already set to 2 replicas minimum in HPA, I have done the same for the egress gateway, because if egress gateway is enabled it could cause the same issue

### DIFF
--- a/helmfile.d/snippets/defaults.yaml
+++ b/helmfile.d/snippets/defaults.yaml
@@ -206,13 +206,13 @@ environments:
             enabled: true
             autoscaling:
               egressgateway:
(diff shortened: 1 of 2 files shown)


staticvoid255:
Extensively tested along with redkubes/otomi-api/pull/372 - working correctly

j-zimnowoda (approved):

PR #1049, opened 2023-03-10
fix: cluster autoscaler memory requests

leiarenee:
Cluster autoscaler is OOM Killed in a crash loop. 
Resource is increased to 256 MBi


j-zimnowoda (approved):

staticvoid255:
@j-zimnowoda plz check my review on this, I think given the nature of the error that limits should also be increased here

PR #1068, opened 2023-04-05
fix: Harbour S3 storage and otomi pull/commit

leiarenee:
- Fix Harbour S3 Backend storage
- Fix `otomi pull and otomi commit` (by [Maurice](https://github.com/Morriz))

### DIFF
--- a/.values/env/settings.yaml
+++ b/.values/env/settings.yaml
@@ -1,1 +1,1 @@
-version: 6
+version: 7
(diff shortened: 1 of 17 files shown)


j-zimnowoda (commented):
Great work. Also the documentation that you have made in Confluence :)

Since the schema has changed it usually means a breaking change. However, we have introduced a migration mechanism that takes care about it. See the `values-changes.yaml` file

I think that adding the following definition to the `values-changes.yaml` should be enough.
```
  - version: 7
    relocations:
      - apps.harbor.persistence.aws: 'apps.harbor.persistence.s3'
 ```
 
 Please run `npm run migrate-values`, so all test files are migrated to a new version ;)

Morriz:
That faulty config never worked before, so no migration is necessary. Right?

PR #1108, opened 2023-05-04
feat: scaleway provider

srodenhuis:
To install Otomi on Scaleway, we need a new provider. By using the Scaleway provider, metrics-server is not enabled (because Scaleway already runs metrics-server by default)

### DIFF
--- a/chart/otomi/values.yaml
+++ b/chart/otomi/values.yaml
@@ -18,7 +18,7 @@ cluster: {}
   # name: 'dev'
 
   ## Set provider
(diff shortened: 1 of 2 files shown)


j-zimnowoda (changes requested):
Hi, I can't find any logic regarding  the metrics-server. Could you cross check with your local version?

srodenhuis:
the provider is used (by existing logic) to avoid metrics-server being enabled in derived.gotmpl

j-zimnowoda (approved):

PR #1111, opened 2023-05-08
fix: cleanup

srodenhuis:
This PR removes unused charts and the knative-service manifest in the team-ns namespace.


srodenhuis:
has been done in other PR

PR #1139, opened 2023-06-01
feat: add otomi-db chart

CasLubbers:
## Checklist
 Merge after: https://github.com/redkubes/otomi-core/pull/1137


values/cloudnative-pg/cloudnative-pg.gotmpl
```diff
@@ -0,0 +1,9 @@
+{{- $v := .Values }}
+{{- $es := $v.apps | get "cloudnative-pg" }}
+
+limits:
+  cpu: 100m
+  memory: 200Mi
+requests:
+  cpu: 100m
+  memory: 100Mi
```
staticvoid255, line 9: Hi there,

Great work so far.
One thing - nowadays we try to avoid baking resource quotas into the templates, so instead of this you can replace with:
`resources: {{- $es.resources | toYaml | nindent 2 }}`

Then you can extend the values-schema.yaml with something like:
```
      cloudnative-pg:
        additionalProperties: false
        properties:
          _rawValues:
          enabled:
            type: boolean
            default: false
          resources:
```

then in defaults.yaml:
```
          cloudnative-pg:
            enabled: false
            resources:
              requests:
                cpu: 100m
                memory: 100Mi
              limits:
                cpu: 100m
                memory: 200Mi
```

PR #1145, opened 2023-06-09
feat: add dependencies to app info

srodenhuis:
Add dependencies to app info.


dennisvankekem (approved):
LGTM, however make sure to either fill in dependencies for each app or add a check in the console and don't show dependancies label if none are present

PR #1149, opened 2023-06-19
feat: make custom team Grafana dashboards persistent

srodenhuis:
Team Grafana dashboards in Otomi are automatically configured. If a dashboard is added manually and the team's Gafana instance restarts, all custom dashboards are lost. This PR adds the option to make custom dashboards persistent, by creating a custom-dashboards configmap in the team-namespace that will then be automatically loaded.

### DIFF
--- a/charts/team-ns/templates/dashboards/configmap.yaml
+++ b/charts/team-ns/templates/dashboards/configmap.yaml
@@ -1,0 +1,18 @@
+{{- $v := .Values }}
+{{- if $v.dashboards}}
+---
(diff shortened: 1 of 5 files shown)


staticvoid255:
Cool pattern injecting into the configmap, but it bypasses the grafana-dashboards chart/releases, which injects dashboards via filesystem (as far as I know).
My opinion is that it would be better to stick with a single approach, either sticking with the grafana-dashboards chart as-is, or refactoring the chart to implement the pattern you proposed here.

PR #1161, opened 2023-06-26
feat: upgrade prom stack

srodenhuis:
- Upgrade kube prometheus stack upstream chart from 36.2.0 to 46.4.1


srodenhuis:
When doing an upgrade, the following error occurs: 

```
Error: UPGRADE FAILED: release prometheus-operator failed, and has been rolled back due to atomic being set: cannot patch "prometheus-operator-prometheus-node-exporter" with kind DaemonSet: DaemonSet.apps "prometheus-operator-prometheus-node-exporter" is invalid: spec.selector: Invalid value: v1.LabelSelector{MatchLabels:map[string]string{"app.kubernetes.io/instance":"prometheus-operator", "app.kubernetes.io/name":"prometheus-node-exporter"}, MatchExpressions:[]v1.LabelSelectorRequirement(nil)}: field is immutable
```
It means there are properties that cannot be edited in the resource.
If the kube-prometheus-stack is stateless the we could simply remove the helm release before the upgrade

Ani1357 (approved):

PR #1168, opened 2023-06-30
fix: added upgrade script for kube prom stack

Ani1357:
The script will uninstall the "prometheus-operator" release before upgrading to any otomi version prior to 0.22.0.


srodenhuis (approved):
LGTM

PR #1170, opened 2023-07-02
fix: multiple builds issues and tekton 1.25 support

srodenhuis:
This PR provides the following fixes for the builds feature in Otomi:

- Remove PSP in tekton chart for 1.25 support
- correct namespace name used for tekton pipelines
- optimized pipeline and pipelinerun for docker (kaniko) using a dedicated push secret in a workspace
- support for git-clone private otomi Gitea repos

Depends-on: https://github.com/redkubes/otomi-tasks/pull/80


j-zimnowoda (changes requested):
In this PR I do not see that it uses corresponding version from otomi-tasks. Could you refer to it?

j-zimnowoda (approved):
Now, I see that the last commit has already changed back versions to main.

PR #1175, opened 2023-07-04
feat: updated tty image

Ani1357:
This PR updates the tty image to use ubuntu as a base image instead of alpine. This is needed to implement the tmux so that we get the same screen session whenever reaching the tty url(the apline setup was not working as expected).
The usbuntu image uses bash as a default shell which has more support in general. As a bonus we can also use bash completion for kubectl.


j-zimnowoda (approved):
These are wonderful improvements 🚀

PR #1196, opened 2023-07-19
feat: scaleway integration

ElderMatt:
This PR makes it possible to deploy Otomi on scaleway.

See #1180


.github/workflows/integration.yml
```diff
@@ -9,9 +9,13 @@ on:
         description: Otomi installation profile
         default: full
         type: string
+      server_providers:
```
j-zimnowoda, line 12: How about renaming it to `cloud_provider` ?

.github/workflows/integration.yml
```diff
@@ -9,9 +9,13 @@ on:
...
+        description: Server where Otomi will be installed
+        type: string
+        default: scaleway
       cluster_region:
```
j-zimnowoda, line 16: The cloud region is not necessary for dev purpose

.github/workflows/integration.yml
```diff
@@ -43,9 +47,7 @@ on:
         description: 'Kubernetes version'
         type: choice
         options:
-          - "['1.23']"
           - "['1.24']"
-          - "['1.25']"
```
j-zimnowoda, line 48: We need  options:
 - "['1.25']"
 - "['1.26']"

.github/workflows/integration.yml
```diff
@@ -130,6 +138,39 @@ jobs:
...
+    steps:
+      - name: Install scw-cli
+        uses: scaleway/action-scw@v0
+        with:
+          token: ${{ secrets.SCW_ACCESS_KEY }}
+      - id: k8s-versions
+        name: Process k8s version input
+        run: |
+          if [ -z '${{ inputs.kubernetes_versions }}' ]; then
+            echo "Kubernetes versions not specified, determine DO supported versions"
```
j-zimnowoda, line 158: Kubernetes versions not specified, determine `Scaleway` supported versions

.github/workflows/integration.yml
```diff
@@ -130,6 +138,39 @@ jobs:
...
+            versions=`scw k8s versions -o json | jq -ce '.[] | .name'`
```
j-zimnowoda, line 159: It is great that you solved the k8s versioning issue

.github/workflows/integration.yml
j-zimnowoda, line 12: It would be more convininent if this parameter is the firs one after branch selection. Now it is 3rd one.
image

j-zimnowoda (changes requested):
Awesome work. Just few minor remarks and anyone in the team can benefit from your work!

.github/workflows/integration.yml
ElderMatt: Good idea, I'll change it.

.github/workflows/integration.yml
ElderMatt: I'll add them

.github/workflows/integration.yml
ElderMatt: good catch, I'll change it

.github/workflows/integration.yml
ElderMatt: Agree, I'll put it to the top

.github/workflows/integration.yml
ElderMatt: it was tough nut to crack 😄

.github/workflows/integration.yml
ElderMatt: If it is something we don't want to support I can delete it.

.github/workflows/integration.yml
```diff
@@ -115,45 +119,207 @@ jobs:
...
+              cni=calico \
+              pools.0.node-type=${{ env.SCALEWAY_NODE_TYPE }} \
+              pools.0.min-size=${{env.SCALEWAY_NODE_POOL_MIN_SIZE}} \
+              pools.0.size=${{env.SCALEWAY_NODE_POOL_MIN_SIZE}} \
+              pools.0.max-size=3 \
+              pools.0.autohealing=true \
+              pools.0.autoscaling=true \
+              pools.0.name=default \
+              version=${{ env.SCALEWAY_K8s_VERSION }} \
+              region=ams1 \
```
j-zimnowoda, line 225: It seems that the region is incorrect 😬 
```
{"message":"invalid region ams1","error":{},"hint":"Region format should look like XX-XXX (e.g. fr-par)"}
```

See more: https://github.com/redkubes/otomi-core/actions/runs/5609329423

j-zimnowoda (changes requested):

.github/workflows/integration.yml
ElderMatt: It has been fixed, it should now go to nl-ams

j-zimnowoda (approved):

PR #1207, opened 2023-07-28
feat: updating cluster name

Ani1357:
This MR updates the name of the cluster that will be created for better readability.


ElderMatt:
Do you have a preview on how the name looks now in DO or Scaleway?

Ani1357:
> 
This is an example of how it looks
image

So if I run the pipeline not it would be like "Ani1357-28-07-10-57". So its {Github Username}-{Day of Month}-{Month}-{Hour}-{Minutes}

ElderMatt (approved):

j-zimnowoda:
Awesome !

PR #1241, opened 2023-08-21
feat: add sources

srodenhuis:


srodenhuis:
This was a PoC but we decided to get source information from the chart repo and the chart.yaml

PR #1251, opened 2023-08-24
fix: k8s 1.26 compatibility

staticvoid255:


ElderMatt (approved):
Ran a cluster with full settings and it worked.

PR #1306, opened 2023-09-29
fix: falco, jaeger and kubeclarity not shared

srodenhuis:
These apps where set to shared and a filter in console was used to not show them. This PR sets these apps to not shared and a PR in console removes logic for not showing them.


dennisvankekem (approved):
LGTM (I was there when it was fixed)

PR #1352, opened 2023-10-26
fix: github actions scaleway fix

Ani1357:
This MR updates the integration and scheduled backups scripts for scaleway. The integration script will not create a new private network with every otomi deployment. The scheduled clean up script will now delete also the respective private network rather than only the cluster.


ferruhcihan (approved):
LGTM

PR #1354, opened 2023-10-27
fix: scheduled scw private networks cleanup

Ani1357:
This MR fixes the scheduled cleanup of the private networks in scaleway.


ferruhcihan (approved):
LGTM

PR #1366, opened 2023-11-06
feat: add scaleway dns

srodenhuis:
Add support for scaleway dns


srodenhuis:
tested with a n internal domain on scalewat. The only thing is that the webhook container runs as root, so we need to do a PR to the upstream project https://github.com/scaleway/cert-manager-webhook-scaleway

PR #1371, opened 2023-11-12
feat: istio traffic control to support canary

srodenhuis:
This PR adds Istio traffic control between multiple services to Otomi.
To test, create a workload with 2 deployments and one service. Add spec.template.metadata.labels: `version: v1` to deployment 1 and `version: v1` to deployment 2. Make sure the selector labels for both deployments match the service.
In Otomi, create a service and enable trafficControl. Then split the weight between the 2 versions. Set to 50/50 for blue/green or 90/10 for canary.


j-zimnowoda (changes requested):
There is a lot of duplicated code. Let's improve it, as it is hard to follow it now and it will be a burden to maintain it. I can help with that.

j-zimnowoda (approved):

PR #1375, opened 2023-11-17
fix: auto update schema

srodenhuis:
Changes in schema and applicationset template


j-zimnowoda (changes requested):
These changes breaks compatibility with previous version.  Why is it needed?

j-zimnowoda (approved):

PR #1378, opened 2023-11-19
fix: app versions in apps.yaml

srodenhuis:
Update versions and titles of multiple apps.


j-zimnowoda (approved):
Validated with `npm run app-versions:csv `

PR #1407, opened 2023-12-03
fix: upgrade Falco and Falco exporter plus add support for GCP

srodenhuis:
- Upgrade Falco from 0.33.1 to 0.36.2
- Upgrade Falco exporter from 0.8.0 to 0.9.7
- Remove sidekick (not required as we provide admins and teams with a Grafana dashboard)
- Let user select correct driver (default = ebpf)


j-zimnowoda (approved):
Tested also with ebpf on Scaleway and it deploys without any problem.

PR #1408, opened 2023-12-05
fix: oauth2-proxy cert

Ani1357:
This PR fixes the oauth2-proxy ingress configuration when "Bring Your Own Certificate" is in use.

### DIFF
--- a/values/oauth2-proxy/oauth2-proxy-raw.gotmpl
+++ b/values/oauth2-proxy/oauth2-proxy-raw.gotmpl
@@ -42,7 +42,11 @@ resources:
             pathType: Prefix
       {{- if not $v.otomi.hasCloudLB }}
       tls: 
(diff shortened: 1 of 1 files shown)


j-zimnowoda (approved):
Successfully tested.

Methodology:
Deploy Otomi with letsencrypt
Copy certificate secret
Set cert-manage issuer to use byo-wildcard-cert and add certificate and key in the otomi values.

Verify if oauth2-proxy ingress references `byo-wildcard-cert` secret:

```
# k get ingress -n istio-system oauth2-proxy -o jsonpath='{.spec.tls}' | jq
[
  {
    "hosts": [
      "auth.tst-7104154643.aks.redkubes.net"
    ],
    "secretName": "byo-wildcard-cert"
  }
]
```

PR #1429, opened 2023-12-18
feat: select secret for external private repo

srodenhuis:
Now builds can only be created when using a repo in Gitea. To support the use of external private Git repo's, a user can now select to use a secret that contains the credentials of a external (like Github or Gitlab) private repo.
External public repo's and private repo's in Gitea are supported by default.


ferruhcihan (approved):
LGTM

PR #1449, opened 2024-01-03
fix: add env vars for paketo

srodenhuis:
To use Gradle in Paketo buildpacks we need to add some additional variables to the pipeline. Note that the values of these variables are now stored as plain text. This will be solved with the introduction of sealed secrets in Otomi.


Ani1357 (approved):
LGTM

PR #1461, opened 2024-01-08
feat: always install prometheus and alertmanger operator

srodenhuis:
Teams need to be able to activate prometheus and alertmanager independent of the platform prometheus and alertmanager. This feat makes sure the prometheus and alertmanager operators are always installed.

Note: this first needs to be tested. Fixtures have been adjusted for this use-case (and need to be restored)


j-zimnowoda (approved):
tested.

1. minimal install => only operator is installed
2. create team => no impact on monitoring
3. enable team monitoring => team prometheus instance created

PR #1505, opened 2024-02-07
ci: reworked cluster and private network deletion

Ani1357:
This PR tries to fix the daily cleanup task in scaleway by adding a retry mechanism. It will try up to 5 times with a distance of 1 minute if the deletion command fails either for clusters or private networks. This PR also updates the scw cli to version 2.26.0


ElderMatt (approved):
LGTM

PR #1563, opened 2024-03-21
feat: adjustments to Grafana team permissions

merll:
This PR implements the following changes:
- Non-admin users are no longer permitted to log into the platform Grafana instance (which would allow for them to see all logs)
- When team users log into another team's Grafana, they are by default assigned the "Viewer" role (before it was "Editor"). Users are still "Editor" in their own team.
- By setting `teams.[name].managedMonitoring.private` to `true`, only members of that team are allowed to log into this Grafana instance at all. Default is `false` for backwards compatibility.

Depends on redkubes/otomi-api#513


ferruhcihan (approved):
Tested 👍 LGTM

j-zimnowoda (approved):
These are very good ideas. Thanks for moving this forward

PR #1565, opened 2024-03-28
fix: removed service account from values changes

merll:
This PR removes the new field serviceAccount from the value-changes file, as no useful default can be provided, and due to lack of functionality no current installation will currently use Loki together with GCS.

This currently blocks a smooth upgrade from the latest release.

Note there is also redkubes/otomi-api#515, although it is not strictly dependent.


ferruhcihan (approved):
LGTM

PR #1570, opened 2024-04-01
fix: metrics server extra args

srodenhuis:
The extra args for metrics server are set on multiple levels (defaults, custom app, derived) which resulted in incorrect configuration. This PR fixes this.


merll:
Have not tested it on Linode, but syntactically looks good.

merll (approved):

PR #1585, opened 2024-04-15
fix: rm opencost

srodenhuis:
This PR removes the opencost app. See related PRs in api and console


merll (approved):

CasLubbers (changes requested):
Can the opencost grafana dashboard also be removed? https://github.com/redkubes/otomi-core/tree/main/charts/grafana-dashboards/opencost?
Also in the values here its mentioned: https://github.com/redkubes/otomi-core/blob/main/charts/grafana-dashboards/values.yaml#L13

CasLubbers (approved):

PR #1602, opened 2024-05-02
feat: remove marketplace integrations

ElderMatt:
This PR removes marketplace integrations from the workflows and updates github actions to newer versions


j-zimnowoda (approved):
Thanks for being proactive and updating the versions.

PR #1616, opened 2024-05-28
fix: handle version prefix in tools version increment script [TOOLS][MAJOR]

CasLubbers:
Script now handles the `v` version prefix and updates correctly.


Ani1357 (approved):
LGTM

PR #1618, opened 2024-05-30
feat: removing hashicorp vault and external-secrets

Ani1357:
## Checklist


versions.yaml
```diff
@@ -1,4 +1,4 @@
-api: main
-console: main
+api: removing_vault
```
CasLubbers, line 1: Don't forget to revert this after merging or just before merging.

upgrades/pre/upgrade-2-12-0.sh
```diff
@@ -0,0 +1,7 @@
+#!/bin/bash
+
+set -eu
+
+if [[ $(helm status -n external-secrets external-secrets-artifacts 2>/dev/null) ]]; then
+  helm uninstall -n external-secrets external-secrets-artifacts
```
CasLubbers, line 6: Do we also need to uninstall te following?:
```
external-secrets
vault-operator
vault-operator-artifacts
```
These helm charts were still installed after upgrading is that correct?

CasLubbers (changes requested):
~Vault and external secrets are removed from the UI only the helm charts still exist in the cluster is that correct?~
Yes it was correct we can merge it :)

CasLubbers (approved):

PR #1623, opened 2024-06-05
feat: improve robustness of initial install

merll:
This PR includes the following changes for making the installation process more robust against temporary connection issues with the Kubernetes API.
* `helm` has been upgraded to the latest release. Since v3.15.0, it has a retry-mechanism to reattempt Kubernetes API requests on certain HTTP status codes or connection timeouts. This is currently hard-coded to 30 retries.
* `helmfile` and the `helm-diff` plugin have been upgraded for ensuring compatibility, while `kubectl` has been upgraded for a closer match to the supported cluster version.
* On initial install, the total amount of queries is reduced by
  - skipping validation against the OpenAPI spec of the cluster's Kubernetes API
  - bypassing `helm diff` and proceeding to a Helm chart install directly
  - running one `helm` instance at a time.
* Also on the regular pipeline, Helm has a rate limiting of 20 queries per second applied. This is most likely only to affect read queries such as CRDs.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM otomi/tools:v2.0.3 as ci
+FROM otomi/tools:v2.2.0 as ci
 
 ENV APP_HOME=/home/app/stack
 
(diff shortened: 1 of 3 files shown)


CasLubbers (approved):
LGTM! ✅ 
Tested it: installation works and after installation I enabled all the apps in one deploy also worked.

PR #1640, opened 2024-07-08
feat: one backend obj for all apps

srodenhuis:
This PR aims to make the configuration of Object Storage and Backups more consistent and easier to maintain.

OBJ is now configured in the platform settings. A user can choose between disabled (no OBJ is configured), minioLocal (for PoC/dev use-cases) and Linode (using Linode OBJ service). Integrated applications like Harbor, cnpg, Loki, Tempo and Velero, consume the OBJ configuration and do not need to be configured independently.

With this PR we do not support any other OBJ service. If a user would like to use AWS S3 or Azure Blob, then the user needs to configure this using rawValues per app.

Other changes:
- Backup schedules for PVs of core apps have been removed
- Velero now only support Linode cloud. Users can use rawValues to configure other infrastructure/cloud providers to create pv backups and snapshots
- Linode is now the default `cluster.provider`. Now there are only 2 providers: `linode` and `custom`
- Linode API key (needed to create volume snapshots) is added to the `settings.backup` section
- All OBJ configuration in API and schema has been removed for `loki`, `velero`, `tempo`, `harbor` and `cnpg`

NOTE:
There is no option in the velero chart to define an additional `env` using a `secretKeyRef`. I made some changes to the chart and will submit a PR to support Linode Velero plugin in Velero Helm chart. See here for more info about the Linode Velero plugin: https://github.com/linode/velero-plugin/tree/main?tab=readme-ov-file


values/harbor/harbor-otomi-db.gotmpl
```diff
@@ -1,7 +1,8 @@
 {{- $v := .Values }}
 {{- $hdb := $v.databases.harbor }}
 {{- $cnpg := $v.apps.cnpg }}
-{{- $sp := $cnpg.storage }}
+{{- $obj := $v.obj.provider }}
+{{- $bu := $v.obj.bucket }}
```
Ani1357, line 5: This should prevent installation errors when obj.provider is set to "disabled"

```suggestion
{{- if ne $obj.type "disabled" }}
  {{- $bu := $v.obj.bucket }}
{{- end }}
```

values/gitea/gitea-otomi-db.gotmpl
```diff
@@ -2,10 +2,31 @@
 {{- $k := $v.apps.gitea }}
 {{- $cnpg := $v.apps.cnpg }}
 {{- $gdb := $v.databases.gitea }}
+{{- $obj := $v.obj.provider }}
+{{- $bu := $v.obj.bucket }}
```
Ani1357, line 6: This should prevent installation errors when obj.provider is set to "disabled"

```suggestion
{{- if ne $obj.type "disabled" }}
  {{- $bu := $v.obj.bucket }}
{{- end }}
```

values/keycloak/keycloak-otomi-db.gotmpl
```diff
@@ -1,11 +1,33 @@
 {{- $v := .Values }}
 {{- $k := $v.apps.keycloak }}
 {{- $cnpg := $v.apps.cnpg }}
+{{- $sp := $cnpg.storage }}
 {{- $kdb := $v.databases.keycloak }}
+{{- $obj := $v.obj.provider }}
+{{- $bu := $v.obj.bucket }}
```
Ani1357, line 7: This should prevent installation errors when obj.provider is set to "disabled"

```suggestion
{{- if ne $obj.type "disabled" }}
  {{- $bu := $v.obj.bucket }}
{{- end }}
```

Ani1357: Solved in [c3e1a9d](https://github.com/linode/apl-core/pull/1640/commits/c3e1a9d623f105fe0f37588d124edebb808cf737)

Ani1357 (approved):
LGTM

PR #1658, opened 2024-07-24
fix: sni hosts

srodenhuis:
This PR fixes the wrong `*` sniHosts when a `CNAME` is used. Using `*` is not supported anymore so we need to set the hosts


Ani1357 (approved):
LGTM.Tested creating service with cname in dev cluster.

PR #1661, opened 2024-07-29
feat: thanos sidecar

srodenhuis:
See APL-132


srodenhuis:
replaced by https://github.com/linode/apl-core/pull/1729

PR #1663, opened 2024-07-30
fix: argocd defaults

srodenhuis:
ArgoCD defaults for new defaults approach


j-zimnowoda (approved):
I have verified the rendered manifests and I see autoscaling being set correctly. 
The resources are also set accordingly to the values.
I really appreciate your findings and the implementation.

PR #1680, opened 2024-08-19
fix: loki storage config without OBJ

srodenhuis:
When OBJ is not configured (or set to disabled), the Loki storageConfig is now configured to use filesystem. This allows to use Loki without OBJ.


ElderMatt:
At the moment the validate-templates is failing on the following: 
image

ElderMatt:
It fixes the Loki errors, but when performing full install it does give the following error for velero. Is this something we want to fix in this PR or make a new PR for it? 
image

ElderMatt (approved):
LGTM

PR #1693, opened 2024-08-27
fix: nginx defaults

srodenhuis:
Migrate API schema defaults to core defaults for nginx


CasLubbers (approved):
Tested and nginx worked as aspected ✅

PR #1694, opened 2024-08-27
fix: enable argo metrics

srodenhuis:
Enable metrics for ArgoCD


CasLubbers (approved):
ArgoCD metrics are available in Prometheus ✅

PR #1701, opened 2024-08-30
feat: add resource finalizers to argocd applications

CasLubbers:
Added resource finalizer to ArgoCD so that applications resources also get cleaned up


CasLubbers:
This is not working as expected

CasLubbers:
Superseded by: https://github.com/linode/apl-core/pull/1727#pullrequestreview-2324693859

PR #1705, opened 2024-09-04
fix: rm civo

srodenhuis:
- Remove webhook for civo dns
- Remove Drone slack notifications configuration


j-zimnowoda (approved):
I fixed migrations. LGTM

PR #1708, opened 2024-09-09
feat: support for Akamai EdgeDNS

merll:
implements [APL-172](https://jira.linode.com/browse/APL-172)


Ani1357:
Tested by using API credentials from the shared service account API client and all works as expected.
Can you also update the value-changes file?

srodenhuis:
It is not a good practice to change the upstream chart. I see external-dns chart has an Akamai provider. Why are the changes in the chart needed?

merll:
> It is not a good practice to change the upstream chart. I see external-dns chart has an Akamai provider. Why are the changes in the chart needed?

The changes were not needed. It just seemed like a better practice to move some more of the API client configuration into a secret, as this is also done in the `cert-manager` configuration. But since these two secrets are stored separately in each namespace, there is no need to, and not worth breaking compatibility with upstream.

I have reverted these changes accordingly.

merll:
> Tested by using API credentials from the shared service account API client and all works as expected. Can you also update the value-changes file?

What should go into the value-changes? I thought for new values this was only for setting defaults. But AFAIK there are no useful defaults to be set here, since it is all individual user input.

Ani1357:
> > Tested by using API credentials from the shared service account API client and all works as expected. Can you also update the value-changes file?
> 
> What should go into the value-changes? I thought for new values this was only for setting defaults. But AFAIK there are no useful defaults to be set here, since it is all individual user input.

You are right, please ignore my comment. Indeed we are only setting defaults there.

srodenhuis:
You should also add the provider to the API: https://github.com/linode/apl-api/blob/main/src/openapi/definitions.yaml

merll:
> You should also add the provider to the API: https://github.com/linode/apl-api/blob/main/src/openapi/definitions.yaml

Done: linode/apl-api#559

Ani1357 (approved):

PR #1712, opened 2024-09-12
fix: rm private ingress

srodenhuis:
Remove private ingress class option (which was Azure specific)


Ani1357 (approved):
Tested by deploying from branch sr-rm-private-ingress(apl-core and apl-api):
- Private ingress option is gone: ✅
- Annotations can be applied to LB svc as expected: ✅

PR #1719, opened 2024-09-16
fix: ex dns resources and dns apiToken Linode

srodenhuis:
- add defaults for external-dns resources
- fix Linode DNS apiToken


Ani1357 (approved):
LGTM

PR #1721, opened 2024-09-17
fix: increase mem limits of apl-gitea operator

srodenhuis:
Increase memory limits for apl-gitea-operator because the operator runs on 80% Limit constantly


ElderMatt (approved):
LGTM

PR #1723, opened 2024-09-23
feat: gitea backup to object storage

merll:
Implements [APL-121](https://jira.linode.com/browse/APL-121), as running the [backup for Gitea](https://docs.gitea.com/administration/backup-and-restore) and uploading to object storage.


values/gitea/gitea-raw.gotmpl
```diff
@@ -37,3 +39,104 @@ resources:
...
+            containers:
+              - image: bitnami/kubectl:1.30
+                name: kubectl
+                command:
+                  - /bin/sh
+                  - -ec
+                  - >-
+                    kubectl exec gitea-0 -- /bin/sh -ec "
+                    if [ ! -f '/backup/.bin/rclone' ]; then
+                      echo 'Installing RClone...' &&
```
CasLubbers, line 108: We can maybe build our own image bitnami/kubectl:1.30 and add rclone to it. This way you don't need to install it every time the cronjob is ran.

merll: No, this will not solve it. The Gitea container needs access to the backup and is therefore also running Rclone, not the Cronjob/Kubectl one. Maintaining our own Gitea image with Rclone is probably not worth the effort. A more "clean" but way more complex solution would be to add a sidecar to the Gitea container. However, that would require modifying the upstream chart, which I have been trying to avoid.

Note also, that Rclone is only installed on the first cronjob run.

CasLubbers: Oh I totally read over the kubectl exec. Was already wonder why that if check is there if the cronjob would restart every time. Thanks for the response.

CasLubbers (approved):
Code looks good to me!

PR #1727, opened 2024-09-24
feat: argo app removal

Ani1357:
This PR ensures that when an app is disabled in the values repo by setting the value of `apps.<app-name>.enabled` to `false`, its corresponding argo application and associated resources are removed from the cluster.

How to test:
1. enable loki via apl console
2. deploy changes
3. wait until the application is deployed
4. go to values repo and set `apps.loki.enabled: false`
5. Observe that corresponding ArgoCd application is removed
6. Observe that corresponding k8s resources are removed


CasLubbers:
Ah I didn't know we removed the finalizers just before deleting the application. What I previously did was setting the finalizers on the application yaml when creating it. That way people can like remove the finalizers if they want to keep stuff but delete the ArgoCD application. Now we will always add the finalizers even if the user doesn't want it. Is that something we want?

Ani1357:
> Ah I didn't know we removed the finalizers just before deleting the application. What I previously did was setting the finalizers on the application yaml when creating it. That way people can like remove the finalizers if they want to keep stuff but delete the ArgoCD application. Now we will always add the finalizers even if the user doesn't want it. Is that something we want?

We are not removing but rather adding the finalizers right before the deletion as described in the [argocd docs](https://argo-cd.readthedocs.io/en/stable/user-guide/app_deletion/#deletion-using-kubectl). In theory your approach should also work but I only tested this one.

CasLubbers (approved):
As discussed in standup. If we disable an application we always want to remove all the resources.

PR #1735, opened 2024-09-27
feat: updating apl-tools [TOOLS][MINOR]

Ani1357:
This PR updates the tools in the apl-tools image. Once merged this commit will create the new 2.7.0 image version via the "APL Tools Build and Versioning" pipeline. This means the first run of the "Build and publish Docker" pipeline will fail since the new 2.7.0 image will not be ready. Rerunning it should solve the issue.


tools/Dockerfile
```diff
@@ -4,29 +4,29 @@ FROM ubuntu:20.04 as builder
...
+ARG KUBECTL_VERSION=1.31.1
 # https://github.com/helm/helm/tags
-ARG HELM_VERSION=3.15.1
+ARG HELM_VERSION=3.16.1
 # https://github.com/databus23/helm-diff/releases
-ARG HELM_DIFF_VERSION=3.9.7
+ARG HELM_DIFF_VERSION=3.9.11
 # https://github.com/jkroepke/helm-secrets/releases
-ARG HELM_SECRETS_VERSION=3.15.0
+ARG HELM_SECRETS_VERSION=4.6.1
```
CasLubbers, line 13: This upgrade has breaking changes. https://github.com/jkroepke/helm-secrets/blob/main/CHANGELOG.md#400---2022-09-11
I think we can solve this by updating `src/common/crypt.ts`

```
enum CryptType {
  ENCRYPT = 'helm secrets encrypt -f',
  DECRYPT = 'helm secrets decrypt -f',
  ROTATE = 'sops --input-type=yaml --output-type=yaml -i -r',
}
```

CasLubbers, line 13: This upgrade has breaking changes. https://github.com/jkroepke/helm-secrets/blob/main/CHANGELOG.md#400---2022-09-11
I think we can solve this by updating `src/common/crypt.ts`

```
enum CryptType {
  ENCRYPT = 'helm secrets encrypt -f',
  DECRYPT = 'helm secrets decrypt -f',
  ROTATE = 'sops --input-type=yaml --output-type=yaml -i -r',
}
```

Ani1357: Reverted to 3.16 which seems to be latest version. Somehow I ended up in the unreleased section.

CasLubbers (approved):
Let's create a ticket for specifically upgrading helm secrets as that requires some breaking changes

PR #1742, opened 2024-10-01
feat: k8s 1.31 upgrade

CasLubbers:
## Checklist


.github/workflows/integration.yml
```diff
@@ -55,10 +55,10 @@ on:
         description: 'Kubernetes version'
         type: choice
         options:
-          - "['1.28']"
           - "['1.29']"
           - "['1.30']"
-        default: "['1.30']"
+          - "['1.31']"
+        default: "['1.31']"
```
ferruhcihan, line 61: line 12 should also be `default: "['1.31']"`

ferruhcihan (changes requested):
These two files also includes `kubernetes_versions` fields:
- .github/workflows/integration-on-schedule-upgrade.yml
- README.md

.github/workflows/integration.yml
CasLubbers: Updated the integration.yml
I think the README.md we only update with the release. Otherwise we get that issue again with a user saying we support k8s 1.31 but the latest release does not support it.

ferruhcihan (approved):
LGTM, Tested with a cluster 👍

PR #1751, opened 2024-10-09
fix: add resources to team config in integration profiles

srodenhuis:


j-zimnowoda (changes requested):
We decided to revert previous changes and open a PR that contains fixes and sort out problem with team defaults.

PR #1760, opened 2024-10-14
feat: ispreinstalled flag

ElderMatt:
This PR adds isPreInstalled flag for the installer.
See: [APL-134](https://jira.linode.com/browse/APL-134)

Also removed Minio dependency for everything


.github/workflows/integration.yml
```diff
@@ -86,7 +90,14 @@ on:
           - letsencrypt_staging
           - letsencrypt_production
         default: letsencrypt_production
-
+      is_pre_installed:
```
CasLubbers, line 93: I would add this as an `install_profile`, something like `minimal-pre-installed`. And maybe set it as the default?

.github/workflows/integration.yml
```diff
@@ -86,7 +90,14 @@ on:
...
+        type: choice
+        description: Fake if Otomi is pre-installed by Installer
```
CasLubbers, line 95: Why `Fake`? Do you mean `False`?

.github/workflows/integration.yml
```diff
@@ -274,12 +286,29 @@ jobs:
           CLIENT_SECRET: ${{ secrets.EDGEDNS_CLIENT_SECRET }}
         run: |
           touch values.yaml
+          [[ '${{ inputs.dns }}' == 'az_dns' ]] && echo "$AZ_DNS" >> values.yaml && domainSuffix="--set cluster.domainSuffix=tst-${{ github.run_id }}.aks.redkubes.net"
```
CasLubbers, line 289: Are these needed? I think we do not use azure anymore right?

CasLubbers (dismissed):

.github/workflows/integration.yml
ElderMatt: This flag is supposed to be set by the APL-Installer, I added this option as a way to fake it for easier testing.

.github/workflows/integration.yml
ElderMatt: Good find, I think this because a merge conflict.

.github/workflows/integration.yml
CasLubbers: Oh I read the sentence completely wrong, apologies.

.github/workflows/integration.yml
ElderMatt: would this in addition to a workflow parameter or instead of the workflow parameter?

ferruhcihan (approved):
LGTM

PR #1779, opened 2024-10-27
fix: add init container for oauth2 proxy

srodenhuis:
Add an additional init container to oauth2 proxy to check availability of OIDC issuer URL before starting oauth2 proxy

### DIFF
--- a/charts/oauth2-proxy/templates/deployment.yaml
+++ b/charts/oauth2-proxy/templates/deployment.yaml
@@ -100,6 +100,9 @@ spec:
       {{- if .Values.terminationGracePeriodSeconds }}
       terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
       {{- end }}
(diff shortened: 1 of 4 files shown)


Ani1357 (approved):
Tested and it works as expected.

CasLubbers (approved):
Works as expected ✅

charts/oauth2-proxy/templates/deployment.yaml
```diff
@@ -100,6 +100,9 @@ spec:
       {{- if .Values.terminationGracePeriodSeconds }}
       terminationGracePeriodSeconds: {{ .Values.terminationGracePeriodSeconds }}
       {{- end }}
+{{- if .Values.extraInitContainers }}
```
CasLubbers, line 103: Maybe a nice to have is to add extra comment here so we know that this is our custom implementation and needs to be re-added if we update the chart

PR #1780, opened 2024-10-28
fix: Trivy indent and capabilities

srodenhuis:
The upgrade of Trivy operator will be postponed. This PR fixes the resource configuration and disables cluster compliance because operator is only used for teams.


Ani1357 (approved):
Tested by templating trivy-operator manifests. Works as expected.

CasLubbers (approved):
✅

PR #1784, opened 2024-10-29
Cnpg upgrade

Ani1357:
Updating Cloud Native PG chart and using postgresql 17.0


values/gitea/gitea-otomi-db.gotmpl
```diff
@@ -9,6 +9,7 @@ name: gitea-db
 storage:
   size: {{ $gdb.size }}
 instances: {{ $gdb.replicas }}
+imageName: ghcr.io/cloudnative-pg/postgresql:15.10
```
CasLubbers, line 12: Do we also want gitea and keycloak to be default at 17.0? And only leave the clusters that currently are on 15 stay on 15?
Then we don't have to upgrade that much clusters manually...

charts/otomi-db/values.yaml
```diff
@@ -3,7 +3,7 @@ labels:
   prometheus: system
   helm.sh/resource-policy: keep
 instances: 2
-imageName: ghcr.io/cloudnative-pg/postgresql:15.3
+imageName: ghcr.io/cloudnative-pg/postgresql:17.0
```
CasLubbers, line 6: 17.2 is currently the latest. I think its best to always pick the latest patch version.
```suggestion
imageName: ghcr.io/cloudnative-pg/postgresql:17.2
```

CasLubbers: see: https://github.com/cloudnative-pg/postgres-containers/pkgs/container/postgresql

values-changes.yaml
```diff
@@ -319,3 +319,7 @@ changes:
   - version: 32
     deletions:
       - 'teamConfig.{team}.managedMonitoring.prometheus'
+    additions:
```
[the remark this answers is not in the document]
Ani1357: From which branch are you updating? It should work if you upgrade from main or latest release.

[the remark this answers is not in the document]
merll: It entirely depends on what `settings.version` is. On a fresh install on the current main, that is already `32` and will not run this migration again. Usually we only need a new version on a release, but his particular case can make it tricky to switch between feature branches.

values-changes.yaml
```diff
@@ -319,3 +319,7 @@ changes:
...
+      - databases.gitea.imageName: ghcr.io/cloudnative-pg/postgresql:15.10
```
merll, line 323: The problem with adding the image specifically here is, that it will make it difficult to deliver minor updates, and also make other necessary changes to come (e.g. should the registry or repo name ever change). I would prefer only keeping the major version recorded here. We could then maintain specific images using a ClusterImageCatalog (https://cloudnative-pg.io/documentation/1.24/image_catalog/)

values-schema.yaml
```diff
@@ -797,7 +797,7 @@ definitions:
       - name
       - type
   backupSchedule:
-    default: 0 0 * * *
+    default: 0 0 0 * * *
```
merll, line 800: I am not sure why this is changed, but it does not look right to me (comparing with https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#schedule-syntax). But we might want to remove this line and move it to defaults. If we make any change here, a regex validation could come in useful. But it's not really related to the purpose of this PR.

merll:
Please also update `chart/chart-index/Charts.yaml` with the new chart version number.

values-schema.yaml
merll: Confusing docs there. They say that it follows the Kubernetes CronJobs expression (which is 5 digits) in the description, but the link leads to Go-library docs. I would suggest we stick to the 5-digit format and update the templates to set the first digit to `0`, for the following reasons:
* Many customers who already have used this might have this wrong written to their values. This is tricky to migrate with the methods we currently have.
* We state "Cron-type expression" in the field description. This is misleading without explicitly referencing these Go-library docs.
* We use CronJobs in other places (e.g. Gitea backup, Kyverno) and it would be inconsistent.
* There is no added value in adding second-precision. There is no practical use for it.

schemas/api-versions/1.31.txt
```diff
@@ -1,22 +1,66 @@
+acme.cert-manager.io/v1
```
merll, line 1: AFAIK only core APIs should be in here.

schemas/gen-k8s-schemas.sh
```diff
@@ -10,6 +10,8 @@ set -ex
 #   X.Y.Z-local - relative references, useful to avoid the network dependency
 
 declare -a K8S_VERSIONS=(
+  v1.31.3
```
merll, line 13: Please don't include unrelated changes here.

values/cloudnative-pg/cloudnative-pg-raw.gotmpl
```diff
@@ -0,0 +1,11 @@
+resources:
+  - apiVersion: postgresql.cnpg.io/v1
+    kind: ClusterImageCatalog
+    metadata:
+      name: postgresql
```
merll, line 5: Now thinking of it, I would use a different name here, e.g.
```suggestion
      name: platform
```

schemas/gen-k8s-schemas.sh
Ani1357: reverted changes

schemas/api-versions/1.31.txt
Ani1357: reverted changes

values/cloudnative-pg/cloudnative-pg-raw.gotmpl
Ani1357: applied suggestion

values-schema.yaml
Ani1357: Updated it to use 5digit cron definitions. The additional 0 will be added in the otomi-db cluster definition.

Ani1357:
Updated PR so that:
- existing APL installations will be upgraded to postgresql v15.10 via the `values-changes` file
- new installations will use imageCatalog (v17 bu default) instead of an exact imageversion(via `imageName` argument)

merll:
This might work, but I find that the use of `imageName` and the PG major version with these conditions in templates makes it less transparent what is happening. This can lead to new trouble, especially in case the migration or other values changes do not go through as intended. Also it introduces two values to the schema, of which one is legacy already.

In order to get this PR through, I would suggest that we move everything involving the image catalog and the PG major upgrade, including values changes, to a separate PR so we can sort things out later. We can upgrade all installations to 15.10 for now, just by changing the image name. When coming back to this, if just ArgoCD's patch mechanism is causing the problem, we can figure out how to do this with a one-off command in an upgrade script, using `kubectl`.

For the backup, we can merge it with this PR. It solves a current issue and also seems backwards compatible with existing installations now.

j-zimnowoda (approved):

values/cloudnative-pg/cloudnative-pg-raw.gotmpl
```diff
@@ -0,0 +1,11 @@
+resources:
+  - apiVersion: postgresql.cnpg.io/v1
```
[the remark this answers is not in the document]
Ani1357: This resource needs to be created only once. Having this in the otomi-db chart would attempt to create/recreate the resource for every database we use (harbor, keycloak, gitea) and I'm not sure what the effects of that would be in the argocd application sync.

CasLubbers (approved):

PR #1795, opened 2024-11-04
fix: change default falco driver

srodenhuis:
Set default Falco driver to modern-bpf


CasLubbers (approved):

Ani1357 (approved):
tested by deploying in

PR #1800, opened 2024-11-06
fix: anti affinity between gitea keycloak and master redis

srodenhuis:
This PR introduces anti affinity between Gitea and the Oauth2 master redis pods to assure all pods are scheduled on different nodes.

### DIFF
--- a/values/gitea/gitea.gotmpl
+++ b/values/gitea/gitea.gotmpl
@@ -28,6 +28,19 @@ podDns:
 
 resources: {{- $g.resources.gitea | toYaml | nindent 2 }}
 
+affinity:
+  podAntiAffinity:
+    preferredDuringSchedulingIgnoredDuringExecution:
+      - podAffinityTerm:
+          labelSelector:
+            matchExpressions:
+              - key: otomi.io/affinity
+                operator: In
+                values:
+                  - anti
+          topologyKey: kubernetes.io/hostname
+        weight: 100
+
 securityContext:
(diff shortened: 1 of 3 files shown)


values/oauth2-proxy/oauth2-proxy.gotmpl
```diff
@@ -110,7 +110,20 @@ redis:
       sidecar.istio.io/inject: "false"
     priorityClassName: otomi-critical
     resources: {{- $r.resources.master | toYaml | nindent 6 }}
-
+    podLabels:
+      otomi.io/affinity: "anti"
+    affinity:
```
merll, line 115: This replaces the (positive) default pod affinity matching labels:
```
matchLabels:                                
  app.kubernetes.io/instance: oauth2-proxy  
  app.kubernetes.io/name: redis             
  app.kubernetes.io/component: master       
```
I currently do not even see what this would be even good for (more than one master for the same Redis instance on the same node?). Just wanted to double-check if you have an idea about this, and if this could become an issue.

values/gitea/gitea.gotmpl
```diff
@@ -28,6 +28,19 @@ podDns:
 
 resources: {{- $g.resources.gitea | toYaml | nindent 2 }}
 
+affinity:
+  podAntiAffinity:
```
merll, line 32: Unfortunately this fails in the upgrade scenario, since updates are not allowed to a running stateful set.

PR #1805, opened 2024-11-08
feat: pre-fill repo URLs

srodenhuis:
- pre-fill repo URLs when repo type is internal
- pre-fill secrets when repo type is external and private


CasLubbers:
Closed this one as we now have coderepos and new build page

PR #1812, opened 2024-11-14
feat: APL-372 updated cnpg endpointurl

Ani1357:
This PR fixes the "backup size growing indefinitely" issue.
This PR updates the endpointUrl for the object storages where the backups and WAL files are saved.
Fixing this result in a successful backup and also a successful enforcement of the retention policy.
In this PR we are also enabling the WAL and Data compression to make the backup smaller.
Tested on a new cluster by setting the retention policy to 1 day.

### DIFF
--- a/charts/otomi-db/templates/cluster.yaml
+++ b/charts/otomi-db/templates/cluster.yaml
@@ -46,6 +46,10 @@ spec:
         secretAccessKey:
           name: minio-creds
           key: MINIO_SECRET_KEY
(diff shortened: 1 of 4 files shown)


CasLubbers (approved):

j-zimnowoda (approved):
Great work!
After upgrade I see that backups are finally completed :)
image

It is worth nothing that the number of walls decreased, which is also expected.

PR #1818, opened 2024-11-19
fix: skip assignment of team-admin to argocd admin roles

merll:
This PR changes the ArgoCD AppProject definition for `team-admin` to not assign the `team-admin` role, i.e. only allowing platform admin to access the project.


Ani1357 (approved):
Tested by upgrading production cluster to this branch. As expected, after upgrade I was not able to see the `team-admin` ArgoCD Project when logged in as a team admin.

j-zimnowoda (approved):
I compared against main branch and it looks good.
image

PR #1835, opened 2024-12-03
feat: add path for gitea

CasLubbers:
- added path for gitea
Is related to:
- https://github.com/linode/apl-api/pull/601
- https://github.com/linode/apl-console/pull/482


Ani1357 (approved):
Tested by upgrading core, api and console on existing cluster and accessing gitea afterwards. Works as expected.

dennisvankekem (approved):
see comment in console for small note

PR #1850, opened 2024-12-10
fix: update values schema to adhere k8s naming

CasLubbers:
Has to be tested with:
- https://github.com/linode/apl-api/pull/604
- https://github.com/linode/apl-console/pull/485#pullrequestreview-2492362921


j-zimnowoda (approved):
I upgraded one cluster to see if it does not conflict. Everything was fine.

Of course there is a slight chance that an existing cluster values does not adhere to the newis regex but if that is the case then kubeapi would not deploy that resources anyway.

Ani1357 (approved):
Tested by upgrading existing cluster.
Created new team.
Created new workload.
Enabled additional apps.

Everything worked as expected.

PR #1856, opened 2024-12-11
feat: APL-281 lower database memory footprint

Ani1357:
This PR updates the memory limits and request for  the platform apps cnpg databases in order to lower its footprint.
This PR also adds a migration in the values-changes in order to update the values for existing APL clusters.


helmfile.d/snippets/defaults.yaml
```diff
@@ -1266,21 +1266,21 @@ environments:
             resources:
               limits:
                 cpu: "1"
-                memory: 1Gi
+                memory: 192Mi
               requests:
                 cpu: 50m
-                memory: 512Mi
+                memory: 64Mi
```
merll, line 1272: Would it not be better to set the requests equal with the limit, for increasing the QoS class (https://cloudnative-pg.io/documentation/1.24/resource_management/)?

j-zimnowoda (approved):
This change is a significant resource consumption reduction. Around ~2GB less memory required for each APL instance. The postgresql parameters seems to be reasonable

values-changes.yaml
```diff
@@ -319,3 +319,10 @@ changes:
   - version: 32
     deletions:
       - 'teamConfig.{team}.managedMonitoring.prometheus'
+    mutations:
```
CasLubbers, line 322: These mutations are not same as the defaults is that correct?

CasLubbers (changes requested):

CasLubbers (approved):

PR #1871, opened 2024-12-17
feat: cnpg upgrade fix

Ani1357:
For both this PRs([APL-282](https://github.com/linode/apl-core/pull/1784), [APL-281](https://github.com/linode/apl-core/pull/1856)) the upgrade was tested separately, so we missed the argo application sync error:
image
Apparently we can update the postrgresql image version or the postrgresql parameters but not both at the same time.

In order to solve this I propose we run 2 values schema changes in 2 different minor version.
The first (version: 32) will pin the postgresql image version for the platform apps(harbor, gitea, keycloak) to version 15.3(actual version) and will update its resource requests/limits along with some [postgresql parameters](https://cloudnative-pg.io/documentation/1.16/postgresql_conf/#the-postgresql-section).  

The next changes (version: 33), can update the imageName value using mutations  .

### DIFF
--- a/helmfile.d/snippets/defaults.yaml
+++ b/helmfile.d/snippets/defaults.yaml
@@ -1335,4 +1335,4 @@ environments:
           upgrade:
             version: main
         # TODO: update this when schema version changes
-        version: 32
+        version: 33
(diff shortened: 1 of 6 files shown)


j-zimnowoda (approved):
I tested the upgrade. 
No issue with DB cluster sync in ArgoCD and values are migrated properly
image

image

image

j-zimnowoda:
image

PR #1892, opened 2025-01-10
chore(chart-deps): update harbor to version 1.16.1

svcAPLBot:
$'This PR updates the dependency **harbor** to version **1.16.1**.'


Ani1357 (approved):
Tested lab scenarios before and after the upgrade. Everything worked as expected.

j-zimnowoda (commented):
I compared rendered manifests and see the following changes:
- Sts harbor-redis is using volumeClaimTemplates instead of PVC
- Registry root password is not in a separate secret `harbor-registry-htpasswd` instead of `harbor-registry`

j-zimnowoda (approved):

PR #1896, opened 2025-01-17
chore(chart-deps): update cert-manager to version v1.16.3

svcAPLBot:
$'This PR updates the dependency **cert-manager** to version **v1.16.3**.'


Ani1357 (approved):
Tested upgrade path. Tested also deleting one of the self-signed certs in the cluster. It was recreated successfully.

merll (approved):

PR #1899, opened 2025-01-17
feat: fix open redirect vulnerability

Ani1357:
This PR fixes an open redirect [vulnerability](https://track.akamai.com/jira/browse/AKAMBB-67) affecting logged in users. The issue was caused by the nginx config in the oauth2-proxy ingress resource. 
The redirect will be allowed only for domains matching a regex expression like the one below:
`^([a-zA-Z0-9-]+\.){0,2}lke317513\.akamai-apl\.net(%2F)?(/.*)?$`
The `lke317513\.akamai-apl\.net` is taking the value of cluster.domainSuffix provided in the apl installer chart values.
You can test the expression here: https://regex101.com/r/Yk6RMh/1


CasLubbers (changes requested):
It works for linode clusters with the clusterId. But does this also work if you do a manual install with your own domain? I haven't tried it but curious. Maybe we can keep it like this for Linode/managed installation but make it less strict if you use something else?

Ani1357:
> It works for linode clusters with the clusterId. But does this also work if you do a manual install with your own domain? I haven't tried it but curious. Maybe we can keep it like this for Linode/managed installation but make it less strict if you use something else?

Good point. Updated the code so that it fits both scenarios. The regex expression is not directly constructed from the domainSuffix value provided in the apl chart values.

CasLubbers (approved):

PR #1900, opened 2025-01-20
feat: added more charts to the chart-index

Ani1357:


j-zimnowoda (approved):
The registries are correct. 
I run it locally and the scripts update correct directories in the charts/*

However I noticed that `tempo-distributed` has corresponding charts/tempo directory.
It seems that we need to introduce a mapping in the chart index.

CasLubbers (approved):

PR #1927, opened 2025-02-05
fix: check for helm secrets version and update it

ElderMatt:
This PR updates the helm secrets package to the correct version for local use


CasLubbers (approved):

CasLubbers:
Was also thinking about the issue you had. In the apl-core code for encrypt/decrypt it's maybe better to check if the library is there otherwise give an error.

ElderMatt:
That would also be a good addition. Perhaps we can run a couple of check before running `npm run server`, `npm bootstrap-dev` or `npm run bootstrap-dev-with-repo` for example.

ferruhcihan (approved):

PR #1931, opened 2025-02-06
ci: add charts and alias

merll:
This PR adds some more charts to the index. It also uses the `alias` to identify charts that we have placed in different directories. In addition it only fetches updates for one repo at a time, instead of all of them each time.


Ani1357 (approved):
Tested on a branch with downgraded charts and it works as expected.

PR #1940, opened 2025-02-13
fix: unique admin password for gitea

merll:
Reintroduces #1910, with a fix for the consequence of Git restarting.

### DIFF
--- a/helmfile.d/snippets/defaults.gotmpl
+++ b/helmfile.d/snippets/defaults.gotmpl
@@ -7,7 +7,10 @@
 environments:
   default:
     values:
-    - teamConfig:
+    - apps:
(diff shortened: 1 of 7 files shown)


ElderMatt (approved):
Tested on cluster by upgrading from an old release (v4.2.4) to this branch, I can now see the message where it is Retrying after 30000ms. After that it is working correctly

Ani1357 (approved):
Tested by upgrading existing cluster(with some teams and users already present).
After upgrade I tested adding teams and enabling apps. Everything worked as expected.
Gitea was not restarted for those changes. Also checked that the gitea-credentials secret is different from the root-credentials secret.

PR #1942, opened 2025-02-18
fix: workload with validatingwebhookcfg

srodenhuis:
Fix to solve argo sync issues when using team-admin to create a workload using a chart with a validatingwebhookconfig resource.


merll (approved):

Ani1357 (approved):
Tested by installing kserver on existing cluster. After the upgrade to this branch the argocd app was synced successfully.

PR #1943, opened 2025-02-18
ci: update workflow to support k8s 1.32 and wait for kubeconfig

merll:
This PR updates the list of versions to the ones currently shown as supported by the Linode API, including 1.32 and removing 1.29. Also a change was made to wait until the Kubeconfig can be retrieved, as relying on all nodes being up and running turned out not to be sufficient.

### DIFF
--- a/.github/workflows/integration.yml
+++ b/.github/workflows/integration.yml
@@ -223,16 +223,15 @@ jobs:
             
             sleep 30
           done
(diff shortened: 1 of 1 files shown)


ElderMatt (approved):
Tested on cluster, it now waits for the kubeconfig. Great job!

Ani1357 (approved):

PR #1944, opened 2025-02-19
fix: removed git error which can expose credentials

Ani1357:
This fix addresses a possible exposure of the otomi-admin credentials in case the `commitAndPush ` fails as showed here:
image


merll (changes requested):
We cannot simply pass over this error, because then there will be no retry. But we can of course replace this with `throw <some message here>`.

merll (approved):

j-zimnowoda (approved):

PR #1947, opened 2025-02-20
chore(chart-deps): update external-dns to version 6.38.0

svcAPLBot:
$'This PR updates the dependency **external-dns** to version **6.38.0**.'


merll:
Need to check on chart version, this does not seem like an upgrade to me.

PR #1949, opened 2025-02-20
chore(chart-deps): update gitea to version 5.0.9

svcAPLBot:
$'This PR updates the dependency **gitea** to version **5.0.9**.'


merll:
This causes Gitea to restart, but all services seem to be operating normally. We have pinned the app version in the values; therefore will not cause an upgrade of Gitea itself.

PR #1963, opened 2025-02-20
ci: update chart index and improve checks

merll:
This PR includes the following changes to the chart update script:
* It checks if the index version corresponds with the chart's version. If not, the update is skipped. When running locally, `CI_GIT_LOCAL_BRANCH_ONLY` can be set to `true` for preparing a fix. Resulting changes should be submitted to a separate PR.
* Any errors leading to skipping a chart are summarized at the end, as they tend to disappear in the log output.
* Some improvements to running the script locally. For example, the checkout hooks are skipped.

Also, two inaccuracies in the chart index were found and fixed, including the version mismatches above and a missing alias for the `tekton-pipeline` chart.

### DIFF
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -15,7 +15,7 @@ dependencies:
     version: 0.23.0
     repository: https://cloudnative-pg.github.io/charts
   - name: external-dns
-    version: 6.20.4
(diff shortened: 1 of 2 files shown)


Ani1357 (approved):
Tested locally, works as expected.

PR #2008, opened 2025-03-21
feat: updated keycloak operator and gitea values

Ani1357:
This PR:
- Decreases the gitea session lifetime from 1h to 10m
- Sets oaut2-proxy cookie expiration to 3min

To be tested in conjunction with this apl-tasks branch: https://github.com/linode/apl-tasks/tree/APL-478-1


CasLubbers:
Tested on a cluster two things. 
1. When you sign out a user in Keycloak manually the user has te re-login ✅ 
2. When you delete a user in APL it's session is remove after some time. (When the applying is done) ✅

CasLubbers (approved):

ElderMatt (approved):

PR #2025, opened 2025-03-31
fix: 5xx errors

CasLubbers:
- Add option for request to prepare values to add files that only needs to be encrypted or decrypted
- Fixed issue where we CD into directories which resulted in race conditions in other processes

### DIFF
--- a/src/cmd/validate-values.ts
+++ b/src/cmd/validate-values.ts
@@ -19,7 +19,7 @@ export const validateValues = async (envDir = env.ENV_DIR): Promise<void> => {
   // TODO: Make this return true or error tree
   // Create an end point function (when running otomi validate-values) to print current messages.
   const argv: HelmArguments = getParsedArgs()
-  d.log('Values validation STARTED')
+  d.log('Values validation STARTED on %s', envDir)
 
   if (argv.l || argv.label) {
     const labelOpts = [...new Set([...(argv.l ?? []), ...(argv.label ?? [])])]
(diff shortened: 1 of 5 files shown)


src/server.ts
```diff
@@ -40,21 +40,21 @@ app.get('/init', async (req: Request, res: Response) => {
...
     await bootstrapSops(envDir)
     await setValuesFile(envDir)
     // Encrypt ensures that a brand new secret file is encrypted in place
-    await encrypt(envDir)
+    await encrypt(envDir, ...(files ?? []))
     // Decrypt ensures that a brand new encrypted secret file is decrypted to the .dec file
-    await decrypt(envDir)
-    await validateValues(envDir)
+    await decrypt(envDir, ...(files ?? []))
+    // await validateValues(envDir)
```
merll, line 52: Can we just remove this? We have a history in case it needs to be restored later.

src/cmd/validate-values.ts
```diff
@@ -19,7 +19,7 @@ export const validateValues = async (envDir = env.ENV_DIR): Promise<void> => {
   // TODO: Make this return true or error tree
   // Create an end point function (when running otomi validate-values) to print current messages.
   const argv: HelmArguments = getParsedArgs()
-  d.log('Values validation STARTED')
+  d.log('Values validation STARTED on %s', envDir)
```
merll, line 22:
```suggestion
  d.log('Values validation STARTED on', envDir)
```

src/server.ts
CasLubbers: Thanks for the reminder I think it's a leftover. Will remove it.

merll (approved):

ferruhcihan (approved):

PR #2074, opened 2025-04-17
fix: correct values order

dennisvankekem:
## 🧹 Checklist

### DIFF
--- a/values-changes.yaml
+++ b/values-changes.yaml
@@ -339,7 +339,6 @@ changes:
   - version: 33
     migrateLegacyValues: true
   - version: 34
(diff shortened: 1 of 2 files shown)


values-changes.yaml
```diff
@@ -339,17 +339,15 @@ changes:
   - version: 33
     migrateLegacyValues: true
   - version: 34
+    teamSettingsMigration: true
```
ElderMatt: I dont think this will work. Cluster who are on 35 or higher won't see this change

CasLubbers (approved):

PR #2078, opened 2025-04-17
feat: upgrade Keycloak to recent release and deprovision operator

merll:
## 📌 Summary

This PR upgrades Keycloak to a more recent and maintained release. Since the previous Operator pattern did not provide a good way of maintaining the lifecycle of this product in our platform, a change is made to a regular Helm install like the majority of our integrated projects. The Bitnami chart was chosen since it provided good practices of pre-configuration, relatively close to our current setup.

For the upgrade, it was necessary to upgrade the platform logout URL redirect. Since this was provided as a 301 HTTP response code from `/logout-otomi`, the latter had to be changed since otherwise browsers cache this response indefinitely and therefore may never follow the updated redirect. The new redirect under `/platform-logout` was changed to a 302 HTTP response code to make changes easier in the future (e.g. should we rename the realm or client for rebranding purposes). As a consequence, Console and API also need to be updated accordingly.

Console: https://github.com/linode/apl-console/pull/545
API: https://github.com/linode/apl-api/pull/701

## 🔍 Reviewer Notes

This to test in particular should be:
* Login and logout of the platform through Console
* OAuth-Proxy usage (e.g. ArgoCD)
* API authentication
* Creating and removing users and teams

For the logout and closing of TTY sessions to work, the matching feature branches of APL-691 will need to be used for API and Console. Also, cleanups were made in Console regarding the login theme, which was causing 404 errors due to missing (but not needed) CSS files. In order to test the updated theme, the values (multi-line string) of the ArgoCD application `keycloak-keycloak` must be manually patched in `initContainers[0].image` to `docker.io/linode/apl-console:APL-691`. When referring to `main`, it will still work just like before, just not benefit from improvements there. These manual changes will be overwritten on any Tekton pipeline run.

## 🧹 Checklist


CasLubbers (approved):

Ani1357 (approved):
Tested upgrade from latest release in conjunction with the update versions of console and api(through raw_values). 
Upgrade script was executed as expected and keycloak operator resources were removed.
Login/Logout works for platform-admin and team users.  Also login to gitea argocd harbor works through oidc.

PR #2108, opened 2025-04-29
fix: updated keycloak init container version rendering

Ani1357:
## 📌 Summary

With this PR we can be more consistent in setting versions for our tools. Now consoleLogin accepts a version tag without the `v` prefix as the other components.

## 🧹 Checklist


CasLubbers (approved):

j-zimnowoda (changes requested):
It works for `consoleLogin: 1.1.1-rc.1`
but
It does not work for `consoleLogin: 1.1.1`

Although it is not entirely correct, I see that we have successfully used the following statement:
```
{{- $isSemver := regexMatch "^[0-9.]+" $version }}
```

j-zimnowoda (approved):

PR #2151, opened 2025-05-13
feat: add apl-operator

CasLubbers:
## 📌 Summary
- Set otomi-pipelines to false so it does not get installed
- Added new apl-operator which takes over the applying from tekton
- apl-operator polls for changes every second. 
  - The changes are either applied for only the teams, when there are only files changed in teams directory
  - Or the whole platform gets applied
  - When all the new commits contain [ci skip] the apply is skipped
- apl-operator reconciles every 5 minutes. It will reapply the whole platform
- The build state is stored in a configmap

## 🧹 Checklist


helmfile.d/helmfile-09.init.yaml
```diff
@@ -45,9 +45,9 @@ releases:
...
       app: core
     <<: *default
-  - name: otomi-pipelines
-    installed: true
-    namespace: otomi-pipelines
-    labels:
-      app: core
-    <<: *default
+#  - name: otomi-pipelines
+#    installed: true
```
j-zimnowoda, line 49: I would actually set it to false, so we ensure that corresponding app is removed

src/common/envalid.ts
```diff
@@ -32,6 +32,8 @@ export const cliEnvSpec = {
   RANDOM: bool({ desc: 'Randomizes the timeouts by multiplying with a factor between 1 to 2', default: false }),
   MIN_TIMEOUT: num({ desc: 'The number of milliseconds before starting the first retry', default: 60000 }),
   FACTOR: num({ desc: 'The factor to multiply the timeout with', default: 1 }),
+  GITEA_URL: str({ default: 'gitea-http.gitea.svc.cluster.local:3000' }),
```
j-zimnowoda, line 35: GITEA_URL could be split into:
GITEA_HOST and GITEA_PORT

j-zimnowoda: Perhaps also instead of GITEA we could use GIT because the operator is git repo provider agnostic

src/operator/apl-operator.ts
```diff
@@ -0,0 +1,256 @@
...
+import { module as validateValuesModule } from '../cmd/validate-values'
+import { setValuesFile } from '../common/repo'
+import { waitTillGitRepoAvailable } from '../common/k8s'
+import { env } from '../common/envalid'
+import path from 'path'
+
+export class AplOperator {
+  private d = terminal('operator:apl')
+  private isRunning = false
+  private pollInterval = 1000
```
j-zimnowoda, line 18: It could be configurable

CasLubbers:
It's working only need to add more unit-tests

src/operator/git-repository.ts
```diff
@@ -0,0 +1,155 @@
...
+    const logResult = await this.git.log({
+      from: fromRevision,
+      to: toRevision,
+    })
+
+    return logResult.all.every((commit) => commit.message.includes(this.skipMarker))
+  }
+
+  async pull(): Promise<{ hasChangesToApply: boolean; applyTeamsOnly: boolean }> {
+    try {
```
ElderMatt, line 100: Or the function name should change or the code should be split up more. At the moment it is not only pulling, it is also checking for changes to apply and if it is only necessary for teams.

src/operator/apl-operator.ts
```diff
@@ -0,0 +1,179 @@
...
+        this.d.info('Reconciliation completed')
+      } catch (error) {
+        this.d.error('Error during reconciliation:', error)
+      }
+
+      await new Promise((resolve) => setTimeout(resolve, this.reconcileInterval))
+    }
+
+    this.d.info('Reconciliation loop stopped')
+  }
```
ElderMatt, line 108: At the moment it is not only polling for changes, but also trying to apply them.
Perhaps have an overarching function called pollingAndApplying which calls the polling function and if a change is found apply it.

src/operator/apl-operator.ts
```diff
@@ -0,0 +1,179 @@
...
+
+  // Only used in tests: run N iterations and exit
+  public async pollForChanges(maxIterations = Infinity): Promise<void> {
```
ElderMatt, line 111: At the moment it is not only polling for changes, but also trying to apply them.
Perhaps have an overarching function called pollingAndApplying which calls the polling function and if a change is found apply it.

src/operator/apl-operator.ts
```diff
@@ -0,0 +1,179 @@
...
+    this.d.info('Starting polling loop')
+
+    for (let i = 0; this.isRunning && i < maxIterations; i++) {
+      if (this.isApplying) {
+        this.d.debug('Skipping polling, apply process is in progress')
+        await new Promise((resolve) => setTimeout(resolve, this.pollInterval))
```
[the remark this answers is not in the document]
CasLubbers: Good one, makes it read it more easily

src/common/bootstrap.ts
```diff
@@ -72,8 +72,10 @@ export const bootstrapGit = async (inValues?: Record<string, any>): Promise<void
   }
 
   if (isCli) {
-    await copyFile(`${rootDir}/bin/hooks/pre-commit`, `${env.ENV_DIR}/.git/hooks/pre-commit`)
+    d.info('Copying pre-commit hook')
+    // await copyFile(`${rootDir}/bin/hooks/pre-commit`, `${env.ENV_DIR}/.git/hooks/pre-commit`)
```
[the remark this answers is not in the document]
CasLubbers: Ah yes and I also need to remove the bin/hooks/pre-commit that file was for drone

src/cmd/apply.ts
```diff
@@ -124,7 +122,8 @@ const applyAll = async () => {
         { streams: { stdout: d.stream.log, stderr: d.stream.error } },
       )
       await cloneOtomiChartsInGitea()
-      await retryCheckingForPipelineRun()
+      // Change this to check if apl-operator successfully deployed
+      // retryCheckingForPipelineRun()
```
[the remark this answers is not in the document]
CasLubbers: Thanks

src/operator/git-repository.ts
CasLubbers: Have try to make the function split up more and also renamed it to syncAndAnalyzeChanges. And then based on what we return I think it should be clear enough what it does?

ElderMatt (approved):

ferruhcihan (approved):
Tested with a cluster both deploying and upgrading scenarios, LGTM 👍🏻

merll (changes requested):
According to my tests, the operator needs more permissions to function properly.

values-changes.yaml
```diff
@@ -356,3 +356,11 @@ changes:
   - version: 39
     deletions:
       - 'teamConfig.{team}.services[].type'
+  - version: 40
+    addAplOperator: true
```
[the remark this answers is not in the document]
CasLubbers: Yes it could be done in an upgrade script, but what I thought was that we rather have all the migrations in the values-schema with a check in there. That way I can also easily re-use the hf function that already exists.

values-changes.yaml
```diff
@@ -356,3 +356,11 @@ changes:
...
+    additions:
+      - apps.apl-operator.resources.operator.requests.cpu: 50m
+      - apps.apl-operator.resources.operator.requests.memory: 128Mi
+      - apps.apl-operator.resources.operator.limits.cpu: '1'
+      - apps.apl-operator.resources.operator.limits.memory: 1Gi
+      - apps.apl-operator._rawValues: {}
```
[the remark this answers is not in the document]
CasLubbers: I need to retest it without it again, but I thought I added this because I didn't saw the app in the Gitea values.

CasLubbers:
> I am seeing lots of errors in combination with the Helm file upgrade, but also the pre-upgrade from the latest release. Apparently the serviceAccount does not have enough permissions. This one for example comes from Helmfile with labels stage=prep:
> 
> ```
>   Error: unable to generate manifests: could not get information about the resource: poddisruptionbudgets.policy "ingress-nginx-platform-controller" is forbidden: User "system:serviceaccount:apl-operator:apl-operator" cannot get resource "poddisruptionbudgets" in API group "policy" in the namespace "ingress"
>   Error: plugin "diff" exited with error
> ```
> 
> But the operator also needs to be allowed to do anything that the upgrade scripts need to do, e.g. managing labels and annotations, creating Jobs etc. Maybe it is too ambitious to be so granular about it. It can break again on any upgrade.

That was already something I was afraid of. It's a bit hard to guess what were gonna need in the future for RBAC. Maybe read access to everything is oke and then for updating, deleting and creating we can be more strict.
Other option is to give it the admin role...

charts/apl-operator/values.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+tolerations: []
+
+affinity: {}
+
+env: {}
+
+kms: {}
+
+giteaPassword: ""
```
j-zimnowoda, line 63: gitPassword

charts/apl-operator/templates/secret.yaml
```diff
@@ -0,0 +1,47 @@
...
+
+---
+apiVersion: v1
+kind: Secret
+metadata:
+  name: gitea-credentials
+  namespace: {{ .Release.Namespace }}
+type: Opaque
+stringData:
+  GIT_USERNAME: otomi-admin
```
j-zimnowoda, line 46: GIT_USERNAME values should come via values

charts/apl-operator/values.yaml
```diff
@@ -0,0 +1,63 @@
...
+
+nodeSelector: {}
+
+tolerations: []
+
+affinity: {}
+
+env: {}
+
+kms: {}
```
j-zimnowoda, line 61: Let's add and example here

```
kms:
  sops:
    age:
      publicKey: "age1examplepublickeyxxxxxxxxxxxxxxxxxxxxxxxxxxxx"
```

values-changes.yaml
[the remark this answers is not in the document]
merll: It is only added after the pipeline is done, but it is already usable by any templates.

values-changes.yaml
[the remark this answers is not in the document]
merll: I had a similar consideration in the Istio upgrade. Generally I think we might want to have clean separation between values repo changes (requiring new / updated values) and one-off scripts required by a release upgrade. But I would not consider this a merge blocker.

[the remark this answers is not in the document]
CasLubbers: Sure I'll post it in our slack to see what the team thinks.

values-changes.yaml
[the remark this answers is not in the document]
CasLubbers: It worked

helmfile.d/snippets/defaults.yaml
```diff
@@ -906,6 +906,16 @@ environments:
                   cpu: "1"
```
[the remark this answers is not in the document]
CasLubbers: Done

merll:
There's another one:
```
Error: unable to generate manifests: cannot patch "istiod-default-validator" with kind ValidatingWebhookConfiguration: validatingwebhookconfigurations.admissionregistration.k8s.io "istiod-default-validator" is forbidden: User "system:serviceaccount:apl-operator:apl-operator" cannot patch resource "validatingwebhookconfigurations" in API group "admissionregistration.k8s.io" at the cluster scope
```

merll (changes requested):
The Istiod Helmfile install still fails, due to missing permissions. Otherwise looks good to me.

merll (approved):

PR #2176, opened 2025-05-26
chore: Upgrade Kiali Operator

merll:
## 📌 Summary

This PR implements a major upgrade of the Kiali Operator, which is required for compatibility with the newer Istio version v1.26.x. It also includes some minor configuration updates and fixes. 

## 🔍 Reviewer Notes

Once the Kiali app is enabled, it should be possible to review the Istio mesh configuration. Tracing configuration had not been implemented completely yet before and should be picked up later.

The session timeout is a known issue and not new with this release: https://github.com/kiali/kiali/issues/5233

## 🧹 Checklist

### DIFF
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -50,7 +50,7 @@ dependencies:
     version: 24.5.7
     repository: https://charts.bitnami.com/bitnami
   - name: kiali-operator
-    version: 1.86.1
+    version: 2.10.0
     repository: https://kiali.org/helm-charts
   - name: knative-operator
     version: 0.1.0
--- a/charts/kiali-operator/Chart.yaml
+++ b/charts/kiali-operator/Chart.yaml
@@ -1,19 +1,20 @@
 apiVersion: v2
-name: kiali-operator
-description: Kiali is an open source project for service mesh observability, refer to https://www.kiali.io for details.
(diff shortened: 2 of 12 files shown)


j-zimnowoda (approved):
After applying changes to Kiali config I can see mesh and istio config
image

image

j-zimnowoda (commented):
Getting some errors:
image

merll:
> Getting some errors: image

Did you enable Kiali before or after upgrading to the feature branch? Because I cannot reproduce it when I upgrade before activating the app. I also see that the service account does have this permission in my cluster.

j-zimnowoda:
> > Getting some errors: image
> 
> Did you enable Kiali before or after upgrading to the feature branch? Because I cannot reproduce it when I upgrade before activating the app. I also see that the service account does have this permission in my cluster.

I enabled Kiali after the cluster upgrade.

merll:
> > > Getting some errors: image
> > 
> > 
> > Did you enable Kiali before or after upgrading to the feature branch? Because I cannot reproduce it when I upgrade before activating the app. I also see that the service account does have this permission in my cluster.
> 
> I enabled Kiali after the cluster upgrade.

Not sure, but I am guessing this is a race condition that causes the error during the installation. Then the message is persisted. I was not able to reproduce this in two attempts.

PR #2178, opened 2025-05-26
ci: restart dev apl-operator after main branch update

merll:
## 📌 Summary

Since we replaced the Tekton pipeline with the APL operator, we no longer need to push fake-commits for triggering an upgrade. However, since the branch never changes, the APL operator needs to become aware of the upstream changes. This PR changes the script to restart the pods.

## 🧹 Checklist


CasLubbers (approved):
Nice, this indeed makes things so much easier

PR #2190, opened 2025-05-30
chore(chart-deps): update harbor to version 1.17.1

svcAPLBot:
$'This PR updates the dependency **harbor** to version **1.17.1**.'

[APL Tasks PR](https://github.com/linode/apl-tasks/pull/238)


values/harbor/harbor.gotmpl
```diff
@@ -7,7 +7,7 @@
 {{- $harborSecretName := ($harborDomain | replace "." "-") }}
 {{- $notarySecretName := ($notaryDomain | replace "." "-") }}
 {{- $externalUrl := printf "https://%s" $harborDomain }}
-{{- $tag := $h | get "image.tag" "v2.6.4" }}
+{{- $tag := $h | get "image.tag" "v2.13.1" }}
```
CasLubbers: I see that the tag is hardcoded in the chart. So we could maybe remove this one and all the values below.

CasLubbers:
Tested it on cluster and works ✅

charts/harbor/Chart.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: v1
-appVersion: 2.12.2
+appVersion: 2.13.1
```
merll, line 2: Please also update this in apps.yaml. I will enhance the CI script shortly, because we keep missing that one.

values/harbor/harbor.gotmpl
merll: We still have the image override in most cases, but we can make it conditional. We should indeed not place a default here, unless we want this to be different from the Helm chart.

merll:
It would be a good occasion to clean up the values.yaml for any defaults which correspond with the chart or have become obsolete. For example, I do longer see the `postgresql:` section in the chart.

ferruhcihan (approved):
Tested with a cluster, LGTM (except for the unresolved comment).

PR #2203, opened 2025-06-03
feat: APL-672 adding ORCS support

Ani1357:
## 📌 Summary

- This PR adds support for [ORCS](https://bits.linode.com/CNS/orcs) (OCI Registry Cache Service) thus enabling APL to not rely on the availability of public registries like docker.io, gcr.io, ghrc.io, quay.io and k8s.io.
- This feature relies on an additional boolean called `useORCS` added under the otomi configuration. When set to true, through the values go-templates it will configure each chart to use the cached registries instead of the public ones. 
- The integration github action was also updated by adding a new boolean which allows such option to be enabled.

## 🔍 Reviewer Notes

Offline rendering should allow to check the images being used when useORCS is set to true/false/unset.
I updated also the chart/apl to set the registry bases on the useORCS value provided. For the time being this can be used also by the community(not only LKE installations)since ORCS is publicly available.
For the time being I was not able to configure the images used by the knative-operator(under `gcr.io/knative-releases/knative.dev/serving/`)
## 🧹 Checklist


values-changes.yaml
```diff
@@ -367,3 +367,6 @@ changes:
   - version: 41
     customFunctions:
       - installIstioHelmCharts
+  - version: 42
+    additions:
+      - otomi.useORCS: false
```
merll: I don't really see a good reason to use `additions` any longer. Adding this to defaults is enough in almost all use cases.

.github/workflows/integration.yml
```diff
@@ -279,6 +283,13 @@ jobs:
           imagePullSecretNames:
             - reg-otomi-github
           EOF
+
+          # If ORCS is enabled, pull image from ORCS cache registry
+          if [[ '${{ inputs.useORCS }}' == 'true' ]]; then
+          cat << EOF > values-container-registry.yaml
```
merll, line 289: Where is this file being read?

merll: Just saw that this block overwrites the file written above. Maybe what is above should go into an "else" block below, so that its more obvious that these two effects are mutually exclusive.

charts/knative-operator/templates/deployments.yaml
```diff
@@ -38,7 +38,7 @@ spec:
         - name: operator-webhook
           # This is the Go import path for the binary that is containerized
           # and substituted here.
-          image: gcr.io/knative-releases/knative.dev/operator/cmd/webhook@sha256:9d0c7b27f2451d69ad0c48752795453bd73d36fbd918583432f70cfc214844fa
+          image: {{ .Values.webhook.image.repo }}:{{ .Values.webhook.image.tag }}
```
merll: This file is in conflict with #2181, since we use the upstram chart there. But I suggest we push the change through in this PR, and then move the image override to "values" in the other PR during conflict resolution.

values-changes.yaml
Ani1357: I agree. Removed.

merll (approved):

chart/apl/templates/job.yaml
```diff
@@ -30,7 +30,11 @@ spec:
           type: RuntimeDefault
       containers:
         - name: install
-          image: {{ $imageName }}:{{ $version }}
+          {{- if .Values.otomi.useORCS }}
+          image: "mirror.registry.linodelke.net/docker/{{ $imageName }}:{{ $version }}"
```
CasLubbers, line 34: Can we also use here use the variable specified in the defaults?

helmfile.d/snippets/defaults.yaml
```diff
@@ -1314,6 +1314,7 @@ environments:
           isMultitenant: true
           nodeSelector: {}
           isPreInstalled: false
+          useORCS: false
```
CasLubbers, line 1320: If we set this default to false and we also would want this for when you install it via cloudmanager we need to update the apl-installer. We can already add the flag there.

values-schema.yaml
```diff
@@ -3115,6 +3115,10 @@ properties:
         default: latest
         description: 'Best pin this to a valid release version found in the repo.'
         pattern: '(v[0-9]+.[0-9]+.[0-9]+|[a-zA-Z]+[a-zA-Z0-9-])'
+      useORCS:
+        type: boolean
+        description: Defines if the OCI Registry Cache Service (ORCS) is used to cache images from the public registry.
+        default: false
```
CasLubbers: This default is overhead as we already have it in the defaults.yaml

CasLubbers (approved):
3 Small comments, not blocking. Upgrade path works.

helmfile.d/snippets/defaults.yaml
Ani1357: Yes, correct. We want to be able to manage this via the apl-installer

chart/apl/templates/job.yaml
Ani1357: No because this is needed to create the helm chart. We dont rely on the defaults here but must be able to set the flag so that the installer image gets pulled form ORCS.

values-schema.yaml
Ani1357: Removed duplicate.

PR #2214, opened 2025-06-06
fix: APL-851 knative service url rendering

Ani1357:
## 📌 Summary

This PR fixes the rendering for knative workload service url rendering.  To be tested with apl-api: [APL-849](https://github.com/linode/apl-api/tree/APL-849)

## 🧹 Checklist

### DIFF
--- a/charts/team-ns/templates/_helpers.tpl
+++ b/charts/team-ns/templates/_helpers.tpl
@@ -41,10 +41,9 @@ otomi.io/team: {{ .Values.teamId }}
 
 {{- define "service.domain" -}}
 {{- $v := .dot.Values }}
-{{- $teamId := (and .isKnativeService (printf "team-%s" $v.teamId)) | default $v.teamId }}
-{{- $svc := (hasKey .s "hasPrefix" | ternary (printf "%s-%s" $teamId (.s.svc | default .s.name)) (.s.svc | default .s.name)) -}}
+{{- $teamSuffix := (and .isKnativeService (printf "team-%s" $v.teamId)) | default $v.teamId }}
 {{- $shared := (and .s.isCore (eq $v.teamId "admin") (hasKey .s "isShared")) | default false -}}
-{{- $host := ($shared | ternary .s.name (printf "%s-%s" .s.name $teamId )) -}}
+{{- $host := ($shared | ternary .s.name (printf "%s-%s" .s.name $teamSuffix )) -}}
 {{- $domain := (index .s "domain" | default (printf "%s.%s" $host $v.domain)) -}}
 {{- print $domain -}}
 {{- end -}}


charts/team-ns/templates/_helpers.tpl
```diff
@@ -41,9 +41,10 @@ otomi.io/team: {{ .Values.teamId }}
 
 {{- define "service.domain" -}}
 {{- $v := .dot.Values }}
-{{- $svc := (hasKey .s "hasPrefix" | ternary (printf "%s-%s" $v.teamId (.s.svc | default .s.name)) (.s.svc | default .s.name)) -}}
+{{- $teamId := (and .isKnativeService (printf "team-%s" $v.teamId)) | default $v.teamId }}
+{{- $svc := (hasKey .s "hasPrefix" | ternary (printf "%s-%s" $teamId (.s.svc | default .s.name)) (.s.svc | default .s.name)) -}}
 {{- $shared := (and .s.isCore (eq $v.teamId "admin") (hasKey .s "isShared")) | default false -}}
-{{- $host := ($shared | ternary .s.name (printf "%s-%s" .s.name $v.teamId )) -}}
+{{- $host := ($shared | ternary .s.name (printf "%s-%s" .s.name $teamId )) -}}
```
merll: This could be cleaned up a lot to become more readable.

- `$teamId` is not a good variable name, since it is different from the values teamId. I would suggest something like teamSuffix or so.
- I do not see that `$svc` is actually used. Besides that, there is at least one repeated nested expression that should go into a  separate variable.
- Working with logical expressions and `default` may work and is common in most languages, but in templates makes things harder to understand. To my understanding, if `$shared` becomes `true`, then `$host` would also read `true`, and could end up being rendered into `$domain`. Maybe this never happens because `domain` is set in this case, but it looks to me like taking chances that this may break suddenly due to wrong input or a minor change.
- I would suggest not to mix logical evaluations and `ternary`.

Ani1357: Great input, thank you. It was indeed difficult also for me to read this function. I applied some of your suggestions:
- renamed `$temaId` to `$teamSuffix`
- removed the `$svc` variable

Regarding the other 2 suggestions:
- `$shared` is set to **true** if `.s.isCore` is **true**, `$v.teamSuffix` equals "admin", and the key "isShared" exists in `.s`. Otherwise, it defaults to **false**.
    - based on that host is either `.s.name` or `.s.name-$teamSuffix`
- we are using ternary the same as showed in the [helm documentation](https://helm.sh/docs/chart_template_guide/function_list/#ternary)

Please suggest other code changes you would like to see here otherwise we can merge this now and do more refactoring later on.

merll: Lets postpone other changes to later on.

PR #2288, opened 2025-06-26
fix: upgrade oauth2 proxy to 7.12.18 and fix session interruption

merll:
## 📌 Summary

This PR upgrades the OAuth2-Proxy Helm chart to version 7.12.18. It also includes some configuration changes:
* All command-line arguments were moved to either Helm-chart provided values, or a configuration file. Commented options were removed.
* The configuration was revised against recent documentation, and therefore some options were changed:
  - The `keycloak-oidc` provider is used, which will require a change in the Keycloak client configuration.
  - `cookie_expiration` was raised from `3m` to `30m`, which is corresponding with the (default) Keycloak setting. This seems to fix the timeouts in Console and other apps such as ArgoCD, as it does not require re-authentication within that time span.
  - The error page has been updated to use a non-script redirect. The same template is also mapped to the `sign_in` page.
  - `show_debug_on_error` is removed, as it is not recommended for production.

## 🔍 Reviewer Notes

The following should be verified, besides essential SSO functionality:
* Posting a request a few minutes after idle time, in different platform screens and SSO apps such as ArgoCD.
* Deleting a user that is signed-in in a private browser tab. That user should be redirected to the login page 1 minute (+ apply time) after the deletion, when posting a request. The access code lifespan is 1 minute, and OAuth2-Proxy should fail to renew against Keycloak.

Keycloak needs to pass an additional token attribute. Therefore, the [following feature branch](https://github.com/linode/apl-tasks/pull/253) must also be used by overriding the apps for apl-keycloak-operator:
```yaml
  _rawValues:
    image:
      tag: APL-894
      pullPolicy: Always
```

## 🧹 Checklist


ferruhcihan (approved):
Tested, LGTM 👍🏻

merll:
> I tired to deploy it with ORCS enabled and obtained error
> 
> [...]
>
> Please set
> 
> ```
> global: 
>   security: 
>     allowInsecureImages: true
> ```
> 
> in the values gotmpl. Read more at [bitnami/charts#30850](https://github.com/bitnami/charts/issues/30850)

This is done now.

j-zimnowoda (approved):

PR #2299, opened 2025-07-01
fix: migrate CloudnativePG backup to plugin for improved resource control

merll:
## 📌 Summary

This PR migrates the backup configuration of platform PostgreSQL clusters to the [new plugin-based configuration](https://cloudnative-pg.io/plugin-barman-cloud/docs/migration/), while the old one is still supported but will be deprecated.

Migrating has the significant advantage, that the WAL archiver and backup no longer runs inside the PostgreSQL main container, but instead in a sidecar. Therefore it allows for setting resource constraints separately. This solves the issue that memory spikes occurred inside the main container during the backup, sometimes leading to an OOM kill.

## 🔍 Reviewer Notes

The upgrade should be tested with and without active backups to the platform databases. A manual backup can be taken (creating a Backup resource), or the ScheduledBackup resource can be deleted to have ArgoCD recreate it and perform an immediate backup.

## Known issues

* The immediate backup after activating the backups invariably fails. The next scheduled occurrence works. This is because the ScheduledBackup triggers before the cluster with updated plugin configuration is ready and the cluster is healthy. The plugin can however not be loaded without being fully configured. Before we go into implementing our own workaround, we should await the following release of ArgoCD (3.10, currently in RC stage). There additional checks on the health of the Cluster resource are added, which together with the sync-wave added to the ScheduledBackup, might fix the issue.
* The Helm chart for the CloudnativePG Barman Plugin is not yet released. It was copied from an open PR of [CloudnativePG](https://github.com/cloudnative-pg/charts), which has mostly addressed review comments and should be merged soon.

## 🧹 Checklist


helmfile.d/snippets/defaults.yaml
```diff
@@ -127,6 +127,13 @@ environments:
               limits:
                 cpu: "1"
                 memory: 1Gi
+            barmanPluginResources:
```
CasLubbers, line 130: This needs to be reflected in the values-schema

CasLubbers (approved):
Migrated to this version on cluster.
Created manual backups for all three databases ✅ 
Created scheduled backups for all three databases ✅ 
Only remark that I have is it took some time for me to figure out that you need to add the method and pluginConfiguration to your manual backup. Otherwise you'll get the error: `cannot proceed with the backup as the cluster has no backup section`
Maybe we can add a quick note to the docs on how to create manual backup.

merll:
We should certainly have more detailed docs on how to schedule a manual backup by creating a `Backup` resource. That can however not be integrated into the current structure very well. The Basebackup+WAL and SQL-based ("emergency") backup should then go into a section on their own. I would suggest handling this in a followup issue.

Upgrade notes should include a notice that which services are expected to restart and become unavailable for a few seconds.

helmfile.d/helmfile-01.init.yaml.gotmpl
```diff
@@ -31,4 +31,5 @@ releases:
     namespace: cert-manager
     labels:
       pkg: cert-manager
+      app: core
```
ElderMatt, line 34: Should this always be the case or only when installed through Linode?

merll: This should always be like this, because the CloudnativePG is also not dependent on the provider.

Ani1357:
Tested by :
- Enabling backups on current release for gitea and harbor. Backups completed(on second attempt)
- Upgraded cluster to the feature branch
- Rerun backups(trigered manually using kubectl-cnpg): Failed
  image
- Enabled harbor and backups but again failed with the same error

More observations:
- the cnpg-system-cloudnative-pg-plugin-barman-cloud argo app is stuck in out-of-sync because of a change in the certificate resources
image
image

merll:
Thanks for the detailed test.

>     * Enabling backups on current release for gitea and harbor. Backups completed(on second attempt)

Probably this is a similar issue, that the Cluster is not ready when the Backup is triggered by the `immediate` flag.

>     * Rerun backups(trigered manually using kubectl-cnpg): Failed
>     * Enabled harbor and backups but again failed with the same error

The `backup` section in the Cluster spec is [supposed to be removed](https://cloudnative-pg.io/plugin-barman-cloud/docs/migration/#step-2-update-the-cluster-for-plugin-wal-archiving). The ObjectStore now references the Cluster, and the plugin makes that connection when a Backup is created. When using the kubectl plugin or creating the resource manually it needs to know that the plugin is used. So running

```shell
kubectl cnpg backup -n gitea gitea-db -m plugin --plugin-name barman-cloud.cloudnative-pg.io
```

got me a working backup.

>     * the cnpg-system-cloudnative-pg-plugin-barman-cloud argo app is stuck in out-of-sync because of a change in the certificate resources

I was able to reproduce that, but I am still not sure why this happens. After any additional change on the ArgoCD configuration, this disappears. I added an exclusion in this PR already, but ArgoCD seems to take a while to adopt the change.

PR #2303, opened 2025-07-02
feat: nickname as username for gitea

ElderMatt:
## 📌 Summary

This PR replaces the username from email to nickname.

The nickname is a transformed email which will help make the users unique and help with the 500 error when users have the same username.


versions.yaml
```diff
@@ -1,5 +1,5 @@
 api: main
 console: main
 consoleLogin: main
-tasks: main
+tasks: APL-895
```
ferruhcihan: Just a reminder to revert this.

ferruhcihan (approved):

PR #2337, opened 2025-07-14
feat: enabling ORCS by default

Ani1357:
This PR enabled the use of ORCS by default after tekton images were moved to new (not working at the moment) registry (ghcr.io)


ElderMatt (approved):

j-zimnowoda (dismissed):
Holding on due to alternative solution: https://github.com/linode/apl-core/pull/2339

PR #2420, opened 2025-08-05
feat: adding ORCS support for kubeflow

Ani1357:
## 📌 Summary

[APL-1012](https://jira.linode.com/browse/APL-1012)

Related PR: https://bits.linode.com/CNS/orcs/pull/92

## 🧹 Checklist


charts/kubeflow-pipelines/values.yaml
```diff
@@ -1,3 +1,8 @@
+global:
+  kubeflowRegistry:
```
CasLubbers, line 6: It would be a bit cleaner to have these in comments. Now they are undefined

charts/kubeflow-pipelines/templates/cache-deployer/deployment.yaml
```diff
@@ -1,3 +1,6 @@
+{{- $v := .Values }}
+{{- $kubeflowRegistry := default "ghcr.io/kubeflow" $v.global.kubeflowRegistry }}
```
CasLubbers: We can make this redundant by defining it in the _helpers.tpl e.g. for kubeflow registry we can do this:
```
{{- define "kfp.registry" -}}
{{- default "ghcr.io/kubeflow" .Values.global.kubeflowRegistry }}
{{- end }}
```
and in the deployment you can do this:
```
        image: {{ include "kfp.registry" . }}/kfp-cache-deployer:{{ .Chart.AppVersion }}
```
you can also do that for all the other registries. So you only have to specify it once.

CasLubbers (approved):

PR #2426, opened 2025-08-06
fix: smoke tests should fail on template rendering failure

ElderMatt:
## 📌 Summary

This PR fixes that when template rendering fails that the pipeline fails instead of just passing.

## 🔍 Reviewer Notes
Create a branch from this branch and `git push --no-verify` with a broken template

OR

Update a gotmpl in values to break.

run docker build for both apl-tools and apl-core. You might need to change the location/name and version of the image in the apl-core docker file on line 1 and line 39.


Dockerfile
```diff
@@ -18,15 +18,32 @@ COPY --chown=app . .
...
+# Run tests with the CI-specific script that has proper Jest flags
+RUN set -e && \
+    if [ "$SKIP_TESTS" = 'false' ]; then \
+        echo "Setting up test environment..." && \
+        ln -s $APP_HOME/tests/fixtures env && \
+        echo "Running CI tests..." && \
+        npm run test:ci && \
+        echo "Cleaning up test environment..." && \
+        rm env && \
+        echo "Tests completed successfully"; \
```
merll: Could this be simplified? I am not sure if the symlink is necessary. Maybe
```suggestion
        echo "Running CI tests..." && \
        ENV_DIR=$APP_HOME/tests/fixtures npm run test:ci && \
        echo "Tests completed successfully"; \
```
would do.

Dockerfile
```diff
@@ -18,15 +18,32 @@ COPY --chown=app . .
 RUN npm config set update-notifier false
 RUN npm ci --ignore-scripts && npm run compile
 
-RUN if [ "$SKIP_TESTS" = 'false' ]; then ln -s $APP_HOME/tests/fixtures env && npm test && rm env; fi
+RUN set -e && \
+    npm config set update-notifier false && \
+    npm ci --ignore-scripts && \
+    npm run compile
```
merll, line 21: Good idea to run them all in a single command, also to reduce unnecessary layers. But these are now duplicated with above lines 18-19.

Dockerfile
merll: Or even simpler, update the `ENV` in line 11, since this is only an intermediate image used in the multistage build.

Dockerfile
ElderMatt: you are correct, I'll remove it

Dockerfile
ElderMatt: I'll remove the symlink and set the env at line 11.

merll (approved):

PR #2427, opened 2025-08-06
fix: replace fs-extra with fs

CasLubbers:
## 🧹 Checklist


src/common/utils.ts
```diff
@@ -23,6 +23,34 @@ export const asArray = (args: string | string[]): string[] => {
   return Array.isArray(args) ? args : [args]
 }
 
+/**
+ * Validates and sanitizes a file path to prevent path traversal attacks.
+ * Ensures the resolved path stays within allowed base directories.
+ */
+export function validateSecurePath(inputPath: string, allowedBasePaths: string[] = [rootDir]): string {
```
ferruhcihan: When I run `npm run bootstrap-dev` I got below error.

> apl-core@4.8.0 bootstrap-dev
> rm -rf /tmp/otomi-bootstrap-dev; CI=1 VALUES_INPUT=$PWD/tests/bootstrap/input-local-dev.yaml ENV_DIR=/tmp/otomi-bootstrap-dev binzx/otomi bootstrap

```Debugger listening on ws://127.0.0.1:4320/a37a504b-fdb6-464c-9826-bc31e2d49a71
For help, see: https://nodejs.org/en/docs/inspector
[dotenv@17.2.0] injecting env (8) from .env (tip: ⚙️  load multiple .env files with { path: ['.env.local', '.env'] })
  otomi:global:error Error: Invalid path: '/tmp/otomi-bootstrap-dev' resolves to '/tmp/otomi-bootstrap-dev' which is outside allowed directories: /Users/username/workspace/linode/apl-core
  otomi:global:error     at validateSecurePath (/Users/username/workspace/linode/apl-core/src/common/utils.ts:46:11)
  otomi:global:error     at decrypt (/Users/username/workspace/linode/apl-core/src/common/crypt.ts:152:25)
  otomi:global:error     at Object.handler (/Users/username/workspace/linode/apl-core/src/cmd/bootstrap.ts:500:11) +0ms```

ElderMatt: Can confirm that I get the same error

CasLubbers: Ah that is from the new function I created to prevent the issue github reported. I can revert it because the API is publicly available

CasLubbers: Fixed

ferruhcihan (approved):

ElderMatt (approved):

PR #2451, opened 2025-08-12
feat: optionally use cnpg backup plugin

merll:
## 📌 Summary

This is a follow-up to #2299, making the changes optional. This was necessary, due to the sidecar execution failing on some clusters for yet unknown reasons.

## 🧹 Checklist

### DIFF
--- a/helmfile.d/snippets/defaults.yaml
+++ b/helmfile.d/snippets/defaults.yaml
@@ -134,6 +134,7 @@ environments:
               limits:
                 cpu: 100m
                 memory: 200Mi
+            useBackupPlugin: false
             _rawValues: {}
           drone:
             enabled: false
@@ -1437,7 +1438,6 @@ environments:
               retentionPolicy: 7d
               schedule: 0 0 * * *
               pathSuffix: keycloak
-          useCnpgPlugin: false
           gitea:
             enabled: false
             retentionPolicy: 7d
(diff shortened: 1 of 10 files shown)


helmfile.d/snippets/defaults.yaml
```diff
@@ -1437,6 +1437,7 @@ environments:
               retentionPolicy: 7d
               schedule: 0 0 * * *
               pathSuffix: keycloak
+          useCnpgPlugin: false
```
j-zimnowoda: Why not adding it at `apps.cnpg` ?

Ani1357:
Tried updating my existing cluster to this branch. When trying to enable a backup through the UI I get this error:
image

merll:
> Tried updating my existing cluster to this branch. When trying to enable a backup through the UI I get this error: ...

Please see my initial comment (Reviewer Notes) on using the feature branch for API. This is necessary since we set `additionalProperties` to `false` in most cases. Otherwise I would also prefer not to expose this via the API.

ferruhcihan (approved):
Tested and LGTM 👍🏻 
Should we consider to update `tests/fixtures/env/settings/platformBackups.yaml` file?

merll:
> Tested and LGTM 👍🏻 Should we consider to update `tests/fixtures/env/settings/platformBackups.yaml` file?

Yes, I'll add that as well.

helmfile.d/snippets/defaults.yaml
merll: I now did.

PR #2465, opened 2025-08-21
feat: add Wiz setup and configuration for Linode Kubernetes Integration

CasLubbers:
## 📌 Summary

Added a wiz install step to our integration.yml

## 🧹 Checklist


.github/workflows/integration.yml
```diff
@@ -253,6 +254,126 @@ jobs:
             sleep 10
           done
           echo LINODE_CLUSTER_CONTEXT=`kubectl config current-context` >> $GITHUB_ENV
+      - name: Set up Wiz
+        if: ${{ inputs.install_profile != 'no-apl' }}
+        run: |
+          # Install required tools for Wiz setup
+          if ! command -v jq &>/dev/null; then
+            sudo apt-get update && sudo apt-get install -y jq
+          fi
```
merll: Pretty sure this is not needed. Please check existing steps.

.github/workflows/integration.yml
```diff
@@ -253,6 +254,126 @@ jobs:
...
+            echo "Installing Helm..."
+            HELM_VERSION="v3.14.2"
+            ARCH="amd64"
+            OS="linux"
+            HELM_TAR="helm-${HELM_VERSION}-${OS}-${ARCH}.tar.gz"
+            curl -fsSL -o "$HELM_TAR" "https://get.helm.sh/${HELM_TAR}"
+            tar -zxvf "$HELM_TAR"
+            sudo mv "${OS}-${ARCH}/helm" /usr/local/bin/helm
+            rm -rf "$HELM_TAR" "${OS}-${ARCH}"
+          fi
```
merll: Most certainly not needed. Other workflow steps use Helm.

.github/workflows/integration.yml
```diff
@@ -253,6 +254,126 @@ jobs:
...
+            enabled: true
+          autoCreateConnector:
+            connectorName: "<connectorName>"
+          webhook:
+            clusterExternalId: "<clusterExternalId>"
+          wiz-admission-controller:
+            enabled: true
+          kubernetesAuditLogsWebhook:
+            enabled: true
+          '
```
merll: If these are static files, they could also be stored in the repo. `sed` can also be used in these files to make substitutions.

.github/workflows/integration.yml
```diff
@@ -253,6 +254,126 @@ jobs:
...
+            imageIntegrityWebhook:
+              enabled: false
+              policies:
+                - my-image-trust-policy
+          '
+
+          WIZ_ADMISSION_CONTROL_TEMPLATE='
+          wizApiToken:
+            clientId: "xudilmphvfal5nrrkhilvzcjjm5cr6unt4wv3gu6uwxsgj6gqcdgm"
+            clientToken: "pZe1En7rjp6bzyWlhkBg1SwiS5efODqdyC3KhWPUEoaN7kNkhqlx6y4xyc1eiZEi"
```
merll: Are these the real tokens?

.github/workflows/integration.yml
```diff
@@ -253,6 +254,126 @@ jobs:
...
+          # Uninstall existing Wiz releases if they exist
+          if helm status "$RELEASE_NAME" -n "$NAMESPACE" &>/dev/null; then
+            echo "⚠️ Helm release '$RELEASE_NAME' already exists. Uninstalling..."
+            helm uninstall "$RELEASE_NAME" -n "$NAMESPACE"
+          fi
+
+          if helm status "wiz-lke-ac" -n "$NAMESPACE" &>/dev/null; then
+            echo "⚠️ Helm release 'wiz-lke-ac' already exists. Uninstalling..."
+            helm uninstall "wiz-lke-ac" -n "$NAMESPACE"
+          fi
```
merll: On a fresh cluster, why would we need an uninstall. And why would one first create the namespace, and then do the uninstall?

ferruhcihan (approved):
I couldn't test it with a cluster or Wiz setup but LGTM from my side, aside from the comments mentioned by @merll.

.github/workflows/integration.yml
```diff
@@ -107,6 +107,7 @@ env:
   BOT_EMAIL: ${{ vars.BOT_EMAIL }}
   BOT_USERNAME: ${{ vars.BOT_USERNAME }}
   DEV_DOMAINS: ${{ secrets.DEV_DOMAINS }}
+  LINODE_API_TOKEN: ${{ secrets.LINODE_TOKEN }}
```
merll: Where is this used?

.github/workflows/integration.yml
CasLubbers: Removed the helm installs. I think because the namespace is not in the helm chart. I just copied over the script and removed some parts.

.github/workflows/integration.yml
CasLubbers: This one is needed in the environment see the document for reference: https://collaborate.akamai.com/confluence/spaces/INFOSEC/pages/1205355383/Linode+-+Kubernetes+Clusters+AutoEnroll+with+Wiz+CSPM+Tool

.github/workflows/integration.yml
CasLubbers: Yes I added them to the secrets just to be sure. But don't think you can use them outside our company

.github/workflows/integration.yml
CasLubbers: NVM I see I removed that part from the original script

merll (approved):
Looks good now!

PR #2500, opened 2025-09-05
fix: update apl-operator before other upgrades

merll:
## 📌 Summary

With this change, the apl-operator checks its own ArgoCD application against the desired version extracted from the values repository. If the revision does not match, it updates itself and awaits a restart with the new version.

This fixes issues with the ArgoCD apps revisions, referring to the charts from this GitHub repository, not matching the values brought along by the apl-operator image.

## 🧹 Checklist

### DIFF
--- a/src/cmd/apply-as-apps.ts
+++ b/src/cmd/apply-as-apps.ts
@@ -1,9 +1,9 @@
 import { mkdirSync, rmSync, existsSync } from 'fs'
-import { writeFile } from 'fs/promises'
+import { readFile, writeFile } from 'fs/promises'
 import { cleanupHandler, prepareEnvironment } from 'src/common/cli'
 import { logLevelString, terminal } from 'src/common/debug'
 import { hf } from 'src/common/hf'
-import { isResourcePresent, k8s, patchContainerResourcesOfSts } from 'src/common/k8s'
+import { ensureAplOperatorRevision, isResourcePresent, k8s, patchContainerResourcesOfSts } from 'src/common/k8s'
 import { getFilename, loadYaml } from 'src/common/utils'
 import { getImageTag, objectToYaml } from 'src/common/values'
 import { appPatches, genericPatch } from 'src/applicationPatches.json'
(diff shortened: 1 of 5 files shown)


src/common/k8s.ts
```diff
@@ -526,6 +526,44 @@ export async function waitForArgoCDAppHealthy(
...
+    version: 'v1alpha1',
+    namespace: 'argocd',
+    plural: 'applications',
+    name: 'apl-operator-apl-operator',
+  })
+  if (process.env.NODE_ENV !== 'development' && application?.status?.sync?.status !== 'Synced') {
+    throw new Error('apl-operator is not yet in Synced state')
+  }
+  const targetRevision = application?.spec?.source?.targetRevision
+  const needsUpdate = targetRevision !== expectedRevision
```
CasLubbers: small nitpick but it reads better for me when you're checking the know good value to the potentially missing value. So turning this around expectedRevision !== targetRevision

src/cmd/apply-as-apps.ts
```diff
@@ -164,12 +164,37 @@ const writeApplicationManifest = async (release: HelmRelease, otomiVersion: stri
...
+export const applyAsApps = async (argv: HelmArguments): Promise<boolean> => {
   const helmfileSource = argv.file?.toString() || 'helmfile.d/'
   d.info(`Parsing helm releases defined in ${helmfileSource}`)
   setup()
   const otomiVersion = await getImageTag()
-
+  try {
+    const needsUpdate = await ensureAplOperatorRevision(
+      env.APPS_REVISION || otomiVersion,
+      async () => {
```
CasLubbers: Callbacks make functions really hard to read and can lead to unexpected things. Do we really need it? Maybe we can split up the ensureAplOperatorRevision in two functions. That first check if it needs updating and the second one updates it and then you can pass through the values.
If you don't think that is a good idea I would at least suggest placing this callback in a seperate function with good naming.

CasLubbers (changes requested):

CasLubbers (approved):
This reads very nicely thanks! ✅

PR #2501, opened 2025-09-05
feat: replace install job by apl-operator

CasLubbers:
## 📌 Summary

- Added an installer class to the apl-operator which executes the installation
- After installation is done the git credentials secrets are created and passed down to operator
- Removed retry mechanism in install function as its now done in installer
- Removed bootstrapping in operator and is now done in installer
- Installing is not done when its finished
- Welcome message is now posted in configmap.
- There is not waiting/checking anymore if APL is functional that must now be done by the user itself.

## 🧹 Checklist

### DIFF
--- a/apps.yaml
+++ b/apps.yaml
@@ -146,7 +146,7 @@ appsInfo:
     chartName: knative-operator
   kserve:
     title: Kserve
-    appVersion: 0.15.2
+    appVersion: 0.16.0
     repo: http://github.com/kserve/kserve
     maintainers: Kserve
     relatedLinks:
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -59,10 +59,10 @@ dependencies:
     version: v1.18.1
     repository: https://knative.github.io/operator
   - name: kserve-crd
-    version: v0.15.2
+    version: v0.16.0
     repository: oci://ghcr.io/kserve/charts/kserve-crd
   - name: kserve
-    version: v0.15.2
+    version: v0.16.0
     repository: oci://ghcr.io/kserve/charts/kserve
   - name: kube-prometheus-stack
     version: 79.5.0
--- a/charts/kserve/Chart.yaml
+++ b/charts/kserve/Chart.yaml
@@ -1,9 +1,9 @@
 apiVersion: v1
-name: kserve-resources
-version: v0.15.2
 description: Helm chart for deploying kserve resources
 keywords:
-  - kserve
-  - modelmesh
+- kserve
+- modelmesh
+name: kserve
 sources:
-  - http://github.com/kserve/kserve
+- http://github.com/kserve/kserve
+version: v0.16.0
--- a/charts/kserve/README.md
+++ b/charts/kserve/README.md
@@ -2,14 +2,14 @@
 
 Helm chart for deploying kserve resources
 
-![Version: v0.15.2](https://img.shields.io/badge/Version-v0.15.2-informational?style=flat-square)
+![Version: v0.16.0](https://img.shields.io/badge/Version-v0.16.0-informational?style=flat-square)
 
 ## Installing the Chart
 
 To install the chart, run the following:
 
 ```console
-$ helm install kserve oci://ghcr.io/kserve/charts/kserve --version v0.15.2
+$ helm install kserve oci://ghcr.io/kserve/charts/kserve --version v0.16.0
 ```
 
 ## Values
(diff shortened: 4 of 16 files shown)


src/cmd/install.ts
```diff
@@ -113,18 +105,7 @@ export const installAll = async () => {
     await retryInstallStep(createCredentialsSecret, initialData.secretName, initialData.username, initialData.password)
     // FIXME: Migrate to use native Git client and stop cd-ing around
     cd(rootDir)
```
merll: These two can go, too, since the `cd` was only required for Helmfile to function properly.
```suggestion
```

src/cmd/install.ts
```diff
@@ -113,18 +105,7 @@ export const installAll = async () => {
...
-        labelOpts: ['pkg=apl-operator'],
-        logLevel: logLevelString(),
-        args: hfArgs,
-      },
-      { streams: { stdout: d.stream.log, stderr: d.stream.error } },
-    )
-    await retryIsOAuth2ProxyRunning()
-    await retryInstallStep(restartOtomiApiDeployment, k8s.app())
-    await printWelcomeMessage(initialData.secretName, initialData.domainSuffix)
+    await createWelcomeConfigMap(initialData.secretName, initialData.domainSuffix)
```
merll: Maybe add a retry to this one as well?

src/common/hf.ts
```diff
@@ -14,7 +14,13 @@ import { existsSync, mkdirSync, writeFileSync } from 'fs'
 import { applyServerSide } from './k8s'
 
 const replaceHFPaths = (output: string, envDir = env.ENV_DIR): string => output.replaceAll('../env', envDir)
-export const HF_DEFAULT_SYNC_ARGS = ['sync', '--concurrency=1', '--sync-args', '--disable-openapi-validation --qps=20']
+export const HF_DEFAULT_SYNC_ARGS = [
+  'sync',
+  '--concurrency=1',
+  '--reuse-values', // Preserve values from existing releases on retry - makes install idempotent
+  '--sync-args',
+  '--disable-openapi-validation --qps=20',
```
merll, line 23: Was this way before, but better would be
```suggestion
  '--disable-openapi-validation',
  '--qps=20',
```

src/operator/installer.ts
```diff
@@ -0,0 +1,130 @@
...
+    try {
+      const configMap = await getK8sConfigMap('apl-operator', 'apl-installation-status', k8s.core())
+      const status = configMap?.data?.status || 'pending'
+      this.d.info(`Current installation status: ${status}`)
+      if (configMap?.data) {
+        this.d.info(`ConfigMap data:`, configMap.data)
+      }
+      return status
+    } catch (error) {
+      this.d.info('No installation status ConfigMap found, assuming pending')
```
merll: I have seen some instances of temporary Kube API hickups leading to this. As 404 is covered in `getK8sConfigMap`, anything else should probably lead to a retry instead.

src/operator/installer.ts
```diff
@@ -0,0 +1,130 @@
...
+    }
+  }
+
+  private async getInstallationStatus(): Promise<string> {
+    try {
+      const configMap = await getK8sConfigMap('apl-operator', 'apl-installation-status', k8s.core())
+      const status = configMap?.data?.status || 'pending'
+      this.d.info(`Current installation status: ${status}`)
+      if (configMap?.data) {
+        this.d.info(`ConfigMap data:`, configMap.data)
```
merll: It currently is echoed to the logs, but seems unnecessarily verbose. I would suggest
```suggestion
        this.d.debug(`ConfigMap data:`, configMap.data)
```
also I would move this above line 63 and also let it show up on an empty response, so it is more clear.

src/common/hf.ts
CasLubbers: I also tried this but then I got the error that --qps=20 is not a valid argument... I'll add a comment above it

src/cmd/install.ts
CasLubbers: I was thinking about this. Now when something fails it will always retry so the retryinstallstep is maybe not necessary. The problem is only that the install command is not idempotent. So it will always retry everything even if its already done.

src/common/hf.ts
merll: Oh, nevermind then. They are both arguments to Helm and not Helmfile, passed on as `--sync-args`.

CasLubbers: I think it's  because it needs to be passed as an argument to the sync-args

src/operator/installer.ts
```diff
@@ -0,0 +1,123 @@
...
+    this.aplOps = aplOps
+    this.d.info('Initializing APL Installer')
+  }
+
+  public async reconcileInstall(): Promise<void> {
+    let attemptNumber = 0
+
+    while (true) {
+      attemptNumber += 1
+      this.d.info(`Starting installation attempt ${attemptNumber}`)
```
ElderMatt: Not really important, but wont the installation attempt reset if the apl-operator pod is killed? I think it would be better to print this message after using getInstallStatus() and setting the attemptNumber again.

CasLubbers: fixed

ElderMatt:
I have the following error during an upgrade from 4.12.0 to this branch. 
image
image

ElderMatt (approved):

PR #2521, opened 2025-09-15
feat: use keycloakx chart instead of bitnami

ElderMatt:
## 🧹 Checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,13 @@
 
 All notable changes to this project will be documented in this file. See [standard-version](https://github.com/conventional-changelog/standard-version) for commit guidelines.
 
----
+### [4.10.1](https://github.com/linode/apl-core/compare/v4.10.0...v4.10.1) (2025-09-17)
+
+
+### Bug Fixes
+
+* apl-operator upgrade ([f601e38](https://github.com/linode/apl-core/commit/f601e38c93d30f453eafd2dd2ceb7a03efc270de))
+
 ## [4.10.0](https://github.com/linode/apl-core/compare/v4.9.0...v4.10.0) (2025-08-21)
 
 ### Bug Fixes
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "apl-core",
-  "version": "4.10.0",
+  "version": "4.10.1",
   "lockfileVersion": 3,
   "requires": true,
   "packages": {
     "": {
       "name": "apl-core",
-      "version": "4.10.0",
+      "version": "4.10.1",
       "license": "Apache-2.0",
       "dependencies": {
         "@apidevtools/json-schema-ref-parser": "14.2.0",
--- a/package.json
+++ b/package.json
@@ -178,5 +178,5 @@
     }
   },
   "type": "commonjs",
-  "version": "4.10.0"
+  "version": "4.10.1"
 }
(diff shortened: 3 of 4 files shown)


chart/chart-index/Chart.yaml
```diff
@@ -46,9 +46,10 @@ dependencies:
   - name: jaeger-operator
     version: 2.46.0
     repository: https://jaegertracing.github.io/helm-charts
-  - name: keycloak
+  - name: keycloakx
+    alias: keycloak
     version: 24.9.0
```
merll: The chart version is numbered differently in this repository. It probably should be
```suggestion
    version: 7.1.3
```

values/keycloak/keycloak.gotmpl
```diff
@@ -4,44 +4,59 @@
...
-{{- if or $v.otomi.linodeLkeImageRepository $v.cluster.defaultStorageClass }}
-global:
-  {{- if $v.otomi.linodeLkeImageRepository }}
-  security:
-    allowInsecureImages: true
-  {{- end }}
-  storageClass: {{ $v.cluster.defaultStorageClass }}
+image:
+  repository: {{ if $v.otomi.linodeLkeImageRepository }}{{ $v.otomi.linodeLkeImageRepository }}/docker/keycloak/keycloak{{ else }}quay.io/keycloak/keycloak{{ end }}
+  tag: "26.3.3"
```
merll: Upstream charts are often upgraded just because the app version changes, and if we set this here, we do not benefit from that.
```suggestion
```

Also this version is a newer patch than we are using currently, so please update `apps.yaml` accordingly.

chart/chart-index/Chart.yaml
ferruhcihan: The version in the chart dependency (24.9.0) doesn't match the actual keycloakx chart version (7.1.3).

ferruhcihan (commented):
Except for the version mismatch, the code looks good.

values/keycloak/keycloak.gotmpl
```diff
@@ -4,44 +4,59 @@
...
 
-{{- if or $v.otomi.linodeLkeImageRepository $v.cluster.defaultStorageClass }}
-global:
-  {{- if $v.otomi.linodeLkeImageRepository }}
-  security:
-    allowInsecureImages: true
-  {{- end }}
-  storageClass: {{ $v.cluster.defaultStorageClass }}
+image:
+  repository: {{ if $v.otomi.linodeLkeImageRepository }}{{ $v.otomi.linodeLkeImageRepository }}/docker/keycloak/keycloak{{ else }}quay.io/keycloak/keycloak{{ end }}
```
merll: To my understanding of https://bits.linode.com/CNS/orcs/pull/101, it should be
```suggestion
  repository: {{ if $v.otomi.linodeLkeImageRepository }}{{ $v.otomi.linodeLkeImageRepository }}/quay/keycloak/keycloak{{ else }}quay.io/keycloak/keycloak{{ end }}
```

values/keycloak/keycloak.gotmpl
```diff
@@ -4,44 +4,59 @@
...
+  tag: "26.3.3"
+  pullPolicy: IfNotPresent
```
merll: This is the default, so I guess the entire block could be optional depending on ORCS.

values/keycloak/keycloak.gotmpl
ElderMatt: you are correct

merll (approved):
Looks good. Tested both install and upgrade. Great work!

ferruhcihan (approved):

PR #2541, opened 2025-09-22
feat: use native minio chart instead of bitnami

ElderMatt:
## 📌 Summary

Now uses native minio helm chart instead of bitnami chart.


ElderMatt:
Waiting on the merging of keycloak

PR #2560, opened 2025-10-01
feat: updated external-dns chart to use official chart 

Ani1357:
## 🧹 Checklist

### DIFF
--- a/values/thanos/thanos.gotmpl
+++ b/values/thanos/thanos.gotmpl
@@ -2,8 +2,15 @@
 {{- $t:= $v.apps.thanos }}
 
 {{- if $v.otomi.linodeLkeImageRepository }}
-image: 
-  registry: "{{- $v.otomi.linodeLkeImageRepository }}/docker"
+image:
+  registry: "{{- $v.otomi.linodeLkeImageRepository }}/quay"
+  repository: "thanos/thanos"
+  tag: v0.36.1
+{{- else }}
+image:
+  registry: "quay.io"
+  repository: "thanos/thanos"
+  tag: v0.36.1
 {{- end }}
 
 ## @section Thanos Query parameters
@@ -28,7 +35,7 @@ queryFrontend:
 ## @section Thanos Compactor parameters
 compactor:
   enabled: true
-  ## By default, there is NO retention set for object storage data. This means that data is stored forever, 
+  ## By default, there is NO retention set for object storage data. This means that data is stored forever,
   ## which is a valid and recommended way of running Thanos.
   retentionResolutionRaw: {{ $t.compactor.retentionResolutionRaw }}
   retentionResolution5m: {{ $t.compactor.retentionResolution5m }}


charts/external-dns/Chart.yaml
```diff
@@ -1,30 +1,31 @@
...
-- name: common
-  repository: oci://registry-1.docker.io/bitnamicharts
-  tags:
-  - bitnami-common
-  version: 2.x.x
-description: ExternalDNS is a Kubernetes addon that configures public DNS servers
-  with information about exposed Kubernetes services to make them discoverable.
-home: https://bitnami.com
-icon: https://dyltqmyl993wv.cloudfront.net/assets/stacks/external-dns/img/external-dns-stack-220x234.png
+appVersion: 0.19.0
```
merll, line 10: Could you please also update this in `apps.yaml`?

values/external-dns/external-dns.gotmpl
```diff
@@ -44,26 +39,23 @@ priorityClassName: otomi-critical
...
-{{- else }}
-  security:
-    allowInsecureImages: false
+  repository: "{{- $v.otomi.linodeLkeImageRepository }}/k8s/external-dns"
+{{- end }}
+
+extraArgs:
+{{- range $dns | get "zoneIdFilters" list }}
+- --zone-id-filter={{ . }}
 {{- end }}
```
merll, line 43: This will result in invalid syntax if the list is empty. This might work (not tested however):
```suggestion
{{- with $dns | get "zoneIdFilters" nil }}
extraArgs:
{{- range . }}
- --zone-id-filter={{ . }}
{{- end }}
{{- end }}
```

values/external-dns/external-dns.gotmpl
```diff
@@ -10,29 +10,24 @@ sources:
...
 
 rbac:
   create: true
-  pspEnabled: false
 
-containerSecurityContext:
+securityContext:
   allowPrivilegeEscalation: false
   readOnlyRootFilesystem: true
   capabilities:
```
merll: Seems we can remove this, as it is already set by default.

merll:
I tested this with a Helm install, and encountered two issues:
* ORCS cannot be used yet, as we have only added the Bitnami image
* The pod fails to start up with the following messages, when I update the Application

```
time="2025-10-03T12:11:49Z" level=error msg="Edgegrid Init Failed"                                                                                                                
time="2025-10-03T12:11:49Z" level=fatal msg="Unable to create instance using environment or .edgerc file"
```

Could you also update the Chart index with the source?

charts/external-dns/templates/deployment.yaml
```diff
@@ -1,949 +1,301 @@
...
-              value: {{ .Values.akamai.clientToken }}
-            {{- end }}
-            {{- if or (.Values.akamai.clientSecret) (.Values.akamai.secretName) }}
-            - name: EXTERNAL_DNS_AKAMAI_CLIENT_SECRET
-              valueFrom:
-                secretKeyRef:
-                  name: {{ template "external-dns.secretName" . }}
-                  key: akamai_client_secret
-            {{- end }}
-            {{- end }}
```
merll, line 402: This part here is essential, and before it was in the Bitnami chart. We will have to add it to our templated values now instead, following the instructions for the providers that we support:
https://github.com/kubernetes-sigs/external-dns/blob/master/charts/external-dns/README.md#providers

values/external-dns/external-dns.gotmpl
Ani1357: tested locally using tests/fixtures by:
1. providing an empty array for `zoneIdFilters`
2. omitting `zoneIdFilters`
3. providing some dummy values

In all cases the templates seems to render correctly

values/external-dns/external-dns.gotmpl
Ani1357: removed

values/external-dns/external-dns.gotmpl
merll: The Helm chart may render this out correctly, and technically a key without a value is valid for most YAML parsers. It is just uncommon to do in Helm charts or their values.

values/external-dns/external-dns-raw.gotmpl
```diff
@@ -0,0 +1,54 @@
+{{- $v := .Values }}
+{{- $d := $v.dns}}
+
+{{- with $d.provider | get "akamai" nil }}
+resources:
+- apiVersion: v1
+  kind: Secret
+  metadata:
+    name: AKAMAI-DNS
```
merll: Results in error on attempting to use this:
```
Failed sync attempt to ffeecf0e0f68417b743b22afe29b55949c016377: one
      or more objects failed to apply, reason: Deployment.apps "external-dns" is invalid:
      spec.template.spec.containers[0].env[3].valueFrom.secretKeyRef.name: Invalid
      value: "AKAMAI-DNS": a lowercase RFC 1123 subdomain must consist of lower case
      alphanumeric characters, ''-'' or ''.'', and must start and end with an alphanumeric
      character (e.g. ''example.com'', regex used for validation is ''[a-z0-9]([-a-z0-9]*[a-z0-9])?(\.[a-z0-9]([-a-z0-9]*[a-z0-9])?)*'')
      (retried 5 times)
```

Ani1357: Sorry, missed that from some copy pasting. Updated

ElderMatt (approved):

PR #2561, opened 2025-10-01
fix: team namespace not being created

ElderMatt:


helmfile.tpl/helmfile-init.yaml.gotmpl
```diff
@@ -27,3 +27,4 @@ releases:
     chart: ../charts/raw
     values:
     - ../values/k8s/k8s-raw.gotmpl
+    - ../values/k8s/k8s-raw-teams.gotmpl
```
CasLubbers, line 38: We can create here a new release k8s-team for example and then add the label team to the k8s-team release and base. Then we can fetch that one based on labels. This way you don't need to create a new helmfile.

src/cmd/apply-teams.ts
```diff
@@ -0,0 +1,62 @@
...
+
+  if (!existsSync(helmfileSource)) {
+    errors.push(`Helmfile teams template not found at: ${helmfileSource}`)
+  }
+
+  d.info(`Parsing team namespaces defined in ${helmfileSource}`)
+
+  const output: ProcessOutputTrimmed = await hf(
+    { fileOpts: helmfileSource, args: 'template' },
+    { streams: { stderr: d.stream.error } },
```
CasLubbers, line 30: Then here you need to add:
labelOpts: ['team=true'],

src/operator/apl-operator.ts
```diff
@@ -87,6 +87,7 @@ export class AplOperator {
       await ensureTeamGitOpsDirectories(env.ENV_DIR)
 
       await commit(false, {} as HelmArguments) // Pass an empty object to clear any stale parsed args
+      await this.aplOps.applyTeams()
```
CasLubbers: Should we put this also in the if for applyTeamsOnly?

values/k8s/k8s-raw.gotmpl
```diff
@@ -40,47 +40,7 @@ resources:
...
-    imagePullSecrets:
-    {{- if $v.otomi | get "globalPullSecret" nil }}
-      - name: otomi-pullsecret-global
-    {{- end }}
-    {{- if $v.apps.harbor.enabled }}
-      - name: harbor-pullsecret
-    {{- end }}
-    {{- end }}
-  {{- end }}
+
```
CasLubbers, line 43: This got me thinking. Should we also reconcile this file in the operator? So adding it to the apply. These resources shouldn't be deleted.

CasLubbers (changes requested):
Nice find

src/operator/apl-operator.ts
ElderMatt: Maybe better yes

PR #2571, opened 2025-10-06
fix: installation reattempts and race condition with operator

merll:
## 📌 Summary

This PR implements small changes in the installation process to make it more robust against temporary failure and race conditions:
* Final steps after the main installation are retried with more focus on the failing command, whereas before it would have reattempted the entire installation: Final commit to the values repo, cloning of the Chart repo, creating the initial admin secret, restarting API.
* If the initial admin secret exists, it is patched instead of failing the installation.
* APL Operator is no longer installed together with all other apps, because its actions could in rare circumstances conflict with the installer that has not finished. Also it creates network policies, that do not consider the installer which needs access to certain resources. APL Operator is installed after all other apps, the installation is also reattempted on failure separately.
* On retrying the entire installation, the delay is now fixed to 30 seconds, whereas before it would make the first reattempt after one second. 30 seconds may give a failed Helm install more time to perform a rollback.
* e2e elements have been removed from the install as they were not used but triggered unnecessary Kubernetes API calls. They could still be applied using the `otomi apply-e2e` command if ever needed.

## 🧹 Checklist


src/cmd/install.ts
```diff
@@ -42,6 +42,24 @@ const setup = (): void => {
...
+  fn: (...args: Args) => Promise<T>,
+  ...args: Args
+): Promise<T> => {
+  const d = terminal(`cmd:${cmdName}`)
+  return await retry(
+    async () => {
+      return await fn(...args)
+    },
+    {
+      retries: 5,
```
ferruhcihan: Looks great overall! Minor suggestion: We might consider using environment-based retry configuration instead of hardcoded values.

src/cmd/install.ts
```diff
@@ -42,6 +42,24 @@ const setup = (): void => {
...
+      onRetry: async (e, attempt) => {
+        d.info(`Retrying (${attempt}/5))...`)
```
ferruhcihan: There is an extra parenthesis `)`

src/cmd/install.ts
merll: I replaced these values with environment variables. They get a little lengthy, since practical values differ on the expected failure conditions. But it is better that way.

PR #2609, opened 2025-10-20
feat: update PostgreSQL to new base images

merll:
## 📌 Summary

This PR updates platform and image catalog images for CloudnativePG databases from the deprecated Debian `bullseye` base images to the recent `trixie`. Since that includes newer libraries, reindexing is required.

Runtime upgrades were changed to run asynchronously per application. This avoids unnecessary waiting cycles (i.e. one app still waiting to sync, while another would be ready), and also upgrade failures of one app affecting another.

NOTE: Upgrade notes should include the following post-upgrade information:

* Workloads using CloudnativePG have their images updated automatically, when using the image catalog (i.e. without manually set `imageName`). These workloads will show a warning in their logs on every connect, and every client will also receive a warning message about needing to run the SQL script below. Only for platform databases, this is performed automatically.
* The following script needs to be run for all CloudnativePG clusters, and for each database (including `postgres`). The duration and expected load depends on the size of the database.
  ```sql
  REINDEX DATABASE <database name>;
  ALTER DATABASE <database name> REFRESH COLLATION VERSION;
  ```
  For performing this step for a cluster without any added or renamed databases in one step, with `kubectl` access to the cluster the following command lines can be used. Before the SQL operations, the primary pod needs to be determined:
  ```bash
  kubectl get cluster -n <team namespace> <cluster name> -o jsonpath='{.status.currentPrimary}'
  kubectl exec -n <namespace> <primary pod> -- psql -d postgres -S -c 'REINDEX DATABASE postgres' -c 'ALTER DATABASE postgres REFRESH COLLATION VERSION'
  kubectl exec -n <namespace> <primary pod> -- psql -d <cluster name> -S -c 'REINDEX DATABASE <cluster name>' -c 'ALTER DATABASE <cluster name> REFRESH COLLATION VERSION'
  ```

## 🧹 Checklist

### DIFF
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -106,7 +106,7 @@ dependencies:
     repository: https://grafana.github.io/helm-charts
   - name: rabbitmq-cluster-operator
     alias: rabbitmq
-    version: 3.20.1
+    version: 4.4.34
     repository: https://charts.bitnami.com/bitnami
   - name: sealed-secrets
     version: 2.17.7


src/common/runtime-upgrades/runtime-upgrades.ts
```diff
@@ -137,4 +138,24 @@ export const runtimeUpgrades: RuntimeUpgrades = [
...
+        post: async (context: RuntimeUpgradeContext) => {
+          await updateDbCollation('gitea', 'gitea-db', 'gitea', context.debug)
+        },
+      },
+      'keycloak-keycloak-otomi-db': {
+        post: async (context: RuntimeUpgradeContext) => {
+          await updateDbCollation('keycloak', 'keycloak-db', 'keycloak', context.debug)
+        },
+      },
+      'harbor-harbor-otomi-db': {
```
CasLubbers, line 154: Harbor is not installed by default. Does it ignore this one then if its not there?

merll: That used to be an issue, but that was already fixed for the Minio upgrade in `src/common/runtime-upgrade.ts`. Non-existent applications are silently ignored. We might want to introduce more verbosity or flexibility on that in the future, but so far on all instances where we need it, applications not being there simply implied they do not need a migration.

src/common/runtime-upgrades/cloudnative-pg.ts
```diff
@@ -0,0 +1,132 @@
...
+  // Each database must be opened separately for reindexing
+  for (const execDbName of ['postgres', databaseName]) {
+    try {
+      const result = await executePSQLOnPrimary(
+        namespace,
+        clusterName,
+        `REINDEX DATABASE ${execDbName}`,
+        execDbName,
+        'postgres',
+        5000,
```
merll: I had this one time out on the first attempt. I will set this to 10s, but I suspect the DB was not entirely ready. Either way, runtime upgrades will be reattempted if one fails.

PR #2626, opened 2025-10-24
fix: deploy namespaces during upgrade

merll:
## 📌 Summary

This PR enhances the runtime upgrade to also re-deploy namespaces in the pre-upgrade phase. This may be necessary due to applications added in the upgraded version.

## 🔍 Reviewer Notes

The rather large diff results from consolidating the same functionality from multiple places. Both installation and upgrade scenario have been tested.

## 🧹 Checklist

### DIFF
--- a/src/common/runtime-upgrade.ts
+++ b/src/common/runtime-upgrade.ts
@@ -55,23 +55,27 @@ export async function runtimeUpgrade({ when }: RuntimeUpgradeArgs): Promise<void
 
     // Application-specific upgrade operations - wait for all applications in this upgrade
     if (upgrade.applications) {
-      for (const [applicationName, applicationUpgrade] of Object.entries(upgrade.applications)) {
-        const applicationOperation = applicationUpgrade[when as keyof typeof applicationUpgrade] as (
-          context: RuntimeUpgradeContext,
-        ) => Promise<void>
-        if (
-          apps.find((app) => app.includes(applicationName)) &&
-          applicationOperation &&
-          typeof applicationOperation === 'function'
-        ) {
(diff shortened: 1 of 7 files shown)


src/common/runtime-upgrade.ts
```diff
@@ -19,6 +20,12 @@ export async function runtimeUpgrade({ when }: RuntimeUpgradeArgs): Promise<void
     return
   }
 
+  d.info('Deploying essential manifests')
+  const essentialDeployResult = await deployEssential(['upgrade=true'])
```
CasLubbers, line 24: The label `upgrade` does not exist in the file `helmfile.tpl/helmfile-init.yaml.gotmpl`

merll: It is added: https://github.com/linode/apl-core/pull/2626/files#diff-cedb490af977058390fc03dd3f21c0f52d0d3b773a9f992c191ac6f756b2a083

merll:
> Is it maybe better to add this to the regular apply? So that the resources in the k8s values get reconciled? This way when someone accidentally deletes a namespace or another resource it gets re-applied

Generally, I would agree that namespaces should be reconciled regularly, but firing up Helmfile, despite filtering, takes relatively long, and we should not do too many of these things unnecessarily. Also, if someone deletes one of the core namespaces, there are a lot of additional problems, such as dependent resources waiting to be terminated, possibly stalling the operator to wait. I would suggest addressing this kind of issue in a separate tool, because it should be a rare condition.

If we rewrite this doing a direct create-or-patch on the KubeAPI, it will be much faster and we could also run it each time. There is a follow-up issue for that: https://track.akamai.com/jira/browse/APL-1223

Ani1357 (approved):

PR #2642, opened 2025-10-29
feat: upgrade argocd image updater and leverage argocd application sources

j-zimnowoda:
## 📌 Summary

- Upgrade argocd-image-updater
- Deploy workloads as Application instead of ApplicationsSets, so they are immediately visible in ArgoCD dashboard.
- Set argocd-application-image-updater to store updated tags in the values repo instead of the charts repo to adhere to the single source of truth rule.
- Get rid of the additional `values` property.

depends-on: https://github.com/linode/apl-api/pull/832

## 🔍 Reviewer Notes

- This change leverages [multiple Application sources](https://argo-cd.readthedocs.io/en/stable/user-guide/multiple_sources/) instead of single source, so there are two values files for a given workload. One that is managed by user and the other that is managed by the argocd-application-image-updater
- The values are migrated automatically. 
- upgrade script removes ApplicationSets without impacting Applications.
- user will be instructed to migrate content of `k8s-deployment/.argocd-source-<team-name>-<workload-name>.yaml` from charts repo to  corresponding env/teams/<team-name>/workloads/<workload-name>.managed.yaml file and to delete the `.argocd-source-<team-name>-<workload-name>.yaml`

## 🧹 Checklist

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,4 +1,4 @@
-api: APL-1195-1
+api: main
 console: main
 consoleLogin: main
 tasks: main
--- a/package-lock.json
+++ b/package-lock.json
@@ -202,7 +202,6 @@
       "integrity": "sha512-e7jT4DxYvIDLk1ZHmU/m/mB19rex9sv0c2ftBtjSBv+kVM/902eh0fINUzD7UwLLNR+jU585GxUJ8/EBfAM5fw==",
       "dev": true,
       "license": "MIT",
-      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.27.1",
         "@babel/generator": "^7.28.5",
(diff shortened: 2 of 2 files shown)


Ani1357:
Tested by upgrading existing cluster with running workloads. After the upgrade,the existing workload gets converted as expected in an application with 2 sources (`values` repo and `charts` repo) but since the old image **sha** was already saved in the `charts` repo, the image tag never gets updated. Deleting the `k8s-deployment/.argocd-source-team-demo-<workload-name>.yaml` solves it.

versions.yaml
```diff
@@ -1,4 +1,4 @@
-api: main
+api: APL-1195-1
```
merll: Make sure to revert this before merge.

j-zimnowoda:
> Deleting the `k8s-deployment/.argocd-source-team-demo-<workload-name>.yaml` solves it.

Great finding! Thanks you. 
For this use case I propose to add the upgrade instructions.

Ani1357 (approved):

ferruhcihan (approved):

PR #2645, opened 2025-10-30
feat: add collect traces command

CasLubbers:
## 📌 Summary

Updated kubectl apply with --server-side to prevent errors  and warnings when applying essential manifets.

---

Adds a new `collect traces` command that generates a comprehensive report of unhealthy Kubernetes resources during APL installation failures. The report is automatically stored in a ConfigMap (`apl-operator/apl-traces-report`) for post-mortem analysis.

  - Monitors 8 resource types: Pods, Deployments, StatefulSets, Nodes, Services, PVCs, PVs, and ArgoCD Applications
  - Automatically runs when `install` or `apply` commands fail
  - Can be invoked manually: `binzx/otomi collect traces`
  - Stores structured JSON report with timestamp and summary statistic

## 🔍 Reviewer Notes

  - **Error handling**: ArgoCD checks gracefully skip if CRD doesn't exist (404/403 errors)
  - **Non-blocking**: Troubleshooting errors are caught and logged but don't prevent retries
  - **ConfigMap storage**: Uses create-or-update pattern from existing operator patterns
  - **Minimal output**: Console shows only count; full report stored in ConfigMap
  - **Integration points**: Hooked into retry catch blocks in `install.ts:149-154` and `apply.ts:90-95`

## 🧹 Checklist

### DIFF
--- a/apps.yaml
+++ b/apps.yaml
@@ -123,7 +123,6 @@ appsInfo:
     license: Apache 2.0
     about: Istio is an open platform for providing a uniform way to integrate microservices, manage traffic flow across microservices, enforce policies and aggregate telemetry data. Istio's control plane provides an abstraction layer over the underlying cluster management platform.
     integration: APL has security best practices built in, and is designed for intrusion. Istio is used by APL as a service mesh to deliver mTLS enforcement for all traffic that is deemed compromisable, egress control to force teams to choose explicit egress endpoints, and advanced routing capabilities such as weight based load balancing (A/B or blue/green testing). Istio is part of the core of APL and can not be disabled.
-    chartName: istiod
   jaeger:
     title: Jaeger Operator
     dependencies: Open Telemetry (otel)
--- a/ci/src/update-helm-chart-deps.mjs
+++ b/ci/src/update-helm-chart-deps.mjs
@@ -83,12 +83,247 @@ async function copyLinodeCfwTemplates(chartDir) {
   return false
 }
 
+const CHART_GROUPS = {
+  istio: ['base', 'istiod', 'gateway'],
+  kserve: ['kserve', 'kserve-crd'],
+  'cloud-firewall': ['cloud-firewall-controller', 'cloud-firewall-crd'],
+}
 const CHART_SKIP_PRECHECK = ['cloud-firewall-crd', 'kserve-crd']
 const CHART_POST_FUNCS = {
   kyverno: renderKyvernoCrdTemplates,
   'otel-operator': renderOtelCrdTemplates,
   'kserve-crd': copyKserveCrdTemplates,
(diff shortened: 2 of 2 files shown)


src/cmd/apply.ts
```diff
@@ -86,6 +87,12 @@ export const apply = async (): Promise<void> => {
         await applyAll()
       } catch (e) {
         d.error(e)
+        // Generate troubleshooting report on apply failure
+        try {
+          await troubleshoot()
```
j-zimnowoda: would the `collectTraces` name better reflect this function?

src/cmd/troubleshoot.ts
```diff
@@ -0,0 +1,398 @@
...
+ * Get pods with issues across all namespaces
+ */
+async function getPodsWithIssues(): Promise<ResourceReport[]> {
+  const coreApi = k8s.core()
+  const response = await coreApi.listPodForAllNamespaces()
+  const pods: ResourceReport[] = []
+
+  await Promise.all(
+    response.items.map(async (pod) => {
+      const namespace = pod.metadata?.namespace || 'default'
```
j-zimnowoda: 'default' => 'unknown'

src/cmd/traces.ts
```diff
@@ -0,0 +1,398 @@
...
+async function getPodsWithIssues(): Promise<ResourceReport[]> {
+  const coreApi = k8s.core()
+  const response = await coreApi.listPodForAllNamespaces()
+  const pods: ResourceReport[] = []
+
+  await Promise.all(
+    response.items.map(async (pod) => {
+      const namespace = pod.metadata?.namespace || 'default'
+      const podName = pod.metadata?.name || 'unknown'
+      const issues: string[] = []
```
j-zimnowoda, line 40: For developers to get the issue context would suggest to rename:
issues => podIssues

src/cmd/troubleshoot.ts
```diff
@@ -0,0 +1,398 @@
...
+ * Main troubleshoot function
+ */
+export async function troubleshoot(): Promise<void> {
+  const d = terminal(`cmd:${cmdName}:troubleshoot`)
+
+  try {
+    d.info('Starting troubleshooting scan...')
+
+    // Gather all failed resources
+    const [pods, deployments, statefulSets, nodes, services, pvcs, pvs, argoApps] = await Promise.all([
```
j-zimnowoda: It may happen that one kubectl API call fails OR CRD is not available (e.g.: Application).
I am wondering what to do it it. Should all report fail OR perhaps we should still store partial report?

src/cmd/apply.ts
CasLubbers: Yes maybe we should rename the function to collect. So you can do otomi collect traces. And the function name would be collectTraces 👍

PR #2651, opened 2025-10-31
fix: PostgreSQL image upgrade considered for 4.12 release

merll:
## 📌 Summary

The upgrade of PostgreSQL base images has already been merged into the 4.12 release. As the current runtime upgrades consider it being only in 4.13, currently post-upgrade runs on every operator run. This is not a huge issue, but unnecessary.

## 🧹 Checklist


merll:
Will be cherry-picked onto release branch.

PR #2683, opened 2025-11-05
feat: removing kiali

Ani1357:
## 🧹 Checklist

### DIFF
--- a/tests/fixtures/env/settings/cluster.yaml
+++ b/tests/fixtures/env/settings/cluster.yaml
@@ -9,6 +9,6 @@ spec:
     k8sContext: linode-dev
     owner: akamai
     provider: linode
-    # linode:
-    #     tier: enterprise
+    linode:
+        tier: standard
     defaultStorageClass: ''


values-changes.yaml
```diff
@@ -378,3 +378,5 @@ changes:
   - version: 44
     additions:
       - cluster.defaultStorageClass: ''
+    deletions:
```
merll, line 382: I think we would need a new version here
```suggestion
  - version: 45
    deletions:
```

merll:
Please also remove remaining occurrences in `tests/`. `tests/fixtures` should clean up after considering above comment and the running `npm run migrate-values`.

values-changes.yaml
ElderMatt: I agree

Ani1357: added new version

Ani1357:
> Please also remove remaining occurrences in `tests/`. `tests/fixtures` should clean up after considering above comment and the running `npm run migrate-values`.

occurrences in `tests/` should be already removed now.

ElderMatt:
I also did not see anything in the tests and test/fixtures folder

PR #2708, opened 2025-11-10
fix: helm operation in progress

ElderMatt:
## 📌 Summary

Fix an issue where installer is restarted and Helm gets stuck in an error.
` Error: UPGRADE FAILED: another operation (install/upgrade/rollback) is in progress`

### DIFF
--- a/src/operator/installer.ts
+++ b/src/operator/installer.ts
@@ -1,14 +1,7 @@
 import * as process from 'node:process'
 import { terminal } from '../common/debug'
 import { hfValues } from '../common/hf'
-import {
-  createUpdateConfigMap,
-  createUpdateGenericSecret,
-  deletePendingHelmReleases,
-  getK8sConfigMap,
-  getK8sSecret,
-  k8s,
-} from '../common/k8s'
+import { createUpdateConfigMap, createUpdateGenericSecret, getK8sConfigMap, getK8sSecret, k8s } from '../common/k8s'
 import { AplOperations } from './apl-operations'
 import { getErrorMessage } from './utils'
 
@@ -66,7 +59,6 @@ export class Installer {
         this.d.error(`Installation attempt ${attemptNumber} failed:`, errorMessage)
         await this.updateInstallationStatus('failed', attemptNumber, errorMessage)
         this.d.warn(`Installation attempt ${attemptNumber} failed, retrying in 1 second...`, getErrorMessage(error))
-        await deletePendingHelmReleases()
 
         // Wait 1 second before retrying
         await new Promise((resolve) => setTimeout(resolve, 1000))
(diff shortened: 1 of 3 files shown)


ElderMatt:
This PR is dependent on APL-940.

CasLubbers (approved):
Works like a charm. Deleted the operator multiple times and install continues

src/common/k8s.ts
```diff
@@ -241,6 +294,49 @@ export const getHelmReleases = async (): Promise<Record<string, any>> => {
...
 
+export const getK8sHelmReleases = async (): Promise<Record<string, any>> => {
+  const coreApi = k8s.core()
+
+  try {
+    const secretsResponse = await coreApi.listSecretForAllNamespaces({
+      labelSelector: 'owner=helm,status',
+    })
+
+    const releases: Record<string, any> = {}
```
j-zimnowoda: Use `Record<string, HelmRelease>` instead of `Record<string, any>`

src/common/k8s.ts
```diff
@@ -241,6 +294,49 @@ export const getHelmReleases = async (): Promise<Record<string, any>> => {
   return status
 }
 
+export const getK8sHelmReleases = async (): Promise<Record<string, any>> => {
```
j-zimnowoda: Use `Record<string, HelmRelease>` instead of `Record<string, any>`

src/common/k8s.ts
```diff
@@ -164,6 +177,46 @@ export const getDeploymentState = async (): Promise<DeploymentState> => {
...
+  namespace: string
+  status: string
+  app_version: string
+  revision: number
+  first_deployed?: string
+  last_deployed?: string
+  chart?: string
+}
+
+export const checkOperationsInProgress = async (): Promise<void> => {
```
j-zimnowoda: I propose to rename it to: `deletePendingHelmReleases`

j-zimnowoda: This function could also return a number of deleted release. It will be easier to unit tests and provides valuable return value.

j-zimnowoda (changes requested):
Please fix types

src/operator/installer.ts
```diff
@@ -59,6 +66,7 @@ export class Installer {
         this.d.error(`Installation attempt ${attemptNumber} failed:`, errorMessage)
         await this.updateInstallationStatus('failed', attemptNumber, errorMessage)
         this.d.warn(`Installation attempt ${attemptNumber} failed, retrying in 1 second...`, getErrorMessage(error))
+        await deletePendingHelmReleases()
```
j-zimnowoda: Given recent changes about initial install, it seems that this cantons would be called at: https://github.com/linode/apl-core/blob/APL-1150/src/cmd/install.ts#L124

@CasLubbers @merll could you confirm ?

PR #2721, opened 2025-11-14
feat: remove velero

ferruhcihan:
## 📌 Summary

This PR removes the velero app from the platform

Ticket: https://track.akamai.com/jira/browse/APL-1260
PRs: [apl-api](https://github.com/linode/apl-api/pull/846) | [apl-console](https://github.com/linode/apl-console/pull/681)

## 🧹 Checklist

### DIFF
--- a/values-changes.yaml
+++ b/values-changes.yaml
@@ -391,3 +391,4 @@ changes:
       - 'obj.provider.linode.buckets.velero'
       - 'platformBackups.persistentVolumes'
       - 'teamConfig.{team}.backups'
+      - 'apps.httpbin'
--- a/apps.yaml
+++ b/apps.yaml
@@ -12,7 +12,7 @@ appsInfo:
     integration: Alertmanager can be activated to send alerts to configured receivers. It is configured by APL to use the global values found under settings/alerts. A team can override global settings to send alerts to their own endpoints.
   argocd:
     title: Argo CD
-    appVersion: 3.1.8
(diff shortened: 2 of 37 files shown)


ElderMatt (approved):

values-changes.yaml
```diff
@@ -385,3 +385,7 @@ changes:
       - 'apps.kured'
       - 'apps.kiali'
       - 'apps.falco'
+  - version: 46
+    deletions:
```
merll: I think this can be added to version 45. Otherwise we would have to make one version per app to be removed, which seems excessive.
```suggestion
```

merll:
With the removal of Velero, `platformBackups.persistentVolumes` in the values becomes obsolete.

merll:
Also there is `team.backups`, which is obsolete.

values-changes.yaml
merll: Although I am not sure if we should add one for every PR, but still it could lead to conflicting numbers. Maybe it is better to stick to releases.

ferruhcihan: Updated.

ferruhcihan:
> Please also drop `platformBackups.persistentVolumes` from schema and values.

> Also there is `team.backups`, which is obsolete.

Removed.

ElderMatt (approved):

merll (approved):

merll:
Looks good now!

PR #2730, opened 2025-11-19
feat: make platform db connection pooling configurable

merll:
## 📌 Summary

This PR replaces hard-coded PostgreSQL parameters with platform values, which are identically set by default, but can be updated by the platform admin. Platform applications using PostgreSQL databases, i.e. Gitea, Keycloak, and Harbor also have the hard-coded connection pool settings replaced with variables. However, the defaults of maximum database connections on the app-side are reduced, for taking a deduction from the database `max_connections` into account.
The default for `superuser_reserved_connections` of 3. This has lead to Harbor not starting, when it was attempting to max out the connection pool.

The max connections on the app-side is set to 28, considering the default of `superuser_reserved_connections` plus one in case there are other maintenance connections. A backport for current releases, which does not involve schema changes, and just updates the hard-coded defaults, is published in #2731.

Other changes to the schema are fixes of wrong locations or cleanups of unused values. In addition it showed that values introduced for each database's `backupSidecarResources` were not applied in the template. This has been added here.

## 🔍 Reviewer Notes

When upgrading to this feature branch, defaults should be added to the values repo. Changes are expected to require restarting the application pods, but not the databases as default settings are identical to current hard-coded parameters.

Note also when upgrading from `main`, you need to set `specVersion` in `env/settings/versions.yaml` back to `44` for re-running schema migrations. Otherwise validations fail.

## 🧹 Checklist


ElderMatt:
I think all the test files can be removed.

helmfile.d/snippets/defaults.yaml
```diff
@@ -336,6 +328,10 @@ environments:
           keycloak:
             adminUsername: otomi-admin
             theme: otomi
+            database:
```
CasLubbers: It would make more sense for me if these connections values are also under the database objects? Then they would end up in the same file where all the other database settings are stored.

merll: They are not database settings, but application settings. When putting these in the same place, I would find it ambiguous how they relate to `max_connections` of the database. And applications can have different ways of connection pooling, that may also change independently from the database (note how Keycloak differs from Gitea and Harbor), while all PostgreSQL databases work the same way.
It only would make sense if we removed these and set them implicitly (e.g. set these numbers based on `max_connections` of the database) but I would prefer not to, as this seems like excessively opinionated template magic to me.

We could rename it however, to make it more clear. Maybe call it `databaseConnection`?

CasLubbers: Ah I saw it wrong. I thought these values where set on the database. But that is the postgresql settings.
Indeed having `databaseMaxConnection` etc would help. Seeing a database object gives me the idea that you can set database things in the application settings.

tests/fixtures/env/teams/demo/workloadValues/wd.yaml
```diff
@@ -1 +1 @@
-a: b
+{}
```
j-zimnowoda: Do not know why but this values should be reverted

values-changes.yaml
```diff
@@ -392,3 +392,7 @@ changes:
       - 'platformBackups.persistentVolumes'
       - 'teamConfig.{team}.backups'
       - 'apps.httpbin'
+      - 'apps.gitea.images'
```
j-zimnowoda, line 398: Please add new version number. Otherwise we need to re-trigger value migration on dev cluster

merll: I can do that, but in regard to other incremental changes here we need to do that either way. It is hard to sync with the version number when multiple PRs are open which use it, therefore I would have preferred to do that only once per release. We should be done with app removals by now however.

tests/fixtures/env/teams/demo/workloadValues/wd.yaml
merll: They are a result of running the workload values migration a second time.

values-changes.yaml
j-zimnowoda: I noticed that not all values migration are idempotent. That is the reason behind my request. But we can also fix the migration functions. Just let me know what you prefer.

merll: I moved it to a newer version, since there was already one for the Jaeger removal.

helmfile.d/snippets/defaults.yaml
merll: I changed the values structure, naming variables like `databaseMaxConnections` instead of nesting them. This is probably more clear then.

PR #2753, opened 2025-11-24
fix: skip initial values after install and isolate bootstrap errors

merll:
## 📌 Summary

This PR isolates checks for an existing installation and based on this, skips loading initial values and an attempt to run an initial installation. This fixes the following issues that can occur:
* After an installation, the branch cannot be changed because the original values take precedence.
* If the bootstrap or validation process fails, the installation status is set to `failed` which then leads to an attempt to re-install the platform, even if that had been completed before.
* When running local dev environments the Gitea password is regenerated on each start, overriding environment variables such as `GIT_PASSWORD`.

## 🧹 Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -163,9 +163,7 @@
     "validate-templates": "ENV_DIR=$PWD/tests/fixtures NODE_ENV=test binzx/otomi validate-templates",
     "validate-templates:all": "set -e; i=31; while [ $i -le 34 ]; do NODE_ENV=test binzx/otomi validate-templates -k 1.$i; i=$(($i+1)); done",
     "validate-values": "ENV_DIR=$PWD/tests/fixtures NODE_ENV=test binzx/otomi validate-values",
-    "bootstrap-dev": "rm -rf /tmp/otomi-bootstrap-dev; VALUES_INPUT=$PWD/tests/bootstrap/input-local-dev.yaml ENV_DIR=/tmp/otomi-bootstrap-dev binzx/otomi bootstrap",
-    "bootstrap-dev-with-repo": "ENV_DIR=/tmp/otomi-bootstrap-dev binzx/otomi bootstrap",
-    "bootstrap-tests-fixtures": "ENV_DIR=$PWD/tests/fixtures binzx/otomi bootstrap"
+    "bootstrap-dev": "rm -rf /tmp/otomi-bootstrap-dev; VALUES_INPUT=$PWD/tests/bootstrap/input-local-dev.yaml ENV_DIR=/tmp/otomi-bootstrap-dev binzx/otomi bootstrap"
   },
   "standard-version": {
     "skip": {


src/operator/main.ts
```diff
@@ -77,8 +77,14 @@ async function main(): Promise<void> {
 
     // Phase 1: Run installation with retry until success
     const installer = new Installer(aplOps)
-    d.info('=== Starting Installation Process ===')
-    await installer.reconcileInstall()
+    const isInstalled = await installer.isInstalled()
```
j-zimnowoda, line 80: The `apl-installation-status` cm may not exists at the very begining, thus the `isInstalled` function may rise error. I see that in the catch block you exit. To my understanding you may enter endless Pod CrashLoop

merll: `getK8sConfigMap` that this function is using is returning `undefined` on a 404 response. That one is handled in the functions used.

src/cmd/bootstrap.ts
```diff
@@ -335,44 +322,34 @@ export const processValues = async (
...
   await deps.writeValues(merge(cloneDeep(allSecrets), cloneDeep(originalInput), cloneDeep({ users })))
   // and do some context dependent post processing:
-  if (deps.isChart) {
-    // to support potential failing chart install we store secrets on cluster
-    if (!(env.isDev && env.DISABLE_SYNC)) await deps.createK8sSecret(DEPLOYMENT_PASSWORDS_SECRET, 'otomi', allSecrets)
-  } else if (!isEmpty(originalInput)) {
-    // cli: when we are bootstrapping from a non empty values repo, validate the input
-    await deps.validateValues()
-    // Ensure newly generated secrets stored in .dec file are encrypted
-    await encrypt()
```
j-zimnowoda, line 345: are you sure that we do need to encrypt ?

src/common/envalid.ts
```diff
@@ -72,5 +72,4 @@ export function cleanEnv<T>(spec: { [K in keyof T]: ValidatorSpec<T[K]> }, optio
 
 export const env = cleanEnv(cliEnvSpec, {})
 
-export const isChart = !!env.VALUES_INPUT
-export const isCli = env.OTOMI_DEV || !isChart
+export const isCli = env.OTOMI_DEV || env.VALUES_INPUT
```
j-zimnowoda: It seems to me that the logic should be like below:
```
isCli = env.OTOMI_DEV || !env.VALUES_INPUT
```

src/cmd/bootstrap.ts
j-zimnowoda: reading it second time and I see that this is part of the block that operator will never enter. All good.

merll: I don't see why we would. This would still be at operator process startup, where we have no changes in the repository, so no new added values.

merll:
> In the package.json the `bootstrap-dev` works but the `bootstrap-dev-with-repo` gives me an error. Do we still need the bootstrap-dev-with-repo?

Same goes for `bootstrap-tests-fixtures`. I am not sure, but I think they were used to update a repository with defaults / secrets. What I should add then is at least an error, rejecting the bootstrap without `VALUES_INPUT`.

j-zimnowoda:
> In the package.json the `bootstrap-dev` works but the `bootstrap-dev-with-repo` gives me an error. Do we still need the bootstrap-dev-with-repo?

Given that  bootstrap is valid only for the installation phase, we do not need the `bootstrap-dev-with-repo`. anymore.

merll:
I dropped the additional bootstrap commands and introduced a validation that requires VALUES_INPUT.

CasLubbers (approved):

PR #2760, opened 2025-11-27
fix: handle upgrade from previous versions

merll:
## 📌 Summary

Latest changes to apl-operator do not consider the case where the ConfigMap `apl-installation-status` does not yet exist. This PR aims to mitigate this.

## 🧹 Checklist


chart/apl/templates/01-installation-status.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: v1
```
j-zimnowoda, line 2: Could you add a comment here ?
```
Lack of this config map would cause skipping the installation phase.
```

merll: It is added now.

PR #2796, opened 2025-12-12
fix: duplicate credentials warning in Argocd

CasLubbers:
## 🧹 Checklist


merll:
In this way of implementing common credentials, does this not counter measures of APL-527, where we try to eliminate secrets from values?

values/argocd/argocd-raw.gotmpl
```diff
@@ -9,3 +10,15 @@ resources:
     data:
       custom-ca-certificates.crt: {{ .Values._derived.caCert | b64enc }}
 {{- end }}
+  - apiVersion: v1
+    kind: Secret
+    metadata:
+      name: argocd-repo-creds-gitea
+      namespace: argocd
+      labels:
+        argocd.argoproj.io/secret-type: repo-creds
```
j-zimnowoda, line 19: Is this label used to instruct argocd about consuming that secret?

CasLubbers: Yes

PR #2800, opened 2025-12-15
feat: add jwt verification for api

CasLubbers:
## 🧹 Checklist

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,4 +1,4 @@
-api: APL-1372
+api: main
 console: main
 consoleLogin: main
 tasks: main


versions.yaml
```diff
@@ -1,4 +1,4 @@
-api: main
+api: APL-1372
```
ferruhcihan: Please remember to revert this line before merging.

dennisvankekem (approved):

versions.yaml
CasLubbers:
```suggestion
api: main
```

PR #2805, opened 2025-12-17
feat: deploy secrets as secretRef in chart values

ElderMatt:
## 📌 Summary

Replaces secrets withs secretRefs for Harbor and Oauth2-Proxy.

Uses environment Variables for Loki.

### DIFF
--- a/apps.yaml
+++ b/apps.yaml
@@ -12,7 +12,7 @@ appsInfo:
     integration: Alertmanager can be activated to send alerts to configured receivers. It is configured by App Platform to use the global values found under settings/alerts. A team can override global settings to send alerts to their own endpoints.
   argocd:
     title: Argo CD
-    appVersion: 3.2.1
+    appVersion: 3.2.3
     repo: https://github.com/argoproj/argo-helm
     maintainers: Argo Project
     relatedLinks:
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -6,7 +6,7 @@ version: 0.1.0
 dependencies:
   - name: argo-cd
     alias: argocd
-    version: 9.1.5
+    version: 9.2.4
     repository: https://argoproj.github.io/argo-helm
   - name: argocd-image-updater
     version: 1.0.1
--- a/charts/argocd/Chart.yaml
+++ b/charts/argocd/Chart.yaml
@@ -1,12 +1,12 @@
 annotations:
   artifacthub.io/changes: |
-    - kind: changed
-      description: Bump argo-cd to v3.2.1
+    - kind: fixed
+      description: Remove Redis references when redis.enabled=false to prevent CreateContainerConfigError
   artifacthub.io/signKey: |
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
 apiVersion: v2
-appVersion: v3.2.1
+appVersion: v3.2.3
 dependencies:
 - condition: redis-ha.enabled
   name: redis-ha
@@ -28,4 +28,4 @@ name: argo-cd
 sources:
 - https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd
 - https://github.com/argoproj/argo-cd
-version: 9.1.5
+version: 9.2.4
--- a/charts/argocd/README.md
+++ b/charts/argocd/README.md
@@ -1100,6 +1100,7 @@ NAME: my-release
 | repoServer.imagePullSecrets | list | `[]` (defaults to global.imagePullSecrets) | Secrets with credentials to pull images from a private registry |
 | repoServer.initContainers | list | `[]` | Init containers to add to the repo server pods |
 | repoServer.lifecycle | object | `{}` | Specify postStart and preStop lifecycle hooks for your argo-repo-server container |
+| repoServer.livenessProbe.enabled | bool | `true` | Enable Kubernetes liveness probe for Repo Server |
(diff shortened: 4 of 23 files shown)


values/loki/loki-raw.gotmpl
```diff
@@ -10,4 +12,20 @@ resources:
...
+        app: loki
+      name: loki-s3-credentials
+    type: Opaque
+    stringData:
+      {{- if eq $obj.type "linode" }}
+      S3_URL: "https://{{ $obj.linode.accessKeyId }}:{{ $obj.linode.secretAccessKey }}@{{ $obj.linode.region }}.linodeobjects.com/{{ $obj.linode.buckets.loki }}"
+      {{- end }}
+      {{- if eq $obj.type "minioLocal" }}
+      S3_URL: "http://otomi-admin:{{ $v.otomi.adminPassword }}@minio.minio.svc.cluster.local:9000/loki"
+      {{- end }}
```
merll: As the main objective is delegating this to Sealed Secrets, the templating would be better left in the `loki.gotmpl`. Therefore, `accessKey` and `secretAccessKey` / `adminPassword` (for minio) should be stored in one variable each. `region` and the bucket name are not sensitive values.

values/loki/loki.gotmpl
```diff
@@ -55,13 +57,9 @@ loki:
       {{- end }}
     {{- if $useObjectStorage }}
     aws:
-      {{- if eq $obj.type "minioLocal" }}
-      s3: http://otomi-admin:{{ $v.otomi.adminPassword }}@minio.minio.svc.cluster.local.:9000/loki
+      s3: ${S3_URL}
```
merll: As mentioned above, probably something like 
```
s3: http://otomi-admin:${OTOMI_ADMIN_PASSWORD}@minio.minio.svc.cluster.local.:9000/loki
```
and 
```
s3: https://${S3_ACCESS_KEY_ID}:${S3_ACCESS_KEY_SECRET}@{{ $obj.linode.region }}.linodeobjects.com/{{ $obj.linode.buckets.loki }}
```

values/oauth2-proxy/oauth2-proxy-raw.gotmpl
```diff
@@ -1,10 +1,21 @@
...
 {{- $cm := $v.apps | get "cert-manager" }}
 {{- $ingress :=  $v.ingress.platformClass }}
 
 resources:
+  - apiVersion: v1
+    kind: Secret
+    metadata:
+      name: oauth2-proxy-client-access
+    type: Opaque
+    stringData:
```
CasLubbers: stringData does not work well with server side apply. See the note at this section: https://kubernetes.io/docs/concepts/configuration/secret/#basic-authentication-secret.
So if its possible its better to use data and encode the strings first. Like you did with the other secrets.

ElderMatt: I can do that, I used stringData because that is what the oauth2-proxy chart uses as well to create their secret.

values/harbor/harbor.gotmpl
```diff
@@ -30,7 +30,7 @@ core:
     {{- end }}
 
   resources: {{- $h.resources.core | toYaml | nindent 4 }}
-  secret: {{ $h | get "core.secret" nil | quote }}
+  existingSecretKey: harbor-core-secret-key
```
j-zimnowoda: `core.existingSecretKey` does not exist in the Harbor helm chart. Did you mean `core.secretName`?


We have not set it before because we are relying on istio mTLS. What is the reason you decided to se it now?

ElderMatt: I mixed some things together in my notes, should now be fixed.

values/harbor/harbor-raw.gotmpl
```diff
@@ -56,10 +56,9 @@ resources:
 - apiVersion: v1
   kind: Secret
   metadata:
-    name: harbor-core-secret-key
+    name: harbor-core-secret
   data:
-    tls.crt: {{ $cm.customRootCA | b64enc }}
-    tls.key: {{ $cm.customRootCAKey | b64enc}}
+    secret: "{{ randAlphaNum 16 | b64enc }}"
```
j-zimnowoda: This will regenerate the secret every time the template is executed. Usually secrets that need to be generated are stored as values. Is there any reason for not adhering this pattern?

values/loki/loki-raw.gotmpl
```diff
@@ -10,4 +12,28 @@ resources:
...
+  {{- end }}
+  {{- if eq $obj.type "minioLocal" }}
+  - apiVersion: v1
+    kind: Secret
+    metadata:
+      labels:
+        app: loki
+      name: loki-s3-minio-credentials
+    type: Opaque
+    stringData:
```
j-zimnowoda: Everywhere else `data` property was used. Why in this case you decided to use `stringData`

ElderMatt: You are right, I just liked using it because Kubernetes automatically encodes it. I will change it

values/harbor/harbor-raw.gotmpl
ElderMatt: I forgot to use the existing value from core.secret. I added it back.

values/loki/loki-raw.gotmpl
j-zimnowoda: thanks

values/harbor/harbor-raw.gotmpl
j-zimnowoda: thanks

values/harbor/harbor.gotmpl
j-zimnowoda: ok

PR #2807, opened 2025-12-17
fix: check for migration on every operator cycle

merll:
## 📌 Summary

This PR changes the apl-operator behavior to check for migrations on every cycle. This avoids errors on upgrades, where mutations are made before the values are validated.

Note that the extra validation step is still skipped on `reconcile`, as it is run later.

## 🧹 Checklist


j-zimnowoda (approved):
One line upper and a huge difference :)

PR #2833, opened 2026-01-08
fix: render list with unique set of source repositories in argocd project

j-zimnowoda:
## 📌 Summary

Many workloads can have exactly the same workload catalog. However the list of permitted source repositories should be contain only unique entries 

## 🔍 Reviewer Notes

See the bin/compare.sh output

## 🧹 Checklist


charts/team-ns/templates/argocd/argocd-project.yaml
```diff
@@ -1,6 +1,14 @@
 {{- $v := .Values }}
 {{- $a := $v.apps.argocd }}
 {{- if $a.enabled }}
+{{- $urls := list }}
+{{- range $i, $item := $v.workloads }}
+  {{- $urls = append $urls $item.url }}
+{{- end }}
+{{- $urls = append $urls $v.gitOps.teamRepoUrl }}
+{{- $urls = append $urls $v.gitOps.valuesRepoUrl }}
+{{- $urls = uniq $urls }}
```
merll: I would also prefer to add `sortAlpha` here, to make sure order changes do not cause updates either. While this is is a list here, values are read from the Git repo and then stored as a map of filename -> workload which is populated asynchronously (so even if the Map maintains the insertion order, this does not help).

charts/team-ns/templates/argocd/argocd-project.yaml
```diff
@@ -1,6 +1,14 @@
 {{- $v := .Values }}
 {{- $a := $v.apps.argocd }}
 {{- if $a.enabled }}
+{{- $urls := list }}
+{{- range $i, $item := $v.workloads }}
+  {{- $urls = append $urls $item.url }}
```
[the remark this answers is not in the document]
merll: I think we can assume these fields exist.

charts/team-ns/templates/argocd/argocd-project.yaml
```diff
@@ -1,6 +1,14 @@
...
+{{- end }}
+{{- $urls = append $urls $v.gitOps.teamRepoUrl }}
+{{- $urls = append $urls $v.gitOps.valuesRepoUrl }}
```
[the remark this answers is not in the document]
merll: Same as above.

charts/team-ns/templates/argocd/argocd-project.yaml
```diff
@@ -13,18 +21,16 @@ spec:
   description: ArgoCD project for team {{ $v.teamId }}
   # Allow manifests to deploy from any Git repos
   sourceRepos:
-  - {{ $v.gitOps.teamRepoUrl }}
-  - {{ $v.gitOps.valuesRepoUrl }}
-  {{- range $v.workloads }}
-  - {{ .url }}
+  {{- range $i, $url := $urls }}
+  - {{ $url }}
   {{- end }}
```
merll: or just
```suggestion
  {{- $urls | toYaml | nindent 4 }}
```

charts/team-ns/templates/argocd/argocd-project.yaml
j-zimnowoda: good point

charts/team-ns/templates/argocd/argocd-project.yaml
j-zimnowoda: done

PR #2870, opened 2026-01-26
feat: remove tempo app

j-zimnowoda:
## 📌 Summary

Depends-on:
https://github.com/linode/apl-core/pull/2870
https://github.com/linode/apl-console/pull/704
https://github.com/linode/apl-api/pull/904

## 🔍 Reviewer Notes
For testing deployment set  `tag: APL-1473`
```
kind: AplApp
metadata:
    name: otomi-console
spec:
    _rawValues:
        image:
            imagePullPolicy: Always
            repository: docker.io/linode/apl-console
            tag: APL-1473
...
```

```
kind: AplApp
metadata:
    name: otomi-api
spec:
    _rawValues:
        image:
            imagePullPolicy: Always
            repository: docker.io/linode/apl-api
            tag: APL-1473
...

```

## 🧹 Checklist


merll:
I am wondering if we should also remove the additional collector `platform-otlp` from `values/otel-operator/otel-operator-raw.gotmpl` and all related values with this. We are deploying this collector which without Tempo just writes everything back to stdout, and we provide no way of changing this.
If a platform admin ever would like to roll out their own tracing solution, they are better off setting up their own OTel collector, and then also have control of the pipeline.

helmfile.d/helmfile-60.teams.yaml.gotmpl
```diff
@@ -135,29 +135,6 @@ releases:
...
-              jsonData:
-                tracesToLogsV2:
-                  datasourceUid: 'loki'
-                  spanStartTimeShift: '1h'
-                  spanEndTimeShift: '-1h'
-                  tags: [{ key: 'k8s.namespace.name', value: 'namespace' }]
-                  filterByTraceID: false
-                  filterBySpanID: false
-                  customQuery: true
-                  query: '{namespace="team-{{ $teamId }}"} |= "$${__span.traceId}"'
```
CasLubbers, line 160: Until line 182 also needs to be removed right? It looks like its all under the same Tempo array item

j-zimnowoda: correct. I provided fix.

j-zimnowoda:
> We also maintain charts/grafana-dashboards. In there the tempo directory should also be removed it still contains the tempo dashboard.

Thanks. I have removed those as well.

j-zimnowoda:
> I am wondering if we should also remove the additional collector `platform-otlp` from `values/otel-operator/otel-operator-raw.gotmpl` and all related values with this. We are deploying this collector which without Tempo just writes everything back to stdout, and we provide no way of changing this. If a platform admin ever would like to roll out their own tracing solution, they are better off setting up their own OTel collector, and then also have control of the pipeline.

Good point. Could you provide separate PR for removing that collector?

PR #2874, opened 2026-01-27
feat: custom kubernetes manifest in the values repo

merll:
## 📌 Summary

This PR enhances the functionality by responding to creation of Kubernetes manifest files in specific directories of the values repository. Files created in `env/manifests/ns/<namespace-name>/` result in creation of an ArgoCD application that creates and deploys everything into the namespace with the directory name. Manifest files stored in `env/manifest/global/` are intended for cluster-wide objects and therefore do not imply a target namespace, and do not create any.
Removing any of the namespace directories removes the ArgoCD application as well. Removing all manifests from `global` does not remove the ArgoCD application, due to risks arising from deleting cluster-wide objects.

In order to avoid unintended changes to unrelated ArgoCD applications, labels are added and used for selection. The platform app updates have been extended accordingly.

## 🔍 Reviewer Notes

Verifying that the namespace matches the directory name has been out of scope of the defined functionality, therefore the manifest can override the target deployment namespace. Additional functionality is planned to be implemented in the API.

## 🧹 Checklist


merll:
Rebased in #2893

PR #2875, opened 2026-01-27
fix: sortUserArraysByName

j-zimnowoda:
## 📌 Summary

The `application/apl-keycloak-operator-apl-keycloak-operator-artifacts` is constantly resending because values for the corresponding chart were not sorted.

See `apl-keycloak-operator-secret` secret at the data.USERS base64 encoded string. 

## 🧹 Checklist

### DIFF
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -6,7 +6,7 @@ version: 0.1.0
 dependencies:
   - name: argo-cd
     alias: argocd
-    version: 9.3.6
+    version: 9.3.7
     repository: https://argoproj.github.io/argo-helm
   - name: argocd-image-updater
     version: 1.0.1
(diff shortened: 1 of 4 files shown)


src/common/repo.ts
```diff
@@ -573,7 +583,7 @@ export async function loadValues(envDir: string, deps = { loadToSpec }): Promise
       await deps.loadToSpec(spec, fileMap)
     }),
   )
-  return sortTeamConfigArraysByName(spec)
+  return sortUserArraysByName(sortTeamConfigArraysByName(spec))
```
CasLubbers: This reads a bit easier for me but for the rest it looks good to me
```suggestion
  sortTeamConfigArraysByName(spec)                                                                                                                                                                        
  sortUserArraysByName(spec)                                                                                                                                                                              
  return spec 
```

j-zimnowoda: Thanks. Indeed it is easier to read.

PR #2876, opened 2026-01-27
feat: run traces command every 5 minutes for 30 minutes

CasLubbers:
## 📌 Summary

- Run traces command every 5 minutes for 30 minutes
- Add creationTimestamp to an event
- It wil also add an empty report if nothing is found
- Reports will not be overriden it will be added to the configmap
## 🧹 Checklist

### DIFF
--- a/ci/src/update-helm-chart-deps.mjs
+++ b/ci/src/update-helm-chart-deps.mjs
@@ -176,7 +176,9 @@ async function checkDependencyUpdates(dependency, allowedUpgradeType) {
     // Get all available versions for the dependency
     allVersions = await $`helm search repo ${dependency.name}/${dependency.name} -l -o json`
       .then((output) => JSON.parse(output.stdout))
-      .then((results) => results.map((entry) => entry.version).filter((version) => semver.valid(version)))
+      .then((results) => results
+        .filter((entry) => semver.valid(entry.version) && entry.name === `${dependency.name}/${dependency.name}`)
+        .map((entry) => entry.version))
   }
 
   if (!allVersions.length) {


src/common/envalid.ts
```diff
@@ -55,6 +55,8 @@ export const cliEnvSpec = {
     desc: 'Target revision to set for ArgoCD applications. If not set, uses the current image tag.',
     default: undefined,
   }),
+  COLLECTION_DURATION_MS: num({ desc: 'Traces collection duration 30 min default in ms', default: 1800000 }),
```
j-zimnowoda: could that be in seconds please?

CasLubbers: Done

src/cmd/traces.ts
```diff
@@ -398,18 +420,70 @@ export async function collectTraces(): Promise<void> {
...
+    // ConfigMap doesn't exist yet, will be created by first collectTraces() call
+  }
+
+  // No ConfigMap yet, use current time (ConfigMap will be created by collectTraces)
+  return Date.now()
+}
+
+export async function runTraceCollectionLoop(): Promise<void> {
+  const d = terminal('cmd:traces:runTraceCollectionLoop')
+  const configMapName = 'apl-traces-report'
```
j-zimnowoda: This should be extracted to a global variable or even to ENV. Consider operator being installed in different NS.

j-zimnowoda: Also take a look at line 417, there are also hardcoded namespace and configmap names.

src/cmd/traces.ts
```diff
@@ -398,18 +420,70 @@ export async function collectTraces(): Promise<void> {
...
+
+  while (Date.now() < endTime) {
+    try {
+      d.info('Running periodic trace collection')
+      await collectTraces()
+    } catch (error) {
+      d.warn('Failed to collect traces:', getErrorMessage(error))
+    }
+
+    const remainingTime = endTime - Date.now()
```
j-zimnowoda: This code prevents setting the last timeout. However it is also also redundant with the `while (Date.now() < endTime)` 

Is it worth introducing more complexity to for very little performance gain?

ElderMatt (approved):

src/cmd/traces.ts
CasLubbers: Done

src/cmd/traces.ts
CasLubbers: Done

PR #2881, opened 2026-01-28
fix: replace kubectl with k8s client library for argocd apps

CasLubbers:
## 🧹 Checklist

### DIFF
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -6,7 +6,7 @@ version: 0.1.0
 dependencies:
   - name: argo-cd
     alias: argocd
-    version: 9.3.6
+    version: 9.3.7
     repository: https://argoproj.github.io/argo-helm
   - name: argocd-image-updater
     version: 1.0.1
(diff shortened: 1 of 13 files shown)


src/cmd/apply-as-apps.ts
```diff
@@ -48,13 +45,43 @@ interface HelmRelease {
...
+    name: string
+    namespace: string
+    labels?: Record<string, string>
+    annotations?: Record<string, string>
+    finalizers?: string[]
+  }
+  spec: Record<string, any>
+}
+
+async function applyArgocdApp(app: ArgocdAppManifest): Promise<void> {
```
j-zimnowoda, line 62: Can you patch something that does not exist (consider very first installation, or app enabled) 
?

CasLubbers: Yes I also did not know this but with Server-Side Apply, a PATCH acts as a declarative apply operation.
If the resource does not exist yet (e.g. first installation), Kubernetes will create it automatically.
This allows us to use a single create-or-update flow.

merll (approved):

src/cmd/apply-as-apps.ts
j-zimnowoda: Fantastic

j-zimnowoda (approved):

PR #2882, opened 2026-01-28
fix: support Loki without object storage

merll:
## 📌 Summary

Loki can run without object storage by using local volumes. It then needs to run in "SingleBinary" mode with one instance. This PR implements this, when object storage is not set up. It also removes values which were not actually functional, as they assumed distributed mode would also work with local storage. The latter does not seem to be supported.

## 🔍 Reviewer Notes

With this change, functionality with object storage should not change. In order to test the different deployment mode, set the platform's object storage provider to "Disabled".

## 🧹 Checklist


Ani1357 (approved):
Tested by deploying a cluster with obj sorage disabled.

PR #2893, opened 2026-01-29
feat: custom kubernetes manifest in the values repo

merll:
## 📌 Summary

This PR enhances the functionality by responding to creation of Kubernetes manifest files in specific directories of the values repository. Files created in env/manifests/ns/<namespace-name>/ result in creation of an ArgoCD application that creates and deploys everything into the namespace with the directory name. Manifest files stored in env/manifest/global/ are intended for cluster-wide objects and therefore do not imply a target namespace, and do not create any.
Removing any of the namespace directories removes the ArgoCD application as well. Removing all manifests from global does not remove the ArgoCD application, due to risks arising from deleting cluster-wide objects.

In order to avoid unintended changes to unrelated ArgoCD applications, labels are added and used for selection. The platform app updates have been extended accordingly.

## 🔍 Reviewer Notes

Verifying that the namespace matches the directory name has been out of scope of the defined functionality, therefore the manifest can override the target deployment namespace. Additional functionality is planned to be implemented in the API.

## 🧹 Checklist

### DIFF
--- a/src/cmd/apply-as-apps.ts
+++ b/src/cmd/apply-as-apps.ts
@@ -397,7 +397,7 @@ export const applyAsApps = async (argv: HelmArguments): Promise<boolean> => {
   return true
 }
 
-export const addGitOpsArgocdApps = async (
+export const addGitOpsApps = async (
   appNames: Set<string>,
   namespaceDirs: string[],
   deps = { getArgocdGitopsManifest, applyArgocdApp },
@@ -428,7 +428,7 @@ export const addGitOpsArgocdApps = async (
   )
 }
 
-export const removeGitOpsArgocdApps = async (appNames: Set<string>) => {
+export const removeGitOpsApps = async (appNames: Set<string>) => {
   d.info(`Removing GitOps apps: ${Array.from(appNames).join(', ')}`)
   await Promise.allSettled(
     appNames.values().map(async (appName) => {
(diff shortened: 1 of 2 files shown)


src/cmd/apply-as-apps.ts
```diff
@@ -351,6 +397,76 @@ export const applyAsApps = async (argv: HelmArguments): Promise<boolean> => {
   return true
 }
 
+export const applyGitOpsApps = async (
```
CasLubbers: This function is quite big making hard to understand for me. If I read it correctly it actually does three things: discovers apps, creates apps, removes apps 
I think line 404 untill 425 could be a separate function:
```
const { toAdd, toRemove, namespaceDirs } = await calculateGitOpsAppsDiff(env.ENV_DIR, deps) 
```
And then the two if blocks `addGitOpsApps` and `removeGitOpsApps`. Could have their own function to add and remove the GitOps apps. 
Then there will be less nesting and more single responsibility per function.

ElderMatt: I agree with splitting it up in multiple functions, but I think it would be nicer to create the discover apps function from line 407 to 425. Because you have to reuse the namespacedirs in the addGitOpsApps as well.

merll: I have restructured this part accordingly.

CasLubbers (approved):

PR #2899, opened 2026-02-02
feat: remove thanos

CasLubbers:
## 📌 Summary

https://github.com/linode/apl-api/pull/908
https://github.com/linode/apl-console/pull/708
## 🧹 Checklist

### DIFF
--- a/values-changes.yaml
+++ b/values-changes.yaml
@@ -421,6 +421,9 @@ changes:
     deletions:
       - 'apps.tempo'
       - 'obj.provider.linode.buckets.tempo'
+  - version: 50
+    deletions:
+      - apps.loki.persistence
   - version: 51
     deletions:
       - 'apps.thanos'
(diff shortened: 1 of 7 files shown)


merll (approved):

values-changes.yaml
```diff
@@ -421,3 +421,8 @@ changes:
     deletions:
       - 'apps.tempo'
       - 'obj.provider.linode.buckets.tempo'
+  - version: 50
```
merll: Please also set this version in `defaults.yaml`.

merll (changes requested):
I had not been paying enough attention to the changes in the `charts/loki` folder. These files are managed upstream and should not be modified by us.

values-changes.yaml
j-zimnowoda: I propose to use `51` because `50` to ensure the values migration happens on clusters that rely on main branch

j-zimnowoda:
I see that retention has changed 
```
@@ spec.retention @@
! ± value change
- 6h
+ 15d
```
I agreed that 6h is very little for metrics retention. Not sure about 15d and RAM usage.

PR #2901, opened 2026-02-02
feat: byo catalog

ferruhcihan:
## 📌 Summary

This PR introduces support for Bring Your Own (BYO) catalog functionality, allowing users to configure custom Helm chart repositories instead of relying solely on the default APL charts repository. 

## 🔍 Reviewer Notes

The related `apl-api` branch was created only for testing purposes and not to break cluster creation. The `api` and `console` changes will be implemented later.

## 🧹 Checklist

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,4 +1,4 @@
-api: APL-1458
+api: main
 console: main
 consoleLogin: main
 tasks: main


versions.yaml
```diff
@@ -1,4 +1,4 @@
-api: main
+api: APL-1458
```
ferruhcihan: Reminder: Please revert this before merging.

helmfile.d/snippets/defaults.gotmpl
```diff
@@ -274,6 +274,12 @@ environments:
       {{- end }}
       otomi:
         adminPassword: {{ randAlphaNum 32 }}
+      catalogs:
```
CasLubbers: This is perfectly fine to add to the defaults.yaml it contains no templating

helmfile.d/snippets/defaults.yaml
```diff
@@ -1210,4 +1210,4 @@ environments:
         users: []
         # TODO: update this when schema version changes
         versions:
-          specVersion: 49
+          specVersion: 50
```
CasLubbers: This should be 53 because of new PR's merged

values-schema.yaml
```diff
@@ -229,6 +228,33 @@ definitions:
...
+        description: 'A lowercase name that starts with a letter and may contain dashes.'
+        pattern: '^[a-z0-9]([-a-z0-9]*[a-z0-9])$'
+      url:
+        type: string
+        description: 'Git repository URL for the catalog'
+        pattern: '^(https?|git|ssh)://.*\.git$'
+      branch:
+        type: string
+        description: 'Git branch or tag to use'
+        default: 'main'
```
CasLubbers: We already set the defaults. Same for enabled below.

src/cmd/migrate.ts
```diff
@@ -667,6 +668,75 @@ const setLokiStorageSchemaMigration = async (values: Record<string, any>): Promi
   }
 }
 
+const GITHUB_CATALOG_URL = 'https://github.com/linode/apl-charts.git'
```
CasLubbers: This is a good contender for an environment variable

src/cmd/migrate.ts
```diff
@@ -667,6 +668,75 @@ const setLokiStorageSchemaMigration = async (values: Record<string, any>): Promi
...
+const SEALED_SECRET_NAME = 'default-catalog-credentials'
+
+const createCatalogSealedSecret = async (
```
CasLubbers, line 672: This is something new creating the sealedsecret in code. Is that something we want to start to do?
We can also keep it simple and create directly the secret.
Or we can create the secret through helmfile

helmfile.d/snippets/defaults.gotmpl
ElderMatt: Agree, no templating is done so can be put in the yaml directly

ferruhcihan: Moved to the defaults.yaml.

helmfile.d/snippets/defaults.yaml
ferruhcihan: Updated.

values-schema.yaml
ferruhcihan: Removed.

src/cmd/migrate.ts
ferruhcihan: Good point. Moved to the environment variable.

src/cmd/migrate.ts
ferruhcihan: The sealed secret approach is intentional here. This secret is only needed temporarily for the Gitea catalog, with this feature we're introducing the option to use the public linode apl-charts repo instead, which needs no credentials. 

Creating it as a SealedSecret:
- Users can see and manage it through the values repository
- Removal is simple
- No additional migration needed if we later move platform secrets to sealed secrets

Creating a regular secret or through helmfile:
- Would not persist in the values repository and not visible.
- Would require a separate migration if we later move platform secrets to sealed secrets.
- Would couple it to the template lifecycle and need a code change to clean it up later.

src/cmd/migrate.ts
j-zimnowoda: I would also rename it GITHUB_CATALOG_URL => GIT_REPO_CATALOG_URL

ferruhcihan: Renamed.

ElderMatt (approved):

versions.yaml
ferruhcihan: Reverted.

PR #2914, opened 2026-02-05
feat: BYO Git

CasLubbers:
Only merge with: 
- https://github.com/linode/apl-console/pull/719
- https://github.com/linode/apl-api/pull/915
## 📌 Summary

- Gitea is now optional. Its default set to true in the defaults.yaml
- Consolidated all the Git config under otomi.git.* jsonpath
- Renamed all Gitea reference that were basically a reference to Git to Git
- Added secret apl-git-credentials and configmap apl-git-config.
- Added a migration script that these secret and configmap getting created after migration
## 🧹 Checklist

### DIFF
--- a/versions.yaml
+++ b/versions.yaml
@@ -1,5 +1,5 @@
-api: APL-1510
-console: APL-1511
+api: main
+console: main
 consoleLogin: main
 tasks: main
 tools: main


versions.yaml
```diff
@@ -1,4 +1,4 @@
-api: main
+api: APL-1510
```
CasLubbers: Reminder to change this back

values-schema.yaml
```diff
@@ -2800,6 +2785,42 @@ properties:
...
+              The base URL of the Git repository (without credentials). Auto-derived for internal Gitea.
+            pattern: '^https?://.+'
+          username:
+            type: string
+            description: |
+              Username for authenticating with the Git repository.
+              Defaults to 'otomi-admin' for internal Gitea.
+          password:
+            type: string
+            description: Password or token for authenticating with the Git repository
```
merll, line 2807: This suggests that also an API token can be used. This should however then be sent as a header and not as a basic auth (through the URL). I have not tested this part, but I did not see any code making this possible.

CasLubbers: For github this works using a PAT. I used the Token (classic) with full repo access. I did not test gitlab or bitbucket.

merll: For now we can leave it as it is. We should open up for both authentication methods, but that's not a large change that can always be added later.

merll (approved):

PR #2917, opened 2026-02-05
feat: add apiToken handling to external-dns-raw template

### LINKED ISSUE #2916: fix/external-dns-cloudflare-api-token
**Describe the bug:** 
otomi apply fails when External-DNS is enabled with the Cloudflare provider. The Helm values template external-dns-raw.gotmpl references a non-existent field .apiSecret, while the Helm schema and documentation define the Cloudflare credential as .apiToken. This causes template rendering to fail and blocks installation.

**To Reproduce**
Helm Install APL-Core v4.13.1

failed to render values files "../values/external-dns/external-dns-raw.gotmpl":
executing "stringTemplate" at <.apiSecret>: map has no entry for key "apiSecret"

**Screenshots:** (if applicable, add screenshots to help explain your problem)
Not applicable (CLI error).

**Cluster(s):**
Cloud: self-managed / on-prem (talos based on this [guide](https://github.com/vaheed/talos-kubernetes-cluster/blob/main/longhorn-3w.md#apl-core-platform)  )

**Additional context:**
The Cloudflare External-DNS configuration in values/external-dns/external-dns-raw.gotmpl expects .apiSecret, but the Helm values schema and External-DNS documentation use .apiToken. Replacing .apiSecret with .apiToken (and setting CF_API_TOKEN) resolves the issue. This appears to be a template bug in APL-Core v4.13.1.
What's wrong: The /values/external-dns/external-dns-raw.gotmpl template file at line 44 is trying to access .apiSecret for the Cloudflare provider, but the Helm schema defines the field as .apiToken. This is a bug in APL-Core v4.13.1.

vaheed:
## 📌 Summary

Fixes a template bug in `external-dns-raw.gotmpl` where the Cloudflare provider references
`.apiSecret`, which is not defined in the Helm values schema. The correct field is `.apiToken`.

### Why this is needed
When External-DNS is enabled with Cloudflare, `otomi apply` fails during Helm values rendering:
executing "stringTemplate" at <.apiSecret>: map has no entry for key "apiSecret"
### Related issue

Fixes #2916

## 🧹 Checklist


values/external-dns/external-dns-raw.gotmpl
```diff
@@ -41,7 +41,7 @@ resources:
   data:
     CF_API_TOKEN: "{{ . | b64enc }}"
   {{- end }}
-  {{- with .apiSecret }}
+  {{- with .apiToken }}
```
merll:
```suggestion
  {{- with . | get "apiSecret" nil }}
```

values/external-dns/external-dns-raw.gotmpl
merll, line 90: This should be replaced with

```suggestion
  {{- with . | get "apiToken" nil }}
```

merll:
Thank you for the report @vaheed and for the PR. Both `apiSecret` and `apiToken` exist in the schema, for allowing for both authentication methods to [CloudFlare supported by external-dns](https://kubernetes-sigs.github.io/external-dns/latest/docs/tutorials/cloudflare/#creating-cloudflare-credentials). Using one or the other value is used to determine which variables to set.

Of course the template rendering should not fail if either of the values is missing. My suggestions above should fix this issue.

merll (approved):

j-zimnowoda (approved):

PR #2926, opened 2026-02-09
fix: compare timestamps when re-encrypting

CasLubbers:
## 🧹 Checklist

### DIFF
--- a/apps.yaml
+++ b/apps.yaml
@@ -230,7 +230,7 @@ appsInfo:
     chartName: trivy-operator
   otel:
     title: OpenTelemetry Operator
-    appVersion: 0.138.0
+    appVersion: 0.144.0
     repo: https://github.com/open-telemetry/opentelemetry-operator
     maintainers: Grafana labs
     relatedLinks:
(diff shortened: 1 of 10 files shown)


src/common/crypt.ts
```diff
@@ -179,33 +178,37 @@ export const encrypt = async (path = env.ENV_DIR, ...files: string[]): Promise<v
...
-          return true
+
+        if (decExists) {
+          // Compare timestamps in whole seconds to avoid sub-second filesystem precision issues
+          // (matchTimestamps syncs them after decryption, so newer .dec means modifications)
+          const encTS = await stat(file)
+          const decTS = await stat(`${file}.dec`)
+          const encSec = Math.floor(encTS.mtimeMs / 1000)
+          const decSec = Math.floor(decTS.mtimeMs / 1000)
+          if (decSec > encSec) {
```
j-zimnowoda, line 191: I am not sure if I understand it correctly. Unmodified encrypted files will always have lower timestamp than decrypted. Are they going to be encrypted again?

CasLubbers: Yes, but when we decrypt we set the same timestamp as the encrypted file. See the `matchTimestamps` function. So when you run decrypt it will always end up with the same timestamp.
And when something changes in an .dec file. When we run encrypt it will always see that it has a newer timestamp so it will encrypt it.

j-zimnowoda: That makes sense. Thanks for explaining it.

PR #2927, opened 2026-02-09
fix: reattempt password validity reset on any expiration error

merll:
## 📌 Summary

This PR aims to fix a sporadically reoccurring issue where the Gitea password is marked as expired while Gitea is either initialized or upgraded. This was implemented before, but only once during an upgrade.

## 🔍 Reviewer Notes

The function `resetGiteaPasswordValidity` has not been modified but moved because it is no longer only part of an upgrade. It was using utility functions of the `k8s` module; therefore `waitTillGitRepoAvailable` was moved out as well for avoiding mutual dependencies. This function was modified to invoke the password validity reset.

## 🧹 Checklist

### DIFF
--- a/charts/apl-network-policies/templates/networkpolicies/otomi-api.yaml
+++ b/charts/apl-network-policies/templates/networkpolicies/otomi-api.yaml
@@ -15,32 +15,36 @@ spec:
   policyTypes:
     - Ingress
   ingress:
-    # Allow access from Istio public ingress gateway
-    - from:
+    - ports:
+        # Allow ingress to api endpoint
+        - protocol: TCP
+          port: 8080
+        # Allow ingress to api endpoint for health checks
+        - protocol: TCP
+          port: 9000
+      # Allow access from Istio public ingress gateway
+      from:
         - namespaceSelector:
             matchLabels:
               name: istio-system
           podSelector:
             matchLabels:
               app.kubernetes.io/instance: istio-ingressgateway-public
-    # Allow access from gitea for webhook notifications
-    - from:
+        # Allow access from gitea for webhook notifications
         - namespaceSelector:
             matchLabels:
               name: gitea
           podSelector:
             matchLabels:
               app: gitea
-    # Allow monitoring access for metrics scraping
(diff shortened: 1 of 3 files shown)


src/cmd/commit.ts
```diff
@@ -5,7 +5,8 @@ import { encrypt } from 'src/common/crypt'
 import { terminal } from 'src/common/debug'
 import { env } from 'src/common/envalid'
 import { hfValues } from 'src/common/hf'
-import { createUpdateConfigMap, createUpdateGenericSecret, k8s, waitTillGitRepoAvailable } from 'src/common/k8s'
+import { waitTillGitRepoAvailable } from 'src/common/gitea'
```
j-zimnowoda, line 8: This function is called only once during the apl-operator start. This is sufficient to cover the initial installation use case.

However, you have mentioned that, Gitea password may get invalidated also during the Gitea upgrade. On APL upgrade, is apl-operator restarted before of after the core app upgrade?

merll: It is restarted before the upgrade. Therefore it may actually run too early. I also considered adding this function to the reconciliation loop, but the first attempt did not seem like a good place. Still checking where it fits well, without being triggered too often.

j-zimnowoda: Got it. It is also very interesting that the Gitea upgrade can cause password invalidation.

merll: I checked, and I could add it to the reconciliation cycle. That has a major drawback, because the Git client used there appears to be suppressing the specific message ("remote: Update your password"), and instead just gives an authentication error. The problem with that is, that the operator will then spawn Gitea CLI commands over and over again, without it having any positive impact. We can accept that, or even have apl-operator reset the password as it knows what it should be. To me that seems much of an overlap with the responsibilities of apl-gitea-operator however.

j-zimnowoda: You are right. Gitea operator should handle this kind scenarios.

CasLubbers:
Need to update this in the BYO Git branch

PR #2945, opened 2026-02-13
feat: set keycloak login url to otomi realm

CasLubbers:
## 🧹 Checklist

### DIFF
--- a/core.yaml
+++ b/core.yaml
@@ -180,7 +180,7 @@ adminApps:
   - name: istio
     tags: [ingress, egress, routing, security, tls, observability, policies]
   - name: keycloak
-    path: /admin/otomi/console
+    path: /admin/otomi/console/
     tags: [auth, sso]
     # isShared: true
     ownHost: true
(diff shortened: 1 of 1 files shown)


core.yaml
```diff
@@ -219,7 +218,7 @@ adminApps:
           - authorization
   - name: loki
     tags: [logging, telemetry, observability]
-    deps: [grafana, prometheus, otel]
```
j-zimnowoda, line 222: Loki depends on Otel. See https://github.com/linode/apl-core/blob/main/values/otel-operator/otel-operator-raw.gotmpl#L60

core.yaml
```diff
@@ -261,7 +260,7 @@ adminApps:
     tags: [secrets, security, observability]
     ownHost: true
   - name: otel
-    tags: [logging, tracing, telemetry, observability]
```
j-zimnowoda, line 264: otel is used for logging

j-zimnowoda: provided fix

core.yaml
j-zimnowoda: fix provided

PR #2957, opened 2026-02-18
fix: escape single quotes in template string generation

CasLubbers:
## 🧹 Checklist

### DIFF
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -77,7 +77,7 @@ dependencies:
     version: 3.13.0
     repository: https://kubernetes-sigs.github.io/metrics-server
   - name: oauth2-proxy
-    version: 10.1.2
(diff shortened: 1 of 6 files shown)


src/common/utils.ts
```diff
@@ -107,12 +107,12 @@ export const gucci = async (
     // Cannot template if key contains regex characters, so skip
     if (stringContainsSome(k, ...'^()[]$'.split(''))) return ''
     const val = typeof v === 'object' ? JSON.stringify(v) : v
-    return `-s ${k}='${val ?? ''}'`
+    const escaped = String(val ?? '').replaceAll("'", "'\\''")
```
j-zimnowoda: It would be beneficial to create a separate function and provide unit tests, so we know exactly the cases we are covering and have reusable code.

CasLubbers: Fixed

PR #2963, opened 2026-02-20
fix: remove otomi.git required in values-schema

CasLubbers:
## 🧹 Checklist


values-schema.yaml
```diff
@@ -2811,16 +2811,9 @@ properties:
             description: |
               Email address to use for Git commits.
               Defaults to 'pipeline@cluster.local' for internal Gitea.
-            format: email
           branch:
             type: string
             description: The branch to use in the Git repository
-        required:
```
j-zimnowoda, line 2818: is it necessary to remove those?

values-schema.yaml
```diff
@@ -2811,16 +2811,9 @@ properties:
             description: |
```
j-zimnowoda, line 2811: I opened PR https://github.com/linode/apl-core/pull/2964 to this one that adds tests automation.

values-schema.yaml
CasLubbers: Unfortunately it did not work for me if I only removed the git required. I tried multiple times with only Git removed. Also saw that it was not in the generated schema but it kept complaining about these, very odd to me.

PR #2965, opened 2026-02-20
feat: support kubernetes gateway api

merll:
## 📌 Summary

This PR migrates from Ingress-Nginx to the Kubernetes Gateway API using Istio as an ingress gateway.

## 🔍 Reviewer Notes
There are two related PRs, that need to be considered during tests:
* linode/apl-api#943: Required for Shell / TTY feature
* linode/apl-charts#42: Required for canary deployments

The migration process is implemented but requires changes in the following PR to work correctly. This should be merged first: 
* linode/apl-core#3043

For avoiding a race condition between the ConfigMap update and the operator already reconciling, deactivate it temporarily:
1. disable auto-sync of `apl-operator` and scale down to `0`, so that it does not sync while you make the following changes:
```sh
kubectl patch applications.argoproj.io -n argocd apl-operator-apl-operator --patch '[{"op": "remove", "path": "/spec/syncPolicy/automated"}]' --type=json
kubectl scale --replicas=0 -n apl-operator deployment apl-operator
```
2. update ConfigMap `otomi/otomi-status` field `deployingVersion` to `4.16.0`:
```sh
kubectl patch configmap otomi-status -n otomi --type merge -p '{"data":{"deployingVersion": "4.16.0"}}'
```
3. then the version can be set to this feature branch (`APL-1595-2`)
4. scale `apl-operator` back up
```sh
kubectl scale --replicas=1 -n apl-operator deployment apl-operator
```

The following related PR has already been merged to reduce the diff of this one. It includes the CRD and a base chart for generic setup of Gateways:
* linode/apl-core#3030

## 🧹 Checklist

### DIFF
--- a/apps.yaml
+++ b/apps.yaml
@@ -218,7 +218,7 @@ appsInfo:
     integration: App Platform uses Sealed Secrets to provide a secure way to store Kubernetes secrets in Git repositories. Sealed Secrets can be used to store secrets in the values repository.
   trivy:
     title: Trivy Operator
-    appVersion: 0.30.0
+    appVersion: 0.30.1
     repo: https://github.com/aquasecurity/trivy-operator
     maintainers: Aqua Security
     relatedLinks:
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -107,5 +107,5 @@ dependencies:
     version: 1.9.1
     repository: https://cdfoundation.github.io/tekton-helm-chart/
   - name: trivy-operator
-    version: 0.32.0
+    version: 0.32.1
     repository: https://aquasecurity.github.io/helm-charts/
--- a/charts/trivy-operator/Chart.yaml
+++ b/charts/trivy-operator/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 0.30.0
+appVersion: 0.30.1
 description: Keeps security report resources updated
 keywords:
 - aquasecurity
@@ -9,4 +9,4 @@ name: trivy-operator
 sources:
 - https://github.com/aquasecurity/trivy-operator
 type: application
-version: 0.32.0
+version: 0.32.1
--- a/charts/trivy-operator/README.md
+++ b/charts/trivy-operator/README.md
@@ -1,6 +1,6 @@
 # trivy-operator
 
-![Version: 0.32.0](https://img.shields.io/badge/Version-0.32.0-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.30.0](https://img.shields.io/badge/AppVersion-0.30.0-informational?style=flat-square)
+![Version: 0.32.1](https://img.shields.io/badge/Version-0.32.1-informational?style=flat-square) ![Type: application](https://img.shields.io/badge/Type-application-informational?style=flat-square) ![AppVersion: 0.30.1](https://img.shields.io/badge/AppVersion-0.30.1-informational?style=flat-square)
 
 Keeps security report resources updated
 
@@ -153,7 +153,7 @@ Keeps security report resources updated
 | trivy.image.pullPolicy | string | `"IfNotPresent"` | pullPolicy is the imge pull policy used for trivy image , valid values are (Always, Never, IfNotPresent) |
 | trivy.image.registry | string | `"mirror.gcr.io"` | registry of the Trivy image |
 | trivy.image.repository | string | `"aquasec/trivy"` | repository of the Trivy image |
-| trivy.image.tag | string | `"0.69.1"` | tag version of the Trivy image |
+| trivy.image.tag | string | `"0.69.3"` | tag version of the Trivy image |
 | trivy.imageScanCacheDir | string | `"/tmp/trivy/.cache"` | imageScanCacheDir the flag to set custom path for trivy image scan `cache-dir` parameter. Only applicable in image scan mode. |
 | trivy.includeDevDeps | bool | `false` | includeDevDeps include development dependencies in the report (supported: npm, yarn) (default: false) note: this flag is only applicable when trivy.command is set to filesystem |
 | trivy.insecureRegistries | object | `{}` | The registry to which insecure connections are allowed. There can be multiple registries with different keys. |
--- a/charts/trivy-operator/templates/specs/eks-cis-1.4.yaml
+++ b/charts/trivy-operator/templates/specs/eks-cis-1.4.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
-    app.kubernetes.io/version: 0.30.0
+    app.kubernetes.io/version: 0.30.1
     app.kubernetes.io/managed-by: kubectl
 spec:
   cron: {{ .Values.compliance.cron | quote }}
--- a/charts/trivy-operator/templates/specs/k8s-cis-1.23.yaml
+++ b/charts/trivy-operator/templates/specs/k8s-cis-1.23.yaml
@@ -6,7 +6,7 @@ metadata:
   labels:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
-    app.kubernetes.io/version: 0.30.0
+    app.kubernetes.io/version: 0.30.1
     app.kubernetes.io/managed-by: kubectl
 spec:
   cron: {{ .Values.compliance.cron | quote }}
--- a/charts/trivy-operator/templates/specs/k8s-nsa-1.0.yaml
+++ b/charts/trivy-operator/templates/specs/k8s-nsa-1.0.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
-    app.kubernetes.io/version: 0.30.0
+    app.kubernetes.io/version: 0.30.1
     app.kubernetes.io/managed-by: kubectl
 spec:
   cron: {{ .Values.compliance.cron | quote}}
--- a/charts/trivy-operator/templates/specs/k8s-pss-baseline-0.1.yaml
+++ b/charts/trivy-operator/templates/specs/k8s-pss-baseline-0.1.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
-    app.kubernetes.io/version: 0.30.0
+    app.kubernetes.io/version: 0.30.1
     app.kubernetes.io/managed-by: kubectl
 spec:
   cron: {{ .Values.compliance.cron | quote}}
--- a/charts/trivy-operator/templates/specs/k8s-pss-restricted-0.1.yaml
+++ b/charts/trivy-operator/templates/specs/k8s-pss-restricted-0.1.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
-    app.kubernetes.io/version: 0.30.0
+    app.kubernetes.io/version: 0.30.1
     app.kubernetes.io/managed-by: kubectl
 spec:
   cron: {{ .Values.compliance.cron | quote}}
--- a/charts/trivy-operator/templates/specs/rke2-cis-1.24.yaml
+++ b/charts/trivy-operator/templates/specs/rke2-cis-1.24.yaml
@@ -7,7 +7,7 @@ metadata:
   labels:
     app.kubernetes.io/name: trivy-operator
     app.kubernetes.io/instance: trivy-operator
-    app.kubernetes.io/version: 0.30.0
+    app.kubernetes.io/version: 0.30.1
     app.kubernetes.io/managed-by: kubectl
 spec:
   cron: {{ .Values.compliance.cron | quote}}
--- a/charts/trivy-operator/values.yaml
+++ b/charts/trivy-operator/values.yaml
@@ -360,7 +360,7 @@ trivy:
     # -- repository of the Trivy image
     repository: aquasec/trivy
     # -- tag version of the Trivy image
-    tag: 0.69.1
+    tag: 0.69.3
     # -- imagePullSecret is the secret name to be used when pulling trivy image from private registries example : reg-secret
     # It is the user responsibility to create the secret for the private registry in `trivy-operator` namespace
     imagePullSecret: ~


helmfile.d/helmfile-08.init.yaml.gotmpl
```diff
@@ -10,6 +10,8 @@ bases:
 {{ readFile "snippets/templates.gotmpl" }}
 {{- $v := .Values }}
 {{- $a := $v.apps }}
+{{- $domain := $v.cluster.domainSuffix }}
+{{- $hosts := list (print "alertmanager." $domain) (print "grafana." $domain) (print print "prometheus." $domain) }}
```
ferruhcihan: Looks like this `print`s twice ?
```suggestion
{{- $hosts := list (print "alertmanager." $domain) (print "grafana." $domain) (print "prometheus." $domain) }}
```

values/kubernetes-gateways/kubernetes-gateways.gotmpl
```diff
@@ -0,0 +1,156 @@
...
+          namespaces:
+            from: All
+    {{- with $ingress | get "loadBalancerIP" nil }}
+    loadBalancerIP: {{ . | quote }}
+    {{- end }}
+    {{- with $ingress | get "gateway" nil }}
+    replicas: {{ . | get "replicas" 3 }}
+    minAvailable: {{ . | get "minAvailable" 1 }}
+    resources:
+    {{- with $ingress | get "resources" nil }}
```
ferruhcihan: Should be `ingress.gateway.resources`, not `ingress.resources`?
```suggestion
    {{- with . | get "resources" nil }}
```

values/kubernetes-gateways/kubernetes-gateways.gotmpl
```diff
@@ -0,0 +1,156 @@
...
+      {{ . | toYaml | nindent 6 }}
+    {{- else }}
+      requests:
+        cpu: 100m
+        memory: 128Mi
+      limits:
+        cpu: 1000m
+        memory: 1024Mi
+    {{- end }}
+    {{- with $ingress | get "autoscaling" nil }}
```
ferruhcihan: Same here, should be `ingress.gateway.autoscaling`?
```suggestion
    {{- with . | get "autoscaling" nil }}
```

values/external-dns/external-dns.gotmpl
```diff
@@ -2,8 +2,8 @@
 {{- $externalDns := $v.apps | get "external-dns" }}
 {{- $dns := $v | get "dns" dict }}
 sources:
-- ingress
-- istio-gateway
+- gateway-httproute
+- gateway-tlsroute
```
ferruhcihan, line 6: Not sure if this is intentional, should we use the `$useLegacyIngress` flag here like in `apl-network-policies.gotmpl` for the ingress enabled?
```suggestion
- gateway-tlsroute
{{- if $useLegacyIngress }}
- ingress
- istio-gateway
{{- end }}
```

helmfile.d/snippets/defaults.yaml
```diff
@@ -1164,4 +1196,4 @@ environments:
             branch: main
             enabled: true
         versions:
-          specVersion: 55
+          specVersion: 57
```
ferruhcihan, line 1199: The `specVersion` should also be updated in `tests/fixtures/env/settings/versions.yaml` for local development.

values/external-dns/external-dns.gotmpl
merll: This is intentional. As soon as this feature gets rolled out the new resource types (Gateway / HTTPRoute) are created and coexist with the old ones (Ingress / Istio Gateway). However, external-dns should only watch one of them, not both, in order to avoid any conflicting updates.

helmfile.d/snippets/defaults.yaml
merll: Done.

values/knative-serving/knative-serving-cr.gotmpl
```diff
@@ -21,7 +23,11 @@ spec:
       cleanup: "{{ $v.otomi.linodeLkeImageRepository }}/gcr/knative-releases/knative.dev/serving/pkg/cleanup/cmd/cleanup:{{ $tag }}"
       net-istio-controller/controller: "{{ $v.otomi.linodeLkeImageRepository }}/gcr/knative-releases/knative.dev/net-istio/cmd/controller:{{ $netIstioTag }}"
       net-istio-webhook/webhook: "{{ $v.otomi.linodeLkeImageRepository }}/gcr/knative-releases/knative.dev/net-istio/cmd/webhook:{{ $netIstioTag }}"
+      net-gateway-api-controller/controller: "{{ $v.otomi.linodeLkeImageRepository }}/gcr/knative-releases/knative.dev/net-gateway-api/cmd/controller:{{ $netGatewayTag }}"
+      net-gateway-api-webhook/webhook: "{{ $v.otomi.linodeLkeImageRepository }}/gcr/knative-releases/knative.dev/net-gateway-api/cmd/webhook:{{ $netGatewayTag }}"
   {{- end }}
+  additionalManifests:
+    - URL: https://github.com/knative-extensions/net-gateway-api/releases/download/knative-{{ $netGatewayTag }}/net-gateway-api.yaml
```
j-zimnowoda, line 30: Would prefer to keep this manifests in apl-core

values-schema.yaml
```diff
@@ -511,6 +513,20 @@ definitions:
         properties:
           annotations:
             $ref: '#/definitions/annotations'
+      gateway:
+        type: object
+        description: 'Gateway deployment configuration'
```
j-zimnowoda: Local Gateway deployment configuration

values-schema.yaml
```diff
@@ -2027,6 +2043,20 @@ properties:
                 default: 1
                 type: integer
                 minimum: 1
+              gateway:
+                type: object
+                description: 'Gateway deployment configuration'
```
j-zimnowoda: Local Gateway deployment configuration

values/otomi-console/otomi-console.gotmpl
```diff
@@ -3,6 +3,8 @@
 {{- $o := $v.apps | get "otomi-console" }}
 {{- $version := $v.versions | get "console" }}
 {{- $isSemver := regexMatch "^[0-9.]+" $version }}
+{{- $httpRoute := tpl (readFile "../../helmfile.d/snippets/routes.gotmpl") $v | fromYaml }}
+{{- $hostname := print "console." $v.cluster.domainSuffix }}
```
j-zimnowoda: Possible improvement, use:
```
$v._derived.consoleDomain
```

values/otomi-api/otomi-api.gotmpl
```diff
@@ -7,12 +7,13 @@
 {{- $git := $v.otomi.git }}
 {{- $defaultPlatformAdminEmail := printf "platform-admin@%s" $v.cluster.domainSuffix }}
 {{- $sopsEnv := tpl (readFile "../../helmfile.d/snippets/sops-env.gotmpl") $sops }}
+{{- $httpRoute := tpl (readFile "../../helmfile.d/snippets/routes.gotmpl") $v | fromYaml }}
+{{- $hostname := print "api." $v.cluster.domainSuffix }}
```
j-zimnowoda: I would offload this to the `$v._derived.apiDomain`

values/oauth2-proxy/oauth2-proxy.gotmpl
```diff
@@ -25,13 +27,12 @@ config:
   configFile: |-
     # Defaults
     email_domains = [ "*" ]
-    upstreams = [ "file:///dev/null" ]
+    upstreams = [ "static://200" ]
```
j-zimnowoda, line 31: Could you add the following comment:
```
# Upon succesfull token verification, oauth2-proxy returns 200 and K8s gateway lets request go to the real app route.
```

values/oauth2-proxy/oauth2-proxy.gotmpl
```diff
@@ -41,7 +42,6 @@ config:
     cookie_refresh = "1m"
     # set to the same as keycloak client idle timeout
     cookie_expire = "30m"
```
j-zimnowoda, line 45: I understand that we do not use oauth-proxy as reverse proxy anymore. Is that correct?

values/oauth2-proxy/oauth2-proxy.gotmpl
```diff
@@ -25,13 +27,12 @@ config:
   configFile: |-
     # Defaults
     email_domains = [ "*" ]
-    upstreams = [ "file:///dev/null" ]
+    upstreams = [ "static://200" ]
     # Custom
     ssl_insecure_skip_verify = {{ $v._derived.untrustedCA }}
     provider = "keycloak-oidc"
     whitelist_domains = {{ $domains | toJson }}
     scope = {{ $o | get "scope" "openid email profile" | quote }}
```
j-zimnowoda, line 36: Could you explain why this configuration parameter is not needed anymore?

values/kubernetes-gateways/kubernetes-gateways.gotmpl
```diff
@@ -0,0 +1,139 @@
...
+      {{- . | toYaml | nindent 6 }}
+    {{- end }}
+    allowedListeners:
+      namespaces:
+        from: Selector
+        selector:
+          matchLabels:
+            type: team
+    listeners:
+      - name: http
```
j-zimnowoda, line 38: Do we need to support http listener by default?

values/kubernetes-gateways/kubernetes-gateways.gotmpl
```diff
@@ -0,0 +1,139 @@
...
+          namespaces:
+            from: All
+    {{- with $ingress | get "loadBalancerIP" nil }}
+    loadBalancerIP: {{ . | quote }}
+    {{- end }}
+    {{- with $ingress | get "gateway" dict }}
+    replicas: {{ . | get "replicas" 3 }}
+    minAvailable: {{ . | get "minAvailable" 1 }}
+    resources:
+    {{- with . | get "resources" nil }}
```
j-zimnowoda: This is nested `with` statement and it may be easy to get lost with following what is `.` 
In such case, could you use named variable instead of "." ?

values/kubernetes-gateways/kubernetes-gateways.gotmpl
```diff
@@ -0,0 +1,139 @@
...
+      {{- . | toYaml | nindent 6 }}
+    {{- else }}
```
j-zimnowoda, line 64: Ingress is a dynamic objects thus it is impossible to set defautls in the defaults.yaml. However the `defaults.gotmpl` alow to render dynamically defaults. Having that you could simplify this template by not needing to define fallbacks.

values/kubernetes-gateways/kubernetes-gateways.gotmpl
merll: No, we do not have to. However, we used to have a HTTP to HTTPS redirect in all Ingress instances, and this is the method to match this functionality.

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
```diff
@@ -0,0 +1,118 @@
...
+      action: CUSTOM
+      provider:
+        name: oauth2-proxy
+      targetRefs:
+        - name: {{ $v.ingress.platformClass.className }}
+          kind: Gateway
+          group: gateway.networking.k8s.io
+      rules:
+        - when:
+            - key: request.headers[authorization]
```
j-zimnowoda: Could you add a comment:
```
# Match HTTP requests wither without the Authorization Header or with but empty value.
```

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
```diff
@@ -0,0 +1,118 @@
+{{- $v := .Values }}
+{{- $a := $v.apps }}
+{{- $authApps := list "console" "tty" }}
+{{- $authAppNamespaces := list "argocd" "harbor" "monitoring" "grafana" "kfp" "otomi" "tekton-pipelines"}}
```
j-zimnowoda, line 3: how about keycloak?

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
```diff
@@ -0,0 +1,118 @@
...
+  - apiVersion: security.istio.io/v1
+    kind: AuthorizationPolicy
+    metadata:
+      name: {{ $ingress.className }}-allowed-ips
+    spec:
+      targetRef:
+        kind: Gateway
+        group: gateway.networking.k8s.io
+        name: {{ $ingress.className }}
+      action: ALLOW
```
j-zimnowoda: Does it implicitly mean that others will be disallowed ?
Shall we create disallowAll fallback in case `sourceIpAddressFiltering` is provided?

values/istiod/istiod.gotmpl
```diff
@@ -49,8 +53,33 @@ meshConfig:
...
+        headersToDownstreamOnAllow:
+          - set-cookie
+        headersToUpstreamOnAllow:
+          - authorization
+          - path
+          - x-auth-request-email
+          - x-auth-request-groups
+          - x-auth-request-user
+        includeRequestHeadersInCheck:
+          - authorization
```
j-zimnowoda, line 69: If the authorization header is present. Is it necessay to forward it to oauth2-proxy. In particular how does it play with the `platform-ouath2-proxy` AuthorizationPolicy?

helmfile.d/snippets/authpolicy.gotmpl
```diff
@@ -0,0 +1,32 @@
...
+{{- end }}
+apiVersion: security.istio.io/v1
+kind: AuthorizationPolicy
+metadata:
+  name: {{ $v.prefix }}-oauth2-proxy
+  namespace: istio-system
+spec:
+  action: CUSTOM
+  provider:
+    name: oauth2-proxy
```
j-zimnowoda: Could you please rename it to sth that is easy to find, e.g.: `oauth2-proxy-ext-authz`

values/istiod/istiod.gotmpl
```diff
@@ -49,8 +53,33 @@ meshConfig:
...
+          - cookie
+        service: oauth2-proxy.istio-system.svc.cluster.local
+        port: 80
+      name: oauth2-proxy
```
j-zimnowoda: Could you rename it to `oauth2-proxy-ext-authz`

helmfile.d/helmfile-60.teams.yaml.gotmpl
```diff
@@ -118,6 +169,9 @@ releases:
               basicAuthUser: {{ $teamId }}
               secureJsonData:
                 basicAuthPassword: {{ $teamSettings.password | quote }}
+        extraManifests:
+          - {{ tpl (readFile "../helmfile.d/snippets/authpolicy.gotmpl") (dict "prefix" (print "po-" $teamId) "gatewayName" $gatewayName "hosts" (list $alertmanagerHostname $grafanaHostname)) | nindent 12 }}
```
j-zimnowoda: Instead of `po-` we could use `monitoring`

helmfile.d/helmfile-08.init.yaml.gotmpl
```diff
@@ -28,5 +30,7 @@ releases:
       - ../values/prometheus-operator/prometheus-operator.gotmpl
       - {{- $a | get "prometheus._rawValues" dict | toYaml | nindent 8 }}
       - alertmanager: {{- $a | get "alertmanager._rawValues" dict | toYaml | nindent 10 }}
-      - grafana: {{- $a | get "grafana._rawValues" dict | toYaml | nindent 10 }}
-
+        grafana: {{- $a | get "grafana._rawValues" dict | toYaml | nindent 10 }}
+        extraManifests:
+          - {{ tpl (readFile "snippets/authpolicy.gotmpl") (dict "prefix" "po" "gatewayName" $v.ingress.platformClass.className "hosts" $hosts) | nindent 12 }}
```
j-zimnowoda: instead of `po` we could use `monitoring`

values/kubernetes-gateways/kubernetes-gateways.gotmpl
j-zimnowoda: okay, if the new version also contains the redirect  HTTP to HTTPS  then let's keep it.

values/knative-serving/knative-serving-cr.gotmpl
merll: It is not technically hard to download this file and place it in Core. But on the following upgrade this is almost certain to be forgotten. The risk is already high enough with the variables above in this file, since the versions are often not identical with the Knative Operator version nor each other. I would prefer finding a proper solution in a different implementation that also considers the maintenance of all these image references.

values-schema.yaml
merll: The description is updated.

values-schema.yaml
merll: The description is updated.

values/oauth2-proxy/oauth2-proxy.gotmpl
merll: Yes. OAuth2-Proxy is contacted by Istio to verify the authentication, and if it fails redirects to the OAuth2-Proxy frontend, which redirects further to Keycloak and back.

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
merll: These are the apps that need access to the OAuth2-Proxy service through a `ReferenceGrant`. Essentailly that is all apps from `core.yaml` that have `ingress.auth=true` Keycloak does not communicate with the service directly, but instead passes information through a redirect.

values/knative-serving/knative-serving-cr.gotmpl
j-zimnowoda: I am more concerned about data integrity. If we fetch some manifests from the outside the cluster I would like to make sure they are always the same.

We could also enhance our CI scripts to track those manifests.

values/oauth2-proxy/oauth2-proxy.gotmpl
j-zimnowoda: Thanks for the confirmation.

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
j-zimnowoda: I understand and should be more specific. I was thinking about keycloak admin console. Can use still login with SSO there?

values/oauth2-proxy/oauth2-proxy.gotmpl
merll: Because the `auth` host is no longer used for the callback. Instead the redirect now goes through the `/oauth2` routes of each host where authentication is needed. Implementing it in the same way as before was not possible, the redirect across different hosts does not work with the extension.

j-zimnowoda: That makes perfect sense now. Thanks.

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
j-zimnowoda: @merll you have reminded me that cloud manger relies on `https://auth.lke${cluster.id}.akamai-apl.net/ready` 
Is the auth host still present?

helmfile.d/helmfile-60.teams.yaml.gotmpl
j-zimnowoda: TY

helmfile.d/helmfile-08.init.yaml.gotmpl
j-zimnowoda: TY

values/istiod/istiod.gotmpl
j-zimnowoda: TY

helmfile.d/snippets/authpolicy.gotmpl
j-zimnowoda: TY

values/oauth2-proxy/oauth2-proxy.gotmpl
j-zimnowoda: The cloud manager check platform readiness by sending HTTP GET request to  `https://auth.lke${cluster.id}.akamai-apl.net/ready`

Is the `auth` host still available ?

values-schema.yaml
j-zimnowoda: TY

values-schema.yaml
j-zimnowoda: TY

values/istiod/istiod.gotmpl
merll: I am not sure if it is necessary, but I found it in two instances of configuration examples. Both of them include them:
https://istio.io/latest/docs/tasks/security/authorization/authz-custom/#define-the-external-authorizer
https://github.com/kubeflow/manifests/blob/master/common/oauth2-proxy/components/istio-external-auth-patches/patches/cm.enable-oauth2-proxy.yaml

Testing with and without consistently is tedious because changes do not always propagate immediately.

values/knative-serving/knative-serving-cr.gotmpl
merll: Yes, we should, just like any other artifact (binaries, image tags) that are not directly included in the Helm charts.

In this particular case we are fetching a release artifact, which is not more likely to change than e.g. an image tag. Also I am not sure if adding this extension is still necessary, since I originally tested with an older version.

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
merll: I am not sure but I think Keycloak does not need the SSO, since it does not rely on OAuth2-Proxy but carries the session info itself. That part has not changed at all.

And yes, I was aware of the auth-domain endpoint. It is set up by changes here:
https://github.com/linode/apl-core/pull/2965/changes#diff-52f33fc0dad82ac7f18d3fa653fd3b2f48d94cd97839ffe2eb9da65f55be6c58R176

values/oauth2-proxy/oauth2-proxy.gotmpl
merll: Yes, see line 176 forward below.

values/kubernetes-gateways/kubernetes-gateways.gotmpl
merll: This is true for the `platformClass`which is in a value structure on its own. However other classes are list elements in `classes`, identified by a unique `className`. I am not sure how well the merge would work through the templates.

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
merll: In fact it was not implicitly disallowing them. A default DENY seems a bit drastic, but I think a reverse of the rule (DENY, and then exclude certain blocks) as I now implemented it should work.

values/kubernetes-gateways/kubernetes-gateways.gotmpl
merll: This is updated now.

merll:
> What would be procedure for allowing arbitrary NS to use platform SSO?

I can answer that question more clearly once I have solved the issue with token verification. It will likely involve adding an AuthorizationPolicy or modifying a common one, if we integrate it into the team-ns chart.

values/otomi-api/otomi-api.gotmpl
```diff
@@ -71,12 +71,26 @@ core:
...
+  {{- end }}
+  hostname: {{ $v._derived.apiDomain }}
+  parentRefs:
+    {{- $httpRoute.parentRefs | toYaml | nindent 4 }}
+  additionalRules:
+    {{- $httpRoute.authRules | toYaml | nindent 4 }}
+
+extraManifests:
+  - {{ tpl (readFile "../../helmfile.d/snippets/authpolicy.gotmpl") (dict "prefix" "tty" "gatewayName" $v.ingress.platformClass.className "host" $v._derived.ttyDomain) | nindent 4 }}
+  - {{ tpl (readFile "../../helmfile.d/snippets/serviceentry.gotmpl") (dict "name" "otomi-api" "host" $v._derived.apiDomain) | nindent 4 }}
```
ferruhcihan: The TTY domain has an AuthorizationPolicy, but no ServiceEntry (it only exists for apiDomain). Is it also necessary to add a ServiceEntry for TTY, or is this intentional?

values/tekton-dashboard/tekton-dashboard-raw.gotmpl
```diff
@@ -0,0 +1,35 @@
...
+            - type: RequestHeaderModifier
+              requestHeaderModifier:
+                remove:
+                  - authorization
+          matches:
+            - path:
+                type: PathPrefix
+                value: /
+        {{- $httpRoute.authRules | toYaml | nindent 8 }}
+  - {{ tpl (readFile "../../helmfile.d/snippets/authpolicy.gotmpl") (dict "prefix" "tekton" "gatewayName" $v.ingress.platformClass.className "host" $hostname) | nindent 4 }}
```
ferruhcihan, line 35: In the `helmfile.d/helmfile-60.teams.yaml.gotmpl`, there are both an AuthPolicy and a ServiceEntry for team-level Tekton, but it seems there's no ServiceEntry for platform-level Tekton. Is it necessary to add a ServiceEntry here?

merll:
> It appears that TTY might not need a ServiceEntry, whereas the Tekton Dashboard could require one. It would be great to confirm this, so we can make sure the configuration is correct for both.

Thanks for pointing out the missing entries. The ServiceEntry is only necessary for platform-internal traffic to find its way when using external host names. Missing some of them would likely go unnoticed in most cases, but for consistency I think it is better to have them in all cases. Therefore I added them.

values/tekton-dashboard/tekton-dashboard-raw.gotmpl
merll: This is added now.

values/otomi-api/otomi-api.gotmpl
merll: This is now extended to `tty` as well.

values/knative-serving/knative-serving-cr.gotmpl
j-zimnowoda: I will add it to the enhancement list

values/istiod/istiod.gotmpl
j-zimnowoda: I added it to the enhancement list

values/kubernetes-gateways/kubernetes-gateways.gotmpl
j-zimnowoda: I added it to the enhancement list

PR #2975, opened 2026-02-23
feat: added LB annotations to preserve IP prior to the Gateway API upgrade

Ani1357:
## 🧹 Checklist

### DIFF
--- a/src/cmd/migrate.ts
+++ b/src/cmd/migrate.ts
@@ -797,7 +797,11 @@ const addLinodeNBAnnotations = async (values: Record<string, any>): Promise<void
 
   d.info(`Found NodeBalancer ID ${nbId}, adding annotations to ingress.platformClass`)
   const annotations: Array<{ key: string; value: string }> = get(values, 'ingress.platformClass.annotations', [])
+  const preserveKey = 'service.beta.kubernetes.io/linode-loadbalancer-preserve'
   const nbIdKey = 'service.beta.kubernetes.io/linode-loadbalancer-nodebalancer-id'
+  if (!annotations.find((a) => a.key === preserveKey)) {
+    annotations.push({ key: preserveKey, value: 'true' })
+  }
   if (!annotations.find((a) => a.key === nbIdKey)) {
     annotations.push({ key: nbIdKey, value: String(nbId) })
   }


src/common/runtime-upgrades/runtime-upgrades.ts
```diff
@@ -167,4 +168,10 @@ export const runtimeUpgrades: RuntimeUpgrades = [
       },
     },
   },
+  {
+    version: '4.16.0',
+    post: async (context: RuntimeUpgradeContext) => {
+      await syncIngressNginxPlatform(context)
```
j-zimnowoda, line 169: should be called only if provider is `linode`

src/cmd/migrate.ts
```diff
@@ -735,6 +735,59 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
...
+  if (!linodeSecret?.token) {
+    d.warn('Linode secret not found or token missing in kube-system, skipping Linode NodeBalancer annotation migration')
+    return
+  }
+  const token = linodeSecret.token as string
+
+  d.info('Querying Linode API for NodeBalancer ID')
+  let nbId: number | undefined
+  try {
+    const response = await fetch('https://api.linode.com/v4/nodebalancers', {
```
CasLubbers, line 776: General question is this not the same external ip-adress that is set on the `ingress-nginx-platform-controller` service? If so then we don't have to fetch it from Linode api.
We also do this in apl-installer and the github deploy actions.

src/cmd/migrate.ts
```diff
@@ -735,6 +735,59 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
...
+        Authorization: `Bearer ${token}`,
+        'Content-Type': 'application/json',
+      },
+    })
+    if (!response.ok) {
+      d.warn(`Failed to fetch NodeBalancers from Linode API: ${response.status} ${response.statusText}`)
+      return
+    }
+    const data = (await response.json()) as { data: Array<{ id: number }> }
+    const nb = data.data?.[0]
```
CasLubbers: Is it correct to always take the first argument? In our shared dev account this would not work then is my guess

src/cmd/migrate.ts
```diff
@@ -735,6 +735,59 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
   set(values, 'catalogs.default', defaultCatalog)
 }
 
+const addLinodeNBAnnotations = async (values: Record<string, any>): Promise<void> => {
+  const d = terminal('addLinodeNBAnnotations')
+  if (env.DISABLE_SYNC) {
```
CasLubbers, line 744: This can be skipped if provider is not Linode

src/cmd/migrate.ts
merll: We are not looking for the IP address, but the id of the NodeBalancer. And in case we have multiple IngressClass instances configured, we should use the IP (or hostname) to unambiguously match this with the Linode API output (below).

src/cmd/migrate.ts
merll: This would be correct as long as there is only one NodeBalancer assigned to the cluster, because the token stored in the cluster only has access to these. However, setting up multiple IngressClasses would result in multiple NodeBalancers being created, and in that case this would not work reliably.

src/common/runtime-upgrades/add-linode-nb-annotations.ts
```diff
@@ -0,0 +1,50 @@
+import retry from 'async-retry'
+import { env } from '../envalid'
+import { hf, HF_DEFAULT_SYNC_ARGS } from '../hf'
+import { k8s } from '../k8s'
+import { logLevelString, terminal } from '../debug'
+import { rootDir } from '../utils'
+import { RuntimeUpgradeContext } from './runtime-upgrades'
+
+const PLATFORM_LB_SERVICE = 'ingress-nginx-platform-controller'
```
merll, line 9: There might be more than one. `platform` is the default.

src/cmd/migrate.ts
```diff
@@ -735,6 +735,59 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
...
+  const annotations: Array<{ key: string; value: string }> = get(values, 'ingress.platformClass.annotations', [])
+  const preserveKey = 'service.beta.kubernetes.io/linode-loadbalancer-preserve'
+  const nbIdKey = 'service.beta.kubernetes.io/linode-loadbalancer-nodebalancer-id'
+  if (!annotations.find((a) => a.key === preserveKey)) {
+    annotations.push({ key: preserveKey, value: 'true' })
+  }
+  if (!annotations.find((a) => a.key === nbIdKey)) {
+    annotations.push({ key: nbIdKey, value: String(nbId) })
+  }
+  set(values, 'ingress.platformClass.annotations', annotations)
```
merll, line 808: ~~I would prefer to call this `serviceAnnotations`, as these are both in ingress-nginx and the Gateway implementation two separate sets of values from the main resource (Ingress or Gateway).~~

nvm we currently render `annotations` for the service. Better to leave it this way then.

merll: Note also that we currently store this in the schema as a list of objects. We should attempt to stay consistent about this.

src/cmd/migrate.ts
merll: Provisioned instances do not provide for creating additional ingress classes. However, technically users could still provision another service of type `LoadBalancer`. So to simplify things, we could
1. if there is only one entry, use that one
2. if there are multiple, match by IP of the LoadBalancer service status field

Ani1357: update code to get the id of the nodebalancer that matches the ip of the `ingress-nginx-platform-controller` service

src/cmd/migrate.ts
```diff
@@ -735,6 +735,81 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
   set(values, 'catalogs.default', defaultCatalog)
 }
 
+const addLinodeNBAnnotations = async (values: Record<string, any>): Promise<void> => {
+  const d = terminal('addLinodeNBAnnotations')
+  const cluster = await loadYaml(`${env.ENV_DIR}/env/settings/cluster.yaml`, { noError: true })
```
j-zimnowoda: since the values are provided as function argument, why do you need to load the cluster.yaml file independently?

Ani1357: not needed, removed.

src/cmd/migrate.ts
```diff
@@ -735,6 +735,80 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
...
+    }
+    nbId = nb.id
+  } catch (error) {
+    d.warn(`Error querying Linode API: ${error}`)
+    return
+  }
+
+  d.info(`Found NodeBalancer ID ${nbId}, adding annotations to ingress.platformClass`)
+  const annotations: Array<{ key: string; value: string }> = get(values, 'ingress.platformClass.annotations', [])
+  const preserveKey = 'service.beta.kubernetes.io/linode-loadbalancer-preserve'
```
merll: I just noticed that I included this one in the templates. There is no need to include it in the values store since it is static. But we can also deduplicate this part later.

merll:
We had been discussing the migration updates not being stored until the end of the reconciliation. This could lead to the NodeBalancer id getting lost of something does not work out as expected. Can we commit changes right after the migration?

src/cmd/migrate.ts
Ani1357: removed

j-zimnowoda:
> We had been discussing the migration updates not being stored until the end of the reconciliation. This could lead to the NodeBalancer id getting lost of something does not work out as expected. Can we commit changes right after the migration?

@merll, I have checked the code and I see that commit is also performed just after migration: https://github.com/linode/apl-core/blob/main/src/operator/apl-operator.ts#L94
Is that sufficient?

src/cmd/migrate.ts
j-zimnowoda: IMHO we should ensure this annotation in LB service before any changes are applied

merll:
> > We had been discussing the migration updates not being stored until the end of the reconciliation. This could lead to the NodeBalancer id getting lost of something does not work out as expected. Can we commit changes right after the migration?
> 
> @merll, I have checked the code and I see that commit is also performed just after migration: https://github.com/linode/apl-core/blob/main/src/operator/apl-operator.ts#L94 Is that sufficient?

Yes, it should be.

src/cmd/migrate.ts
Ani1357: reverted the removal

PR #2985, opened 2026-02-25
feat: support subdirectories for byo catalogs

ElderMatt:
## 📌 Summary

Adds chartsPath to the values-schema


versions.yaml
```diff
@@ -1,5 +1,5 @@
-api: main
-console: main
-consoleLogin: main
+api: APL-1564
```
ElderMatt: Change this back after approvals

j-zimnowoda (approved):

PR #3000, opened 2026-03-02
feat: added apl-operator livenessProbe

Ani1357:
## 🧹 Checklist

### DIFF
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -6,7 +6,7 @@ version: 0.1.0
 dependencies:
   - name: argo-cd
     alias: argocd
-    version: 9.4.6
+    version: 9.4.7
     repository: https://argoproj.github.io/argo-helm
   - name: argocd-image-updater
     version: 1.0.1
--- a/charts/argocd/Chart.yaml
+++ b/charts/argocd/Chart.yaml
@@ -1,7 +1,9 @@
 annotations:
   artifacthub.io/changes: |
+    - kind: fixed
+      description: Align timeout.reconciliation default with upstream (180s -> 120s)
     - kind: added
-      description: Added support for custom statefulset labels
+      description: Add timeout.reconciliation.jitter config default (60s)
   artifacthub.io/signKey: |
     fingerprint: 2B8F22F57260EFA67BE1C5824B11F800CD9D2252
     url: https://argoproj.github.io/argo-helm/pgp_keys.asc
@@ -28,4 +30,4 @@ name: argo-cd
 sources:
 - https://github.com/argoproj/argo-helm/tree/main/charts/argo-cd
 - https://github.com/argoproj/argo-cd
-version: 9.4.6
+version: 9.4.7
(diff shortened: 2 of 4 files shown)


charts/apl-operator/templates/deployment.yaml
```diff
@@ -56,6 +57,30 @@ spec:
             - name: git-config
               mountPath: /home/app/.gitconfig
               subPath: .gitconfig
+        - name: watchdog
```
j-zimnowoda: Why sidecar is better than implementing the liveness probe in the apl-operator?

Ani1357: IMO the sidecar is better isolated from eventual pod failures but maybe it is to much for now. 
Updated code to use a liveness probe.

charts/apl-operator/templates/deployment.yaml
```diff
@@ -46,6 +46,16 @@ spec:
                 name: apl-sops-secrets
             - secretRef:
                 name: apl-git-credentials
+          livenessProbe:
```
j-zimnowoda, line 49: This is much better 👍

src/operator/k8s.ts
```diff
@@ -72,6 +73,7 @@ export async function updateApplyState(
     }
 
     d.info(`Apply state updated for commit ${state.commitHash}`)
+    writeFileSync('/tmp/heartbeat', '')
```
j-zimnowoda: Could you extract this to a separate function with proper description, so SWE immidiately knows its purpose.
```
/**
* This function updates a file and its timestamp is used for liveness probes.
*/
updateHeartBeatFile(){

}
```

j-zimnowoda (approved):

src/operator/k8s.ts
```diff
@@ -29,6 +30,15 @@ export const k8s = {
   },
 }
 
+/**
+ * Writes an empty file to /tmp/heartbeat to update its modification timestamp.
```
j-zimnowoda, line 34: How about the initial installation phase?
Will the liveness probes and lack of /tmp/heartbeat file  cause any problems?

Ani1357: no, because of the `initialDelaySeconds: 180`

CasLubbers: I think its a good idea to also call the function during the installation phase. Because the installation can take some time and it would be really bad if it restarts because of it. Maybe in the while loop from the reconcile installation?

CasLubbers: Now that I think of it we are only adding the livenessprobe in charts/ and not in chart/. So the livenessprobe will only be applied after the argocd application is applied. Then the installation was successful so then it would not be needed. But we have a mismatch between charts/ and chart/. Is that something we accept?

j-zimnowoda: I think that we should unify chart/apl with charts/apl-operator but **not** as a part of this PR. The proposed solution by @Ani1357 is good enough. Thank you.

PR #3009, opened 2026-03-05
fix: hardcoded main branch in operator

CasLubbers:
## 🧹 Checklist


src/cmd/pull.ts
```diff
@@ -12,7 +12,7 @@ const cmdName = getFilename(__filename)
 export const pull = async (): Promise<void> => {
   const d = terminal(`cmd:${cmdName}:pull`)
   const allValues = await hfValues()
-  const branch = allValues?.apps?.['otomi-api']?.git?.branch ?? 'main'
+  const branch = allValues?.otomi?.git?.branch ?? 'main'
```
j-zimnowoda: I would use getRepo() so we constantly rely on GitRepoConfig interface

PR #3012, opened 2026-03-05
feat: increase lieveness probe failureThreshold for apl-operator

j-zimnowoda:
## 📌 Summary
The failure threshold is a bit too low to decide if a pod should be killed. 

## 🧹 Checklist


CasLubbers (approved):

merll (approved):

Ani1357 (commented):
With this change, in case the operator gets stuck, the earlier restart would happen after 18min(15min + 3 X 60s X failureThreshold).
How about lowering the test to check so it tests after 10 min (600s) and keep the failureThreshold at 3, so it takes 13min in the best case. I think after 13min is still reasonable to consider the operator stuck.

j-zimnowoda:
> 18min(15min + 3 X 60s X failureThreshold).

@Ani1357 From where the 15 minutes coming ?

PR #3029, opened 2026-03-10
fix: setting branch in argocd app

CasLubbers:
## 🧹 Checklist


values/team-ns/team-ns.gotmpl
```diff
@@ -66,6 +66,7 @@ defaultIngressGatewayLabel: {{ $v._derived.defaultIngressGatewayLabel }}
 canaryIngressGatewayLabel: {{ $v._derived.canaryIngressGatewayLabel | quote }}
 
 gitOps:
+  branch: {{ $v.otomi.git.branch | default "main" }}
```
j-zimnowoda: defaults should be set in defaults.yaml

PR #3057, opened 2026-03-18
fix: add upper bound for runtime upgrade

ferruhcihan:
## 📌 Summary

This PR adds upper bound for the runtime upgrade function.

## 🧹 Checklist

### DIFF
--- a/src/common/runtime-upgrade.ts
+++ b/src/common/runtime-upgrade.ts
@@ -95,7 +95,7 @@ export function filterRuntimeUpgrades(
   }
   const targetVersion = deployingVersion ? semver.coerce(deployingVersion) : null
   return rUpgrades.filter((rUpgrade) => {
-    if (!semver.gt(rUpgrade.version, currentVersion)) return false
+    if (semver.lte(rUpgrade.version, currentVersion)) return false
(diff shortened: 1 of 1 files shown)


src/common/runtime-upgrade.ts
```diff
@@ -84,10 +84,19 @@ export async function runtimeUpgrade({ when, deploymentState }: RuntimeUpgradeAr
...
+  deployingVersion?: string,
+): RuntimeUpgrades {
   const currentVersion = semver.coerce(version)
   if (!currentVersion) {
     throw new Error(`Unsupported version format: ${version}`)
   }
-  return rUpgrades.filter((rUpgrade) => semver.gt(rUpgrade.version, currentVersion))
+  const targetVersion = deployingVersion ? semver.coerce(deployingVersion) : null
+  return rUpgrades.filter((rUpgrade) => {
+    if (!semver.gt(rUpgrade.version, currentVersion)) return false
```
merll: The negation makes this harder to follow. I would prefer

```suggestion
    if (semver.lte(rUpgrade.version, currentVersion)) return false
```

merll (approved):

PR #3058, opened 2026-03-18
fix: apps not using desired storageclass in linode

ElderMatt:
This PR fixes the storage class being used for gitea-valkey and oauth2-proxy, it is now set to linode-block-storage instead of linode-block-storage-retain. Migration has also been added to delete and recreate the pvcs.

### DIFF
--- a/helmfile.d/snippets/defaults.yaml
+++ b/helmfile.d/snippets/defaults.yaml
@@ -1190,4 +1190,4 @@ environments:
           aiEnabled: false
         users: []
         versions:
-          specVersion: 58
+          specVersion: 59
--- a/tests/fixtures/env/settings/versions.yaml
+++ b/tests/fixtures/env/settings/versions.yaml
@@ -3,4 +3,4 @@ metadata:
     name: versions
     labels: {}
 spec:
-    specVersion: 58
+    specVersion: 59
--- a/values-changes.yaml
+++ b/values-changes.yaml
@@ -453,7 +453,8 @@ changes:
     customFunctions:
       - setIngressDefault
   - version: 58
-    customFunctions:
-      - valkeyAndOauth2RedisPVCMigration
     additions:
       - apps.tekton.enabled: true
+  - version: 59
+    customFunctions:
+      - valkeyAndOauth2RedisPVCMigration


src/cmd/migrate.ts
```diff
@@ -831,6 +1001,7 @@ const customMigrationFunctions: Record<string, CustomMigrationFunction> = {
   workloadValuesMigration,
   setLokiStorageSchemaMigration,
   setDefaultAplCatalog,
+  ValkeyAndOauth2RedisPVCMigration,
```
merll: Since this is a function and not a class, please stick to the naming convention regarding capitalization.

src/cmd/migrate.ts
```diff
@@ -520,6 +520,125 @@ async function appExists(name: string): Promise<boolean> {
...
+    throw error
+  }
+}
+
+function isK8sTlsHttpError(error: unknown): boolean {
+  const message = error instanceof Error ? error.message : String(error)
+  return message.includes('HTTP protocol is not allowed when skipTLSVerify is not set or false')
+}
+
+const sleep = async (ms: number): Promise<void> => new Promise((resolve) => setTimeout(resolve, ms))
```
merll: We have at least one implementation in a library (e.g. zx) that is ready to use.

src/cmd/migrate.ts
```diff
@@ -520,6 +520,125 @@ async function appExists(name: string): Promise<boolean> {
...
+      return false
+    }
+    if (isK8sTlsHttpError(error)) {
+      return false
+    }
+    throw error
+  }
+}
+
+function isK8sTlsHttpError(error: unknown): boolean {
```
merll: Why is this function needed? Are we doing something similar in other migrations, or should we?

merll:
I am fine with the main functionality of this branch, but it contains lots of unrelated changes which make it harder to read and later follow up in case someone is looking into old code. I also doubt such changes can be properly QA'ed when they are not used in any recent environment.

I would suggest either
* reverting the unrelated changes
* or removing old migrations entirely. The Istio migration is from 9 months ago and if someone still needs to upgrade a cluster this old, it should be done in smaller increments.

src/cmd/migrate.ts
ElderMatt: It was to get around a build error, but just saw we can skip it as well so implemented that instead.

ElderMatt:
We did not set the storageClass for gitea-valkey which is why it was defaulting to `linode-block-storage-retain`. For Oauth2 I'm not quite sure.

ElderMatt:
It seems for older cluster before spec version 44 we never set the defaultStorageClass so it is ''. In specVersion 44 we added the defaultStorageClass but there was no migration to set the value to `linode-block-storage`

PR #3097, opened 2026-03-31
feat: ensure manifest folder in values repo

ElderMatt:
This PR ensure that the manifests folders are always present in the values with a .gitkeep file.

image

### DIFF
--- a/apps.yaml
+++ b/apps.yaml
@@ -230,7 +230,7 @@ appsInfo:
     chartName: trivy-operator
   otel:
     title: OpenTelemetry Operator
-    appVersion: 0.144.0
+    appVersion: 0.148.0
     repo: https://github.com/open-telemetry/opentelemetry-operator
     maintainers: Grafana labs
     relatedLinks:
--- a/chart/chart-index/Chart.yaml
+++ b/chart/chart-index/Chart.yaml
@@ -81,7 +81,7 @@ dependencies:
     repository: https://oauth2-proxy.github.io/manifests
   - name: opentelemetry-operator
     alias: otel-operator
-    version: 0.105.1
+    version: 0.109.0
     repository: https://open-telemetry.github.io/opentelemetry-helm-charts
   - name: policy-reporter
     version: 3.7.3
--- a/charts/otel-operator/Chart.yaml
+++ b/charts/otel-operator/Chart.yaml
@@ -1,5 +1,5 @@
 apiVersion: v2
-appVersion: 0.144.0
+appVersion: 0.148.0
 description: OpenTelemetry Operator Helm chart for Kubernetes
 home: https://opentelemetry.io/
 icon: https://raw.githubusercontent.com/cncf/artwork/a718fa97fffec1b9fd14147682e9e3ac0c8817cb/projects/opentelemetry/icon/color/opentelemetry-icon-color.png
@@ -12,4 +12,4 @@ name: opentelemetry-operator
 sources:
 - https://github.com/open-telemetry/opentelemetry-operator
 type: application
-version: 0.105.1
+version: 0.109.0
--- a/charts/otel-operator/crds/crd-opentelemetry.io_opampbridges.yaml
+++ b/charts/otel-operator/crds/crd-opentelemetry.io_opampbridges.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.20.1
   creationTimestamp: null
   labels:
     app.kubernetes.io/name: opentelemetry-operator
--- a/charts/otel-operator/crds/crd-opentelemetry.io_targetallocators.yaml
+++ b/charts/otel-operator/crds/crd-opentelemetry.io_targetallocators.yaml
@@ -4,7 +4,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.20.0
+    controller-gen.kubebuilder.io/version: v0.20.1
   creationTimestamp: null
   labels:
     app.kubernetes.io/name: opentelemetry-operator
@@ -1339,6 +1339,8 @@ spec:
                 type: boolean
               hostPID:
                 type: boolean
+              hostUsers:
+                type: boolean
               image:
                 type: string
               imagePullPolicy:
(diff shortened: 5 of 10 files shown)


CasLubbers (changes requested):
The implementation is not wrong, but we already have a similar function for this: `ensureTeamGitOpsDirectories` https://github.com/linode/apl-core/blob/main/src/common/utils.ts#L218
We can consolidate the logic and re-use it for the manifests folder. I think it would also be good to call it in the same place.

src/common/utils.ts
```diff
@@ -215,10 +217,27 @@ async function ensureKeepFile(keepFilePath: string, deps = { writeFile }): Promi
   await deps.writeFile(keepFilePath, '')
 }
 
+export function ensureManifestDirectories(): void {
+  ;[operatorEnv.GITOPS_MANIFESTS_NS_PATH, operatorEnv.GITOPS_MANIFESTS_GLOBAL_PATH].forEach((p) =>
```
j-zimnowoda: why looping here?
just call twice the ensureDirectoryWithGitkeepAsync()

src/operator/validators.ts
```diff
@@ -6,6 +6,14 @@ dotenv.config()
 export const operatorEnv = cleanEnv(process.env, {
   GIT_ORG: str({ desc: 'Git organisation', default: 'otomi' }),
   GIT_REPO: str({ desc: 'Git repository', default: 'values' }),
+  GITOPS_MANIFESTS_NS_PATH: str({
```
j-zimnowoda: GITOPS_NS_MANIFESTS_RELATIVE_PATH

src/operator/validators.ts
```diff
@@ -6,6 +6,14 @@ dotenv.config()
...
+    desc: 'Path to the gitops manifests namespace',
+    default: 'env/manifests/namespaces',
+  }),
+  GITOPS_MANIFESTS_GLOBAL_PATH: str({
```
j-zimnowoda: GITOPS_GLOBAL_MANIFESTS_RELATIVE_PATH

src/common/utils.ts
```diff
@@ -215,10 +217,27 @@ async function ensureKeepFile(keepFilePath: string, deps = { writeFile }): Promi
...
+    ensureDirectoryWithGitkeepAsync(pathModule.join(env.ENV_DIR, p)),
+  )
+}
+
+async function ensureDirectoryWithGitkeepAsync(dirPath: string, deps = { fs, path: pathModule }) {
+  await deps.fs.promises.mkdir(dirPath, { recursive: true })
```
j-zimnowoda: we usually add functions to deps if they require filesystem or network access. The `path.join` does not require it.

please be explicit in adding deps instead of adding the whole fs module you could simply add `access` , `writeFile` and `mkdir`

src/common/utils.ts
```diff
@@ -1,14 +1,16 @@
...
 import cleanDeep, { CleanOptions } from 'clean-deep'
 import { createHash } from 'crypto'
-import { existsSync, readFileSync } from 'fs'
+import fs, { existsSync, readFileSync } from 'fs'
 import { readdir, readFile, writeFile } from 'fs/promises'
 import { glob } from 'glob'
 import walk from 'ignore-walk'
 import { dump, load } from 'js-yaml'
 import { omit } from 'lodash'
+import * as pathModule from 'path'
```
j-zimnowoda: import only those function that are required in this file

src/common/utils.ts
ElderMatt: Agree, was too much thinking about adding things in the future.

src/common/utils.ts
ElderMatt: Did not know, I have changed it.

src/common/utils.ts
```diff
@@ -205,20 +206,29 @@ export const getSchemaSecretsPaths = async (teams: string[]): Promise<string[]>
...
-    await $`mkdir -p ${dirname(keepFilePath)}`
+export async function ensureManifestDirectories(): Promise<void> {
+  await ensureDirectoryWithGitkeepAsync(join(env.ENV_DIR, operatorEnv.GITOPS_MANIFESTS_NS_RELATIVE_DIR_PATH))
+  await ensureDirectoryWithGitkeepAsync(join(env.ENV_DIR, operatorEnv.GITOPS_GLOBAL_MANIFESTS_RELATIVE_PATH))
+}
+
+async function ensureDirectoryWithGitkeepAsync(
+  dirPath: string,
+  deps = { access, writeFile, mkdir, glob },
+): Promise<void> {
```
j-zimnowoda: glob is not used in this function

src/common/utils.ts
```diff
@@ -236,10 +246,7 @@ export async function ensureTeamGitOpsDirectories(
 
   await Promise.allSettled(
     keepFilePaths.map(async (keepFilePath) => {
-      await ensureKeepFile(keepFilePath, deps)
-      if (!existsSync(dirname(keepFilePath))) {
-        await $`mkdir -p ${dirname(keepFilePath)}`
-      }
+      await ensureDirectoryWithGitkeepAsync(dirname(keepFilePath), deps)
```
j-zimnowoda: is forwarding deps object necessary here?

src/operator/apl-operator.test.ts
```diff
@@ -70,9 +76,11 @@ jest.mock('./apl-operations', () => ({
 describe('AplOperator', () => {
   let aplOperator: AplOperator
   let defaultConfig: AplOperatorConfig
+  const fs = require('fs')
```
j-zimnowoda: I am not sure if this is the best way to mock it. We use deps as dependency injection pattern which makes packing deps very easy.

j-zimnowoda (approved):

src/cmd/bootstrap.ts
```diff
@@ -452,6 +461,7 @@ export const bootstrap = async (
   const originalValues = await deps.processValues()
   await deps.handleFileEntry()
   await deps.bootstrapSops()
+  deps.ensureManifestDirectories()
```
ferruhcihan:
```suggestion
  await deps.ensureManifestDirectories()
```

src/common/utils.ts
```diff
@@ -205,20 +206,26 @@ export const getSchemaSecretsPaths = async (teams: string[]): Promise<string[]>
...
-  if (existsSync(keepFilePath)) return
-  // create the .keep file
-  await deps.writeFile(keepFilePath, '')
 }
 
 export async function ensureTeamGitOpsDirectories(
   envDir: string,
   values: Record<string, any>,
-  deps = { writeFile, glob },
+  deps = { access, writeFile, mkdir, glob },
```
ferruhcihan: It seems these new dependencies are not used here. Can we remove them?

src/common/utils.ts
ElderMatt: it is not. its removed

src/common/utils.ts
ElderMatt: Yes they can be removed.

PR #3098, opened 2026-03-31
fix: set correct loki url when no obj storage

CasLubbers:
## 📌 Summary

Same logic is used in the prometheus operator: https://github.com/linode/apl-core/blob/main/values/prometheus-operator/prometheus-operator.gotmpl#L275
## 🧹 Checklist

### DIFF
--- a/helmfile.d/helmfile-60.teams.yaml.gotmpl
+++ b/helmfile.d/helmfile-60.teams.yaml.gotmpl
@@ -157,7 +157,7 @@ releases:
               uid: loki
               type: loki
               access: proxy
-              {{- if eq $v.obj.provider.type "minioLocal" "linode" }}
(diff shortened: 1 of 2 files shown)


helmfile.d/helmfile-60.teams.yaml.gotmpl
```diff
@@ -164,7 +164,11 @@ releases:
               uid: loki
               type: loki
               access: proxy
+              {{- if eq $v.obj.provider.type "minioLocal" "linode" }}
```
merll:
```suggestion
              {{- if eq $v.obj.provider.type "linode" }}
```
since we do not support `minioLocal` any longer.

CasLubbers: Also fixed the other reference

PR #3103, opened 2026-04-01
feat: use smart git-http-server

CasLubbers:
## 📌 Summary

Can only test with ORCS disabled. The PR is not yet merged, because of the recent cache registry issues.
## 🧹 Checklist


j-zimnowoda (commented):
1. I have successfully deployed. :white_check_mark: 
2. I like the fact that it is using istio sidecar, so mTLS can be set between communicating parties. :white_check_mark: 
3. I am wondering if mtls should be enforced though :warning:

4. I was also able to clone the repo::white_check_mark:
```
kubectl port-forward --namespace git-server svc/git-server 9090:80
git clone http://127.0.0.1:9090/otomi/values.git
```
```
kubectl get secret git-server-credentials -n git-server -o jsonpath='{.data.password}'
```
5. After enabling gitea things get messy, so fine tuning needs to be done so the platform does not try to switch to gitea. :warning:

charts/git-server/values.yaml
```diff
@@ -0,0 +1,31 @@
+repoName: otomi/values.git
+
+image:
+  repository: docker.io/clubbers1892/simple-git-server
```
j-zimnowoda: Once PoC is approved we will host it at docker.io/linode

charts/git-server/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+  pullPolicy: IfNotPresent
+
+initImage:
+  git:
+    repository: alpine/git
+    tag: latest
+    pullPolicy: IfNotPresent
+
+persistence:
+  size: 1Gi
```
j-zimnowoda: 256Mi should be more than enough

helmfile.d/helmfile-03.init.yaml.gotmpl
```diff
@@ -12,6 +12,15 @@ bases:
 {{- $a := $v.apps }}
 
 releases:
+  - name: git-server
+    installed: {{ not ($a | get "gitea.enabled") }}
```
j-zimnowoda: pls, make it independent from git. 
The `git-server` is going to be default mode. It should be possible to switch it to Gitea but that falls into to BYO git procedure and should not differ from switching to Github or other 3rd party provider.

charts/git-server/templates/deployment.yaml
```diff
@@ -0,0 +1,78 @@
...
+        - name: init-repo
+          image: {{ .Values.initImage.git.repository }}:{{ .Values.initImage.git.tag }}
+          imagePullPolicy: {{ .Values.initImage.git.pullPolicy }}
+          command:
+            - sh
+            - -ec
+            - |
+              if [ ! -d /data/{{ .Values.repoName }} ]; then
+                mkdir -p /data/{{ dir .Values.repoName }}
+                git init --bare --initial-branch=main /data/{{ .Values.repoName }}
```
CasLubbers: This branch needs to be configurable

charts/git-server/values.yaml
```diff
@@ -0,0 +1,31 @@
...
+
+image:
+  repository: docker.io/clubbers1892/simple-git-server
+  tag: v0.1.1
+  pullPolicy: IfNotPresent
+
+initImage:
+  git:
+    repository: alpine/git
+    tag: latest
```
ferruhcihan: We should use a pinned tag instead of `latest`.

ferruhcihan (commented):
Nice work, overall looks good to me. Just consider adding a `securityContext` (runAsNonRoot, drop capabilities) and/or an egress rule to the network policy to harden the deployment.

charts/git-server/templates/pvc.yaml
```diff
@@ -0,0 +1,16 @@
+apiVersion: v1
+kind: PersistentVolumeClaim
+metadata:
+  name: git-server-data
+  namespace: {{ .Release.Namespace }}
```
ferruhcihan, line 5:
```suggestion
  namespace: {{ .Release.Namespace }}
  annotations:
    "helm.sh/resource-policy": keep
```
Should we add this annotation to prevent data loss?

CasLubbers: I do not think we need it. After installation we would like that the git-server gets fully teared down. If we add this policy the PVC will keep on existing forever.

CasLubbers:
~Still need to test migration~
Migration is working. Git-server is disabled when migrating and git values are not adjusted.

charts/git-server/values.yaml
j-zimnowoda: We also need to review the code of simple-git-server ?

j-zimnowoda (dismissed):

charts/apl-network-policies/templates/networkpolicies/git-server.yaml
```diff
@@ -0,0 +1,50 @@
...
+              app.kubernetes.io/name: argocd-repo-server
+    # Allow APL API access
+    - from:
+        - namespaceSelector:
+            matchLabels:
+              name: otomi
+          podSelector:
+            matchLabels:
+              app.kubernetes.io/name: otomi-api
+    # Allow team build access for git clone
```
j-zimnowoda: This one is not needed because the server hosts only the values repo.

charts/git-server/templates/pvc.yaml
j-zimnowoda: yes @CasLubbers, you are right about that.

charts/git-server/templates/deployment.yaml
```diff
@@ -0,0 +1,73 @@
...
+kind: Deployment
+metadata:
+  name: git-server
+  namespace: {{ .Release.Namespace }}
+  labels:
+    {{- include "git-server.labels" . | nindent 4 }}
+spec:
+  replicas: 1
+  strategy:
+    type: Recreate
```
j-zimnowoda, line 11: It is great that you have chosen that type because the Pod relies on PV which is ReadWriteOnce, so any eviction to another node is going to be smooth (avoid the multiattach error).

This service also does not need to be HA and short unavailability is fine.

values/apl-network-policies/apl-network-policies.gotmpl
```diff
@@ -5,6 +5,7 @@
 
 # Simple per-app network policy configuration
 netpols:
+  gitServer: {{ not ($a | get "gitea.enabled") }}
```
j-zimnowoda: I would rather check if `git-server` is enabled

j-zimnowoda (commented):
Great work @CasLubbers
I left few comments but overall it is looking very good.

charts/git-server/templates/deployment.yaml
CasLubbers: Yes exactly! Nice that you've noticed it.

values/apl-network-policies/apl-network-policies.gotmpl
CasLubbers: Ah yes you're correct. This one is still al left over from the PoC

src/cmd/apply-as-apps.ts
```diff
@@ -163,7 +163,7 @@ export const getArgocdGitopsManifest = (name: string, targetNamespace?: string)
       {
         path,
         repoURL,
-        targetRevision: 'HEAD',
+        targetRevision: branch,
```
j-zimnowoda, line 166: Good catch

src/cmd/commit.ts
```diff
@@ -157,18 +157,14 @@ export const commit = async (
     // passwords because K8s secrets didn't exist yet. Now that secrets are decrypted,
     // we need to update the URL with the real credentials.
     await $git`git remote set-url origin ${remote}`.nothrow().quiet()
-  } else {
-    await setIdentity(username ?? 'otomi-admin', email, env.ENV_DIR)
-    // the url might need updating (e.g. if credentials changed)
-    await $git`git remote set-url origin ${remote}`
   }
   // let's wait until the remote is ready
   if (values?.apps!.gitea!.enabled ?? true) {
```
j-zimnowoda: That is incorrect. We should not wait for gitea even if it is enabled

src/common/envalid.ts
```diff
@@ -41,9 +41,6 @@ export const cliEnvSpec = {
     desc: 'The maximum amount of times to retry single post-installation steps',
     default: 5,
   }),
-  GIT_URL: str({ default: 'gitea-http.gitea.svc.cluster.local' }),
```
j-zimnowoda, line 44: That is fine. We do not need them anymore as those will come from the values repo.

src/common/values.ts
```diff
@@ -8,6 +8,7 @@ import { $ } from 'zx'
 import { decrypt, encrypt } from './crypt'
 import { terminal } from './debug'
 import { env } from './envalid'
+import { getRepo } from './git-config'
```
j-zimnowoda: it does not seem to be used here

src/common/values.ts
```diff
@@ -125,14 +126,18 @@ export const writeValues = async (inValues: Record<string, any>, overwrite = fal
...
-
+  const d = terminal('common:values:deriveSecrets')
   const secrets = {}
   const htpasswd = (
     await $`htpasswd -nbB ${values.apps.harbor.registry.credentials.username} ${values.apps.harbor.registry.credentials.password}`
   ).stdout.trim()
 
   set(secrets, 'apps.harbor.registry.credentials.htpasswd', htpasswd)
 
+  const gitHtpasswd = (
```
j-zimnowoda, line 136: Since the username comes from values repo and defaults.yaml already specifies it then there is no need to set a fallback value here.

src/operator/apl-operator.ts
```diff
@@ -170,6 +168,16 @@ export class AplOperator {
     this.d.info('Git polling loop stopped')
   }
 
+  private async reloadGitCredentials(): Promise<void> {
+    try {
```
j-zimnowoda, line 172: It would be worth comparing those credentials and informing user if a mismatch has been discovered.

values/apl-gitea-operator/apl-gitea-operator-raw.gotmpl
```diff
@@ -25,7 +25,7 @@ resources:
       name: apl-gitea-operator-secret
       namespace: apl-gitea-operator
     spec:
-      refreshInterval: 1h
+      refreshInterval: 1m
```
j-zimnowoda, line 28: yes, we need this to enable git switch. There should be an appropriate comment explaining that.

values/apl-operator/apl-operator-raw.gotmpl
```diff
@@ -6,7 +6,7 @@ resources:
   metadata:
     name: apl-git-credentials
   spec:
-    refreshInterval: 1h
+    refreshInterval: 1m
```
j-zimnowoda, line 9: ditto

values/argocd/argocd-raw.gotmpl
```diff
@@ -16,7 +16,7 @@ resources:
       name: argocd-repo-creds-gitea
       namespace: argocd
     spec:
-      refreshInterval: 1h
+      refreshInterval: 1m
```
j-zimnowoda, line 19: ditto

j-zimnowoda (approved):
Few minor improvements needed. Overall good.

src/cmd/commit.ts
j-zimnowoda: fixed

src/cmd/commit.ts
```diff
@@ -157,18 +157,15 @@ export const commit = async (
...
   }
-  // let's wait until the remote is ready
-  if (values?.apps!.gitea!.enabled ?? true) {
+
+  if (values?.otomi?.git?.repoUrl === 'http://gitea-http.gitea.svc.cluster.local:3000/otomi/values.git') {
+    // let's wait until the remote is ready
     await waitTillGitRepoAvailable(remote)
   }
   // continue
   await encrypt()
```
j-zimnowoda, line 167: Although it is our of scope of this PR, I do not understand why this function is still needed. 
cc @ferruhcihan

j-zimnowoda: I see that inside that function there is logic to skip it if unnecessary.

src/operator/apl-operator.ts
j-zimnowoda: TODO

j-zimnowoda (approved):
After fixing few minor issues (wrong eso version, wrong helmfile release order) I successfully deployed APL with light-weight git server. Well done @CasLubbers !

src/common/values.ts
j-zimnowoda: TODO

PR #3123, opened 2026-04-10
ci: add ACL and firewall rules in deploy action

merll:
## 📌 Summary

On test deployments using the GitHub action, this PR adds a restrictive ACL to the control plane. In addition, Linode Cloud Firewall Controller is configured to only allow traffic from internal addresses.

Autoscaling has been removed since it was both set to min and max = 3.

## 🔍 Reviewer Notes

Job https://github.com/linode/apl-core/actions/runs/24333919669/job/71045997719 has been run during testing.

## 🧹 Checklist


CasLubbers (approved):

.github/workflows/integration.yml
```diff
@@ -129,7 +129,8 @@ env:
   DEV_DOMAINS: ${{ secrets.DEV_DOMAINS }}
   LKE_TIER: ${{ inputs.lke_tier }}
   LKE_VERSION: ${{ inputs.kubernetes_version }}
-  LKE_CP_ACL_IPV4: '0.0.0.0/0'
+  LKE_CP_ACL_IPV4: ${{ vars.LKE_CP_ACL_IPV4 }}
+  LKE_APL_FIREWALL_INBOUND_RULES: ${{ vars.LKE_APL_FIREWALL_INBOUND_RULES }}
```
CasLubbers: Is it safe enough to put these ip-adresses in the vars? If they are not publicly known we could better put them in a secret

merll: Not sure how sensitive they are, but I moved them.

PR #3127, opened 2026-04-13
chore: remove unused tracing collector

merll:
## 📌 Summary

This is a follow-up from removing Tempo. There is no more backend in the platform for tracing. The OTLP collector would at this point still gather logs from connected workloads and Istio, but just log them to stdout. Distributed tracing needs to be redesigned with Istio ambient mesh and the Gateway API.

## 🧹 Checklist


CasLubbers (approved):

ElderMatt (approved):
Dont forget to update the version change to 61

PR #3138, opened 2026-04-14
fix: remove hardcoded command from apl-operator chart

CasLubbers:
## 🧹 Checklist


j-zimnowoda (approved):

charts/apl-operator/values.yaml
```diff
@@ -56,6 +56,10 @@ tolerations: []
 
 affinity: {}
 
+command: []
+
+args: []
```
j-zimnowoda: `commandArgs` property  could be considered instead of `args`

merll (approved):

PR #3144, opened 2026-04-15
feat: generate redis secret for ArgoCD deployment

ElderMatt:
This PR creates a argocd-redis secret during install and migration. This is to help with the issue that argocd might occasionally get stuck during an update and it tries to reconcile itself.


src/cmd/migrate.ts
```diff
@@ -882,6 +934,53 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
...
+      } catch (error) {
+        if (error instanceof ApiException && error.code === 404) {
+          d.debug(`Could not restart ${target.kind}/${target.name}: not found`)
+          continue
+        }
+        throw error
+      }
+    }
+  } catch (error) {
+    d.error('Failed to remove legacy ArgoCD redis secret, continuing migration:', error)
```
ferruhcihan:
```suggestion
    d.error('Failed to create/update ArgoCD redis secret, continuing migration:', error)
```

src/cmd/migrate.ts
```diff
@@ -882,6 +934,53 @@ const setDefaultAplCatalog = async (values: Record<string, any>): Promise<void>
...
 }
 
+export const addRedisSecretForArgoCD = async (values: Record<string, any>): Promise<void> => {
+  const d = terminal('addRedisSecretForArgoCD')
+  const argocdNamespace = 'argocd'
+
+  try {
+    const parsedArgs = getParsedArgs()
+    if (parsedArgs?.dryRun || parsedArgs?.local || env.DISABLE_SYNC) {
+      d.info('Skipping cluster cleanup in dry-run/local/dev mode')
```
ferruhcihan:
```suggestion
      d.info('Skipping ArgoCD redis secret creation in dry-run/local/dev mode')
```

src/cmd/install.ts
```diff
@@ -166,6 +167,18 @@ const install = async (): Promise<void> => {
...
 
+const prepareMandatorySecrets = async (): Promise<void> => {
+  const d = terminal(`cmd:${cmdName}:prepareMandatorySecrets`)
+  d.info('Preparing mandatory secrets for installation')
+
+  // Ensure ArgoCD Redis Secret exists and has Helm ownership metadata before Helm applies Argo CD.
+  d.info('Creating argocd-redis secret when possible')
+  const values = (await hfValues()) as Record<string, any>
+  await createArgoCdRedisSecret(values).catch((error) => {
+    d.error('Failed to create argocd-redis secret:', getErrorMessage(error))
```
ferruhcihan: It seems that if the secret creation fails, the installation will continue after this log. Should we log this with `d.warn` or do something else here, because ArgoCD will also fail?

src/common/utils.ts
```diff
@@ -245,6 +247,78 @@ export async function ensureTeamGitOpsDirectories(envDir: string, values: Record
   return keepFilePaths
 }
 
+export const createArgoCdRedisSecret = async (values: Record<string, any>): Promise<void> => {
```
ferruhcihan: This looks more like a `k8s` function than a utils function. Should we move it to `src/common/k8s.ts`?

ferruhcihan (changes requested):

src/common/utils.ts
ElderMatt: I think that is a good idea, I'll move it!

src/cmd/install.ts
ElderMatt: We can do a warning yes.

src/common/k8s.ts
```diff
@@ -896,6 +896,78 @@ export async function setArgoCdAppSync(
   )
 }
 
+export const createArgoCdRedisSecret = async (values: Record<string, any>): Promise<void> => {
+  const d = terminal('common:utils:createArgoCdRedisSecret')
```
ferruhcihan:
```suggestion
  const d = terminal('common:k8s:createArgoCdRedisSecret')
```

ferruhcihan (approved):

CasLubbers (approved):

PR #3152, opened 2026-04-16
ci: added job filter to avoid pending jobs

merll:
## 📌 Summary

In an attempt to sort out forever-pending tasks, e.g. on external contributions, this adds a filter on the repository.

## 🧹 Checklist


.github/workflows/main.yml
```diff
@@ -5,6 +5,13 @@ on:
       - '**'
     tags-ignore:
       - '*'
+  pull_request:
```
j-zimnowoda, line 8: will that automatically trigger all our GH Actions pipelines for external PRs?
Should we consider that changes to files in .github/ should not trigger them?

merll: Changes in `.github/` are not handled in the workflows before they are merged, not even for internal ones. That can only done with a workflow_dispatch.

.github/workflows/main.yml
```diff
@@ -5,6 +5,13 @@ on:
...
+    types:
+      - opened
+      - synchronize
+      - reopened
+    branches:
+      - '**'
```
j-zimnowoda: I am wondering if we should restrict branch naming and not allowing branches that could overlap with our tagging system, e.g.. to prevent pushing v4.16.9 branch.
What is your take?

merll: We would not be prevent the creation of the branch. This should not happen on external repositories anyway. However, the previous approach does not seem to lead the way we want to. Therefore I reversed changes to the trigger and added filters instead. That should skip the tasks that are stuck in a pending state.

j-zimnowoda (approved):

PR #3153, opened 2026-04-16
ci: fix dev cluster update after merge

merll:
## 📌 Summary

This PR updates the dev cluster's ACL temporarily, for triggering a restart of updated services after merge to main. The ACL is reset back to its default afterwards.

## 🔍 Reviewer Notes

The script `ci/scripts/trigger_dev.sh` can be tested locally by setting the following variables (assumes authenticated linode-cli):
* `DEV_CLUSTER_ID`: can be any valid cluster ID, but the cluster must run Harbor since the script assumes the operator is deployed.
* `LKE_CP_ACL_IPV4`: comma-separated list of valid IP ranges. Check dev cluster for the actual values.

## 🧹 Checklist

### DIFF
--- a/ci/scripts/trigger_dev.sh
+++ b/ci/scripts/trigger_dev.sh
@@ -23,6 +23,7 @@ echo "Updating ACL"
 linode-cli lke cluster-update "${update_args[@]}"
 
 echo "Restart platform deployments"
+set +e
 kubectl -n otomi rollout restart deployment/otomi-api
 kubectl -n otomi rollout restart deployment/otomi-console
 kubectl rollout restart deployment -n apl-harbor-operator apl-harbor-operator
--- a/ci/scripts/trigger_dev.sh
+++ b/ci/scripts/trigger_dev.sh
@@ -3,11 +3,24 @@
 
 set -e
 
-echo "Decode and set the Kubernetes configuration for the dev environment"
-if [ -z "$KUBECONFIG" ]; then
-  echo $DEV_KUBECONFIG_64 | base64 -d >.kubeconfig
-  export KUBECONFIG=$(pwd)/.kubeconfig
+echo "Fetch the Kubernetes configuration for the dev environment"
+export KUBECONFIG=$(mktemp -d)/.kubeconfig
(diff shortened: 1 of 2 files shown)


ci/scripts/trigger_dev.sh
```diff
@@ -3,11 +3,24 @@
...
+  IFS=',' read -ra acl_ips <<< "${LKE_CP_ACL_IPV4}"
+  for acl_ip in "${acl_ips[@]}"; do
+    update_args+=(--control_plane.acl.addresses.ipv4 "$acl_ip")
+  done
 fi
+reset_args=("${update_args[@]}")
+update_args+=(--control_plane.acl.addresses.ipv4 "$CURRENT_IP")
+
+echo "Updating ACL"
+linode-cli lke cluster-update "${update_args[@]}"
```
j-zimnowoda, line 23: Is the ACL update immediate?
If so then you can simply resolve that comment.

j-zimnowoda (approved):

ci/scripts/trigger_dev.sh
```diff
@@ -17,3 +30,6 @@ kubectl rollout restart deployment -n apl-keycloak-operator apl-keycloak-operato
 kubectl rollout restart deployment -n apl-gitea-operator apl-gitea-operator
 kubectl rollout restart deployment -n otomi-operator otomi-operator
 kubectl rollout restart deployment -n apl-operator apl-operator
+
+echo "Resetting ACL"
+linode-cli lke cluster-update "${reset_args[@]}"
```
ElderMatt, line 47: If one of the kubectl commands fail then the reset won't happen, it would be nice to add some cleanup so that the reset happens even if something fails.

ci/scripts/trigger_dev.sh
merll: The API recorded it immediately, but after a couple of more tests it turned out that there was in fact a delay in the propagation to the control plane. Therefore I added a waiting cycle which should fix this issue, or fail the script if it never goes through.

ci/scripts/trigger_dev.sh
merll: I changed the error handling so that part of the script does not fail on error.

ElderMatt (approved):

PR #3178, opened 2026-04-22
feat: restart ArgoCD controller in case of unrecoverable error

merll:
## 📌 Summary

This PR checks unrecoverable errors (nil pointer exception) on all ArgoCD applications, and if it occurs, restarts the ArgoCD Controller. Some other code regarding resource updates has been moved for better interoperability.

## 🔍 Reviewer Notes

The controller restart can be tested by updating any ArgoCD application using a script:
```sh
kubectl patch -n argocd application.argoproj.io/apl-gitea-operator-apl-gitea-operator \
  --type merge \
  -p '{"status":{"operationState":{"phase":"Failed","message":"runtime error: invalid memory address or nil pointer dereference"}}}'
```

This however will not be automatically be resolved by the controller and the fake error status remains. Since there is no sync process that failed and will be restarted, the status will not be cleared until another change on the application requires a sync, or it is triggered automatically.

The actual problem at hand is not reproducible this way.

## 🧹 Checklist


src/common/k8s.ts
```diff
@@ -835,6 +835,19 @@ export async function waitForArgoCDAppHealthy(
   )
 }
 
+export function argoCdHasUnrecoverableErrors(applications: Record<string, any>[]): string | undefined {
```
j-zimnowoda, line 838: unit tests please.

Below one example from real cluster
```
apiVersion: argoproj.io/v1alpha1
kind: Application
metadata: {}
spec: {}
status:
  health:
    lastTransitionTime: "2026-04-23T06:57:14Z"
    status: Healthy
  operationState:
    finishedAt: "2026-04-23T06:57:12Z"
    message: successfully synced (all tasks run)
    phase: Succeeded
    startedAt: "2026-04-23T06:57:12Z"
```

j-zimnowoda (approved):

src/common/k8s.ts
merll: Added a unit test for this function now, but also for `checkArgoCdController` which is using it.

j-zimnowoda: Thank you!

PR #3190, opened 2026-04-29
feat: add app label to http route

j-zimnowoda:
## 📌 Summary

Unify HTTPRoute labels.

Depends-on: https://github.com/linode/apl-api/pull/985

## 🔍 Reviewer Notes

HTTPRoute for ksvc is managed by ksvc serving.

## 🧹 Checklist


charts/team-ns/templates/routes.yaml
```diff
@@ -17,7 +17,8 @@
 {{- $backendName := $s.svc | default $s.name }}
 {{- $backendPort := $s.port | default 80 }}
 {{- $useTrafficControl := dig "trafficControl" "enabled" false $s }}
-{{- $serviceName := print $s.name "-" $v.teamId }}
+{{- $serviceName := $s.name }}
+{{- $serviceLabel := $s.name }}
 {{- $defaultHostname := print $serviceName "." $v.domain }}
```
merll: Note that the team suffix is needed for the hostname, since that one is shared across the cluster and needs not to conflict with workloads in other namespaces.

```suggestion
{{- $serviceName := $s.name }}
{{- $serviceLabel := $s.name }}
{{- $defaultHostname := print $serviceName "-" $v.teamId "." $v.domain }}
```

Could we also set the label below to `$serviceName`? In that case we would not need the duplicate variable.

merll (approved):

PR #3213, opened 2026-05-07
fix: hostNetwork support in alertmanager spec

ferruhcihan:
## 📌 Summary

This PR fixes the alert manager hostNetwork field not declared in schema error.

## 🧹 Checklist


merll (approved):
Approving this fix, pending check on more recent version of the chart for future releases.

merll (changes requested):
I checked an updated chart and realized that the underlying problem is something else. The value `hostNetwork` is not part of the chart, that has a matching CRD. But we never update our CRDs. We need to solve the problem there, not by modifying charts, because we can run into similar issues with other apps that define custom resources.
On install, we do this via `kubectl apply`. When this is a managed ArgoCD app, this should be included in the app from there.

merll:
This problem seems to be specific to this chart, since we explicitly exclude the CRDs from deployment. I opened #3219 as an alternative solution, that has ArgoCD take care of it. Need to test with a fresh cluster however.

ferruhcihan:
> This problem seems to be specific to this chart, since we explicitly exclude the CRDs from deployment. I opened #3219 as an alternative solution, that has ArgoCD take care of it. Need to test with a fresh cluster however.

Thanks for the alternative solution, I'll test it with a fresh cluster.

ferruhcihan:
The issue can be resolved with the alternative solution (https://github.com/linode/apl-core/pull/3219), so this PR is being closed.

PR #3218, opened 2026-05-07
chore: upgrade base image of core to v2.11.2

merll:
## 📌 Summary

This PR upgrades the base image that is used to build Core, and some binary references used for Gitea application backup. The base image is created using Ubuntu 26 (upgrade from 24), and utility binaries (e.g. Git, Kubectl, Helm, Helmfile etc) were upgraded to more recent versions, either matching the versions used in other places of the charts (e.g. Tekton, ArgoCD) or to the latest minor version.

## 🧹 Checklist


j-zimnowoda (approved):
Confirmed that the artifacts exists.

PR #3219, opened 2026-05-07
fix: include monitoring crds in chart for handling upgrades reliably

merll:
## 📌 Summary

This PR includes the CRDs required by Kube-Prometheus-Stack in the chart, considering them on both initial install and during upgrades.

They still must be applied during initial install, as other applications include ServiceMonitor resources.

## 🧹 Checklist


ferruhcihan (approved):
Tested with a fresh cluster and upgrade scenarios. It works, LGTM.

PR #3222, opened 2026-05-08
fix: failed connection between grafana and alertmanager

ferruhcihan:
## 📌 Summary

This PR fixes Grafana's Alertmanager datasource access by adding grafana/po-grafana to the monitoring AuthorizationPolicy. (Please check https://github.com/linode/apl-core/pull/3068)

## 🧹 Checklist


merll (approved):

merll:
Is this working for the team Alertmanager? Or do we need to add the exemption there as well?

merll:
> Is this working for the team Alertmanager? Or do we need to add the exemption there as well?

nvm, that should not be an issue. The Alertmanager of each team is only configured in the same team's Grafana instance. They all run in the same namespace, which already has the rules set accordingly.

PR #3234, opened 2026-05-12
fix: pvc migration for the gitea-gitea-valkey

ferruhcihan:
## 📌 Summary

This PR fixes the PVC migration for the `gitea-gitea-valkey`. 

LKE Enterprise upgrade from `v4.15.3` to `v5.0.0` could leave Gitea unavailable due to ArgoCD sync being permanently disabled. Enterprise clusters use `linode-block-storage-retain` as the default StorageClass, which triggers a PVC migration that disables ArgoCD sync for `gitea-gitea-valkey`. The sync was never re-enabled because `applyAll()` exited early before reaching `applyAsApps` when an operator version change was detected. 

The issue is fixed by making the migration idempotent on re-runs, and throwing on timeout instead of silently continuing during the migration.

## 🧹 Checklist


src/cmd/migrate.ts
```diff
@@ -618,20 +615,31 @@ async function migrateStatefulSetPvc(opts: {
...
+    if (error instanceof ApiException && error.code === 404) {
+      opts.d.info(`StatefulSet ${opts.statefulSetName} not found — already deleted. Skipping migration.`)
+      return
+    }
+    throw error
+  }
 
-    await waitForPodsDeletion(opts.namespace, opts.pvcLabelSelector)
-    await deletePvcsByLabel(opts.namespace, opts.pvcLabelSelector)
+  const app = await getArgoCdApp(opts.appName, k8s.custom())
```
merll, line 618: If I understand this change correctly, the auto-sync of the ArgoCD application is disabled now after the StatefulSet is patched. I think this could be subject to a race condition, as ArgoCD could sync and restore the number of replicas to > 0, effectively blocking the PVC deletion even after the sync is disabled already.

src/cmd/apply.ts
```diff
@@ -37,6 +37,14 @@ export const applyAll = async (): Promise<void> => {
   const revisionUpdated = await updateOperatorApplication(env.APPS_REVISION || tag)
   if (revisionUpdated) {
     d.info('Operator has pending update to a different revision. Pausing until restart.')
+    // Apply ArgoCD Application CRs so any sync policies disabled during migration
```
merll: With this change, we are already applying all applications with current values. This may work to solve this particular problem, but can cause other issues since pre-upgrade scripts (below) do not run for different releases. The purpose of this early exit is that values (migrations), runtime upgrades, and all app revisions (but most importantly the apl-operator that includes the Helm charts in the image) are in sync at all times. It was moved here recently on purpose.

I think we need to address the underlying problem instead, causing a waiting cycle. What I found to happen during a "otomi.version" change in earlier releases is that
1. the operator updates its image, but not the app revision
2. it restarts, and detects the difference in revision, and pauses
3. due to no change in the image, it does not restart, but completes changes on the next reconciliation cycle, leaving apps out of sync for a long time

The inconsistency from 1. is caused in an earlier release, but only resolved after the upgrade. For fixing this **during** the upgrade, I would suggest that the operator does not only check its revision, but also its tag. If the revision does not match, but the tag does
1. either we could either exit the operator with a zero code, and wait for a restart which would likely happen faster than the reconciliation
2. or the operator could just continue. This would only be the case for this release, as the upgrade should be done consistently on the next one. Probably this works good as well.

merll (changes requested):
I have concerns that while solving this particular issue this implementation will cause another one in following releases.

src/cmd/migrate.ts
ferruhcihan: Updated.

src/cmd/apply.ts
ferruhcihan: Updated, and the tag check has been added.

merll: My suggestion of checking the tag was ambiguous. The change will likely not solve the issue (just tested by switching to a feature branch, and still see that issue). What I meant was, to check the `tag` of the ArgoCD application values.

ferruhcihan: The real issue is the silent return during the migration. This part implemented to reduce the reconcile iterations but it might be unnecessary and overcomplicated. So this part has been reverted.

merll (approved):

PR #3289, opened 2026-05-29
ci: optimize docker cache usage

j-zimnowoda:
## 📌 Summary

Install dependencies before copying the source code to take advantage of Docker layer caching

## 🧹 Checklist

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -14,7 +14,7 @@ ENV DISABLE_SYNC='1'
 ENV NODE_PATH='dist'
 
 # Install dependencies before copying the source code to take advantage of Docker layer caching
-COPY --chown=app  package*.json ./
+COPY --chown=app package*.json ./
 RUN npm config set update-notifier false
 RUN npm ci --ignore-scripts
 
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -13,13 +13,13 @@ ENV VERBOSITY='2'
...
-COPY --chown=app . .
-
-RUN set -e && \
-    npm config set update-notifier false && \
-    npm ci --ignore-scripts && \
-    npm run compile
+# Install dependencies before copying the source code to take advantage of Docker layer caching
+COPY --chown=app  package*.json ./
+RUN npm config set update-notifier false
+RUN npm ci --ignore-scripts
```
merll, line 19: Could we concatenate these two? That would make one (very small) layer less in the image.

```suggestion
RUN npm config set update-notifier false && npm ci --ignore-scripts
```

j-zimnowoda: As far as I understand, this is multi stage build, so the final layer is calculated in the last stage.

merll: Sure, I only saw the updated portion so it does indeed not make much of a difference.

merll (approved):

merll (approved):

PR #3325, opened 2026-06-08
fix: added authorization policies per team to per-team deployed apps

merll:
## 📌 Summary

This PR introduces more AuthorizationPolicy CRs (Istio) that limit access to certain services by the `group` claim extracted from the JWT token. This affects Grafana, Alertmanager, and Tekton which are all deployed separately for each team if enabled.

## 🔍 Reviewer Notes

Prometheus is shared between all platform teams, therefore access should be restricted only to all authenticated users. The platform admin should be able to access all team's apps. However, regular team users of "team1" should not be able to access platform apps and other team's apps:
* grafana.<domain> (although Grafana also restricts through its own authentication)
* alertmanager.[domain]
* tekton.[domain]
* grafana-team2.[domain]
* alertmanager-team2.[domain]
* tekton-team2.[domain]
but should be able to open
* prometheus.[domain]
* grafana-team1.[domain]
* alertmanager-team1.[domain]
* tekton-team1.[domain]

## 🧹 Checklist


helmfile.d/snippets/authpolicy-jwt.gotmpl
```diff
@@ -20,10 +24,13 @@ spec:
     matchLabels:
       otomi.io/auth-policy: {{ $v.name }}
   rules:
+    {{- if or (not $allowGroups) $excludeNamespaces $excludeAccounts }}
```
j-zimnowoda: Does the `not $allowGroups` mean `onlyAuthenticate`? If so then I propose to create here such a variable. I am wondering if this should be part of this template interface or simply derived from provided arguments.

values/kubernetes-gateways/kubernetes-gateways-raw.gotmpl
```diff
@@ -90,7 +90,15 @@ resources:
...
   {{- end }}
+  {{- /* Generic AuthorizationPolicy that allows access by any authenticated user */}}
   - {{ tpl (readFile "../../helmfile.d/snippets/authpolicy-jwt.gotmpl") (dict "name" "platform") | nindent 4 }}
+  {{- /* Generic AuthorizationPolicy that only allows access by platform-admin group */}}
+  - {{ tpl (readFile "../../helmfile.d/snippets/authpolicy-jwt.gotmpl") (dict "name" "platform-admin" "allowGroup" "platform-admin") | nindent 4 }}
+  {{- /* Generic AuthorizationPolicy that only allows access by each team and platform-admin group */}}
+  {{- range $teamId := (keys (omit $v.teamConfig "admin") | sortAlpha) }}
+  {{- $groupName := print "team-" $teamId }}
+  - {{ tpl (readFile "../../helmfile.d/snippets/authpolicy-jwt.gotmpl") (dict "name" "platform-team" "namespace" $groupName "allowGroups" (list "platform-admin" $groupName)) | nindent 4 }}
+  {{- end }}
```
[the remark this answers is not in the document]
j-zimnowoda: Seems like a valid point during the initial install.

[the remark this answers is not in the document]
merll: Team namespaces are not created in `helmfile-60.teams.yaml.gotmpl`, but in `helmfile.tpl/helmfile-init.yaml.gotmpl`. That one is used by a function `deployEssential` called early on initial install, and with a filter before every team-apply.
It would have been logical to place this more close to the team manifests, but adding another `raw` item, including the ArgoCD application and everything that is attached to it seemed out of proportion for this.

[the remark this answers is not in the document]
j-zimnowoda: You are right. So as long as apl creates namespaces at the very beginning of the deployment, there is no risk of having deadlock.

j-zimnowoda (approved):
LGTM

helmfile.d/snippets/authpolicy-jwt.gotmpl
merll: I can add this variable. However, I am not sure what you mean with deriving this from other arguments. "allow specific groups" vs. "allow any authenticated user" are the two options that are mutually exclusive. Therefore, no groups means we allow any authenticated access.

The other arguments are needed for other purposes (e.g. allowing specific k8s service accounts).

values/tekton-dashboard/tekton-dashboard-teams.gotmpl
```diff
@@ -12,7 +12,7 @@ image:
 
 podLabels:
   otomi.io/auth: platform
-  otomi.io/auth-policy: platform
+  otomi.io/auth-policy: platform-team
```
j-zimnowoda, line 15: Shouldn't it be `platform-admin` ?

j-zimnowoda: Could you clarify the context in which we use `platform-admin` vs `platform-team`?

merll: `platform-team` is correct here, since it is the per-team Tekton Dashboard. There is a comment where the AuthorizationPolicy is being created: https://github.com/linode/apl-core/pull/3325/changes#diff-466b1931dc5f1dd079bee49eb64b63ecdf6baa4c19a24394e521c68e3ef04315R97
The policy also permits the `platform-admin` group to access the service, in addition to the team users. This is a generic policy, which would technically also work for other services. The monitoring stack requires some exemptions, and therefore has a policy on its own.

PR #3340, opened 2026-06-12
fix: prometheus exporter scrape targets

merll:
## 📌 Summary

This PR fixes the scrape targets set up with Prometheus, and processed by the Prometheus Blackbox Exporter:
* The domain field was removed from services.
* For Knative Services, the host name is generated in a slightly different pattern.
* CNAME are added where configured.

## 🧹 Checklist


j-zimnowoda:
Readable diff
image
image

j-zimnowoda (approved):

PR #3363, opened 2026-06-18
feat: generate team loki password

j-zimnowoda:
## 📌 Summary

Read: https://github.com/linode/apl-core/issues/3362
Depends-on: https://github.com/linode/apl-api/pull/1019

## 🧹 Checklist


charts/apl-operator/templates/deployment.yaml
```diff
@@ -5,7 +5,7 @@ metadata:
   labels:
     {{- include "apl-operator.labels" . | nindent 4 }}
 spec:
-  replicas: 1
+  replicas: {{ .Values.scale.replicas | default 1 }}
```
j-zimnowoda, line 8: We should be able to set replicas, so in the development mode we can scan it to zero.

merll (approved):

PR #3423, opened 2026-07-06
fix: default storageClasses

### LINKED ISSUE #3386: Lack of PVC Encryption For APL Generated Resources
# Is your feature request related to a problem? Please describe.

On Linode/LKE, APL-provisioned PersistentVolumeClaims end up unencrypted at rest unless a downstream consumer goes out of their way to prevent it, and there's no single APL setting that reliably enforces encryption cluster-wide. Three distinct gaps make this frustrating:

1. No global "encrypt all storage" knob. cluster.defaultStorageClass only steers charts that respect it. There is no platform-level setting that guarantees every PVC APL creates lands on an encrypted, correctly-tagged StorageClass.

2. Some charts hardcode storageClass: linode-block-storage and can't be overridden. At least two ship today with the class baked in and ignore cluster.defaultStorageClass:

         a. gitea-valkey → PVC valkey-data-gitea-valkey-primary-0 (namespace gitea)
         b. oauth2-proxy redis-HA → PVC data-oauth2-proxy-redis-ha-server-0 (namespace istio-system)

          These provision on the stock LKE class, which is unencrypted and untagged, with no way to fix it via values.

3.  Loki's storage PVC isn't covered by cluster.defaultStorageClass either, so it has to be pinned to an encrypted class explicitly or it silently lands on an unencrypted, Delete-reclaim class.

The net effect: operators who need encryption-at-rest (a common compliance requirement) cannot trust that APL workloads are encrypted, and the failure is silent — a volume can appear fine while being unencrypted. This is made worse by a Linode CSI driver quirk where misspelled parameter keys (e.g. encryption: enabled instead of encrypted: "true", or volume-tags instead of volumeTags) are ignored without any warning, so a StorageClass can look encrypted but not be.

# Describe the solution you'd like

A first-class, platform-level way to enforce storage encryption for all APL-managed PVCs on Linode. Ideally:

- A single APL/values setting (e.g. cluster.storage.encrypted: true or an encryption-aware defaultStorageClass) that is honored by every chart APL deploys — no per-chart exceptions.
Charts that currently hardcode storageClass: linode-block-storage (gitea-valkey, oauth2-proxy redis-HA) updated to respect cluster.defaultStorageClass (or a dedicated storage-class override value) instead of a baked-in class.
- Loki's storage PVC brought under the same cluster.defaultStorageClass coverage as everything else.
- The shipped Linode StorageClass(es) verified to use the exact CSI parameter keys the driver honors (linodebs.csi.linode.com/encrypted: "true", linodebs.csi.linode.com/volumeTags: "..."), since misspellings fail silently.
Describe alternatives you've considered

We currently work around this entirely downstream, which is brittle and high-maintenance:

- A Terraform-deployed encrypted StorageClass (block-storage-retain, Retain reclaim, volumeBindingMode: Immediate) applied before apl-operator installs, set as cluster default.
- A Kyverno admission ClusterPolicy that rewrites the two hardcoded PVCs from linode-block-storage → the encrypted class at creation time.
- A second Kyverno ClusterPolicy that continuously demotes the stock LKE default StorageClass (since Flux keeps re-promoting it on ~10m reconcile), replacing an earlier PostSync Job that only ran on Argo sync and left gaps.
- Explicitly pinning Loki to the encrypted class in values.
- A post-bootstrap audit (audit-pvc-storageclass) that flags any PVC that escaped onto an unencrypted class.

These work but are inherently racy: the Kyverno mutations are admission-only, so any PVC created before the webhook is ready slips through permanently, and already-provisioned unencrypted volumes can only be remediated by destroying and recreating the workload + PVC (data loss). All of this would be unnecessary if APL enforced encryption natively.

# Additional context

- Affected charts observed: gitea-valkey (gitea ns) and oauth2-proxy redis-HA (istio-system ns) hardcode the storage class; Loki's storage PVC is uncovered by cluster.defaultStorageClass.
- Linode CSI silent-failure detail: the driver ignores unrecognized/misspelled parameter keys with no event or warning, so encryption misconfiguration is invisible until you inspect the actual volume in the Linode API. Correct keys are linodebs.csi.linode.com/encrypted: "true" and linodebs.csi.linode.com/volumeTags.
- Untagged volumes are an additional problem beyond encryption: they escape tag-filtered cleanup and leak Linode block-storage quota as orphaned volumes.
- Remediation for an already-escaped PVC today: scale down the owning StatefulSet/Deployment, delete the PVC, re-sync so the replacement is created on the encrypted class — irreversible for existing data.

j-zimnowoda:
## 📌 Summary

#3386

## 🧹 Checklist


values/gitea/gitea-valkey.gotmpl
```diff
@@ -1,9 +1,11 @@
 {{- $v := .Values }}
 
 global:
-  defaultStorageClass: {{ if eq $v.cluster.provider "linode" }}linode-block-storage{{ else }}{{ $v.cluster.defaultStorageClass | default "" }}{{ end }}
   security:
     allowInsecureImages: true
+  {{- with $v.cluster | get "defaultStorageClass" nil }}
```
j-zimnowoda, line 6: Upgrade path still needs to be implemented

merll (approved):

PR #3443, opened 2026-07-14
ci: agent skills add chart

j-zimnowoda:
## 📌 Summary
This pull request introduces several improvements to the developer workflow, documentation, and test coverage for adding and exposing Helm charts and apps. The main focus is on enforcing template diff checks, adding reusable skills for common tasks, and improving testability and automation in the CI process.

## 🔍 Reviewer Notes

Tested skills:
```
/apl-add-app https://kubernetes-sigs.github.io/headlamp/ headlamp
/apl-add-app-ingress headlamp --configure-public-security=true
```

## 🧹 Checklist


ci/src/add-app-helm-chart.ts
```diff
@@ -0,0 +1,402 @@
...
+    'releases:',
+  ]
+
+  parts.push(
+    `  - name: ${name}`,
+    `    installed: {{ $a | get "${name}.enabled" }}`,
+    `    namespace: ${name}`,
+    '    labels:',
+    `      pkg: ${name}`,
+    '      app: core',
```
CasLubbers: We do not want external users to add new apps to the core right? Otherwise it would be better to offer that as an option.

ci/src/add-app-helm-chart.ts
```diff
@@ -0,0 +1,402 @@
...
+  console.log(`Vendored chart into charts/${name}`)
+}
+
+export function createHelmfile(root: string, name: string): string {
+  const next = findHighestHelmfileNumber(path.join(root, 'helmfile.d')) + 1
+  const padded = String(next).padStart(2, '0')
+  const relPath = `helmfile.d/helmfile-${padded}.${name}.yaml.gotmpl`
+  const fullPath = path.join(root, relPath)
+
+  const parts: string[] = [
```
CasLubbers, line 229: If possible, it would be easier that these templated blocks are in a separate file. That way it is also more future proof. Same for the block on line 250

ci/src/configure-public-exposure.ts
```diff
@@ -0,0 +1,381 @@
...
+    throw new Error(`Could not locate release ${name} in ${path.basename(helmfilePath)}`)
+  }
+
+  const block = [
+    `  - name: ${name}-artifacts`,
+    `    installed: {{ $a | get "${name}.enabled" }}`,
+    `    namespace: ${name}`,
+    '    labels:',
+    `      pkg: ${name}`,
+    '      app: core',
```
CasLubbers: Same here core should not be set.

ci/src/configure-public-exposure.ts
```diff
@@ -0,0 +1,381 @@
...
+    return
+  }
+
+  const releaseIdx = lines.findIndex((line) => line === `  - name: ${name}`)
+  if (releaseIdx < 0) {
+    throw new Error(`Could not locate release ${name} in ${path.basename(helmfilePath)}`)
+  }
+
+  const block = [
+    `  - name: ${name}-artifacts`,
```
CasLubbers, line 140: Same as above, would be better if these blocks are in a separate file. Same for the block on line 238

.github/skills/apl-add-app/SKILL.md
```diff
@@ -0,0 +1,51 @@
+---
+name: apl-add-app
+description: Add and wire a new Helm chart into APL Core by updating chart index, vendoring the chart, scaffolding Helmfile/defaults/values files, and running required validation checks. Use when adding a new app chart and the user provides or asks for name and repository; the script resolves the latest official stable semver chart version automatically.
+argument-hint: <name> <repository>
+---
+
+# Add Helm Chart
+
+## Quick start
```
CasLubbers, line 9: When adding an app there are changes required in apl-console right? I think it is for the icon?
Would be nice if we could move that to apl-core so only changes are needed in apl-core.
Otherwise we could ask the user where its apl-console repository is located and make the change.

CasLubbers (changes requested):
Some required changes and nice to haves.

.github/skills/apl-add-app/SKILL.md
j-zimnowoda: That is right. I haven't made my mind about the full process.

ci/src/add-app-helm-chart.ts
j-zimnowoda: Very good point!

ci/src/configure-public-exposure.ts
j-zimnowoda: Nice catch!

PR #3465, opened 2026-07-28
fix(schema): don't mark generator-backed x-secret fields as required

### LINKED ISSUE #3422: Generator-backed x-secret fields are marked required in values.schema.json, so any install path that validates before apl-operator hard-fails
**Describe the bug:**

Some `apps.*` fields are `x-secret`s **with a generator** (e.g. `apps.loki.adminPassword`, `x-secret: '{{ randAlphaNum 20 }}'`) yet are simultaneously marked **`required`** in the helm chart's `values.schema.json`. The generator that would fill them only runs inside otomi's own bootstrap pipeline. So any install path that validates the JSON schema *before* apl-operator runs — including a plain `helm install apl/apl` — fails schema validation on a field the platform was supposed to generate.

This is a latent trap for **every** generator-backed x-secret, not just Loki; Loki is just where we hit it first because we drive a raw `helm install` in bootstrap.

**To Reproduce**

1. Do a raw `helm install apl/apl -f values.yaml` (no otomi bootstrap in front) and omit `apps.loki.adminPassword` — on the theory that its x-secret generator fills it.
2. Install fails immediately:
   ```
   Error: values don't meet the specifications of the schema(s):
   apps.loki: adminPassword is required
   ```
3. The `randAlphaNum 20` generator never ran — it only executes inside otomi's own bootstrap, which a direct `helm install` bypasses.

**Expected behavior:**

A field the platform can generate should not be `required` in the schema that gates a supported install path. Either (a) drop `required` for x-secret fields that carry a generator, or (b) run the generator (or a schema default) early enough that every documented install path — including `helm install` — passes validation. As-is, the "generator" and "required" flags contradict each other depending on how you install.

**Cluster(s):** Linode / Akamai LKE-Enterprise (any fresh cluster; failure is at `helm install`, pre-cluster-state).

**Versions:** apl-core `6.0.0` (repro'd through `v6.0.0-rc.12` → GA).

**Additional context:**

We worked around it by supplying `apps.loki.adminPassword` ourselves from a Terraform `random_password` rendered into the values, i.e. we re-implemented the generator outside apl-core purely to satisfy schema validation. Worth auditing every generator-backed x-secret in `values-schema.yaml` for the same `required` contradiction.

aweingarten:
## 📌 Summary

Fixes #3422.

`apps.loki.adminPassword` is an x-secret with a generator (`x-secret: '{{ randAlphaNum 20 }}'`) and is at the same time listed in its parent's `required` array. The generator only runs inside the apl-operator bootstrap pipeline, so any install path that validates values-schema *before* that point — a plain `helm install apl/apl` included — hard-fails on a field the platform was about to fill in:

```
Error: values don't meet the specifications of the schema(s):
apps.loki: adminPassword is required
```

The two flags contradict each other depending on how you install. This drops the `required` entry so the generator can do its job, and adds a test asserting the invariant across the whole schema: **a secret with a non-empty `x-secret` generator is never `required`.**

## 🔍 Reviewer Notes

- I walked every `required` list in `values-schema.yaml` against the corresponding property. `apps.loki.adminPassword` is the **only** field that violates this today — all twelve other required x-secrets are `x-secret: ''` (operator-supplied), which is correct and unaffected. So the behavioural change is limited to Loki; the test is what keeps the class closed.
- The test walks the raw schema rather than the dereferenced one, so it also covers `required` lists inside `definitions`/`oneOf` branches.
- Verified the test is not vacuous: it fails on `main` (reporting `apps.loki.adminPassword`) and passes with the schema change.
- Downstream we worked around this by generating the Loki password ourselves in Terraform and rendering it into the values, i.e. re-implementing apl-core's generator outside apl-core purely to satisfy schema validation. This removes that need.

Verified locally: `jest`, `eslint`, `tsc --noEmit` all clean. The helmfile-backed gates (`validate-values`, `validate-templates`, `lint:hf`) need `helmfile`/`gucci`, which I don't have locally — leaving those to CI.

## 🧹 Checklist

- [x] Unit tests added/updated


CasLubbers (approved):
Nice find! This is indeed regression and should be removed. It is already covered in the apl-operator

PR #3466, opened 2026-07-28
feat(obj): source linode accessKeyId from obj-secrets

### LINKED ISSUE #3459: Object storage (linode): source accessKeyId from obj-secrets like secretAccessKey, so key rotation needs no git write
### Summary

The `linode` object-storage provider sources the two halves of the S3 credential asymmetrically: `secretAccessKey` is read from the `obj-secrets` Secret (great — an operator can populate it via ESO and rotate freely), but `accessKeyId` is **inlined from git settings**. Because Linode OBJ keys rotate as a *pair*, this forces any operator that rotates keys to write the new `accessKeyId` back into the values git repo on every rotation — carrying a non-secret identifier through git solely because its paired secret is handled cleanly.

Request: let `accessKeyId` also be sourced from `obj-secrets` (or a native Secret), symmetric with `secretAccessKey`.

### Current behavior (apl-core v6.0.0)

From the templating:

`values/loki/loki-raw.gotmpl`
```gotmpl
AWS_ACCESS_KEY_ID:     {{ $obj.linode.accessKeyId }}                 # inlined from env/settings/obj.yaml (git)
AWS_SECRET_ACCESS_KEY: '{{ "{{ .secretAccessKey | toString }}" }}'
  - secretKey: secretAccessKey
    remoteRef:
      key: obj-secrets                                              # from the obj-secrets Secret (ESO-friendly)
      property: provider_linode_secretAccessKey
```

`values/harbor/harbor-raw.gotmpl` does the same (`S3_STORAGE_ACCOUNT` / `REGISTRY_STORAGE_S3_ACCESSKEY` ← `{{ $obj.linode.accessKeyId }}`; secret ← `obj-secrets.provider_linode_secretAccessKey`).

And `env/settings/obj.yaml` (`AplObjectStorage`) carries `spec.provider.linode.accessKeyId` as a plain setting (schema `additionalProperties: false`, so the secret half is deliberately kept out).

### Why it's a problem

A managed-App-Platform operator that provisions **and rotates** the OBJ access keys out-of-band (our case: Linode Landing Zone) can deliver `secretAccessKey` into `obj-secrets` via ESO/OpenBao and rotate with **zero git writes** — exactly right. But because `accessKeyId` is git-inlined and a rotation mints a *new pair*, the operator must also push the new `accessKeyId` into the values repo each rotation to keep it paired. That forces a git-sync reconciler holding a `Contents: write` token — a lot of moving parts to carry a **non-secret identifier that changes on the same cadence as the secret already handled cleanly**.

### Proposed options

1. **Extend the `obj-secrets` contract**: add a `provider_linode_accessKeyId` property and template `AWS_ACCESS_KEY_ID` from it (mirroring `secretAccessKey`). Operators populate one Secret with both keys via ESO; no git writes on rotation.
2. **Optional secret ref**: add `spec.provider.linode.accessKeyIdSecretRef` (nativeSecrets-style). When set, source `accessKeyId` from the referenced Secret; otherwise fall back to the inline settings value (fully backward compatible).

Either makes the two halves of the credential rotate together through one ESO-managed Secret, and lets external operators drop the git-write token + reconciler they otherwise need only for the access-key id.

### Environment

- apl-core `v6.0.0` (managed App Platform / apl-operator `1.16.0`)
- Provider: `linode`; consumers observed: Loki, Harbor (same pattern in gitea/keycloak/kubeflow-pipelines `-raw.gotmpl`)

aweingarten:
## 📌 Summary

Fixes #3459 — implements the direction @j-zimnowoda agreed with there ("it makes perfect sense to keep the id and token all together").

The two halves of the Linode object-storage credential were sourced asymmetrically: `secretAccessKey` came from the `obj-secrets` Secret (ESO-friendly, rotatable without touching git), while `accessKeyId` was inlined into every consumer's ExternalSecret template straight from git settings.

Linode OBJ keys rotate as a **pair**, so an operator that provisions and rotates them out-of-band had to push the new `accessKeyId` back into the values repo on every rotation — carrying a non-secret identifier through git purely because its paired secret is already handled cleanly, and standing up a git-sync reconciler with a `Contents: write` token to do it.

This is option 1 from the issue:

- `values-schema.yaml`: `obj.provider.linode.accessKeyId` gains `x-secret: ''`, so it flows into `obj-secrets` as `provider_linode_accessKeyId` through the existing sealed-secret machinery — the same route `secretAccessKey` already takes.
- All six consumers template it from the ExternalSecret instead of inlining it: loki, harbor (×2), gitea, keycloak, kubeflow-pipelines.

Rotation becomes: write one Secret. No git writes at all.

## 🔍 Reviewer Notes

- **No new contract.** The key name (`provider_linode_accessKeyId`) is derived by the existing `buildSecretToNamespaceMap` path→underscore rule, not hardcoded — same rule that produces `provider_linode_secretAccessKey`.
- **`required` is unchanged.** `accessKeyId` stays in the `required` list next to `secretAccessKey`; this changes *where* the value is stored, not whether it must be supplied.
- **Upgrade path — worth a second opinion.** `generateSecrets()` merges input values over generated ones and then picks by schema x-secret path, so an `accessKeyId` arriving in `VALUES_INPUT` is picked up and sealed on the next bootstrap, and `stripAllSecrets` drops it from the settings written to disk. That self-migrates installs where the value comes from the helm values. If you know of a supported topology where `accessKeyId` lives **only** in the values repo and never in `VALUES_INPUT`/stored cluster secrets, that case would want a `values-changes.yaml` entry to relocate it — happy to add one, I just couldn't find such a path.
- Fixtures were updated to match the new classification: `accessKeyId` moves out of `tests/fixtures/env/settings/obj.yaml` and into the `obj-secrets` sealed-secret fixture, mirroring `secretAccessKey`.
- Consumers outside this repo that read `obj.provider.linode.accessKeyId` from settings (console/API) would need the same treatment — flagging in case that's in scope for you.

Verified locally: `jest` (including `sealed-secrets` and `bootstrap` suites), `eslint`, `tsc --noEmit` all clean. The helmfile-backed gates (`validate-values`, `validate-templates`, `lint:hf`) need `helmfile`/`gucci`, which I don't have locally — leaving those to CI.

## 🧹 Checklist

- [x] Unit tests added/updated


aweingarten:
No reviewer findings on this one across three Copilot passes, and nothing from me to change — flagging it as ready whenever someone has time.

For a reviewer, the thing worth checking is the symmetry claim: `accessKeyId` and `secretAccessKey` are now both sourced from the `obj-secrets` Secret via ExternalSecrets, so rotating Linode object-storage credentials is a single-Secret write rather than a Secret write plus a git-backed settings edit. The `x-secret: ''` marking in `values-schema.yaml` is what routes `accessKeyId` through the existing sealed-secret pipeline; `src/common/obj-secret-paths.test.ts` asserts both halves are discovered as secret schema paths.

All six Linode OBJ consumers were updated together — Loki, Harbor (x2), Gitea, Keycloak, Kubeflow Pipelines — so there is no consumer left reading the old inline value.

merll:
While I generally agree with this change, the step which migrates all `x-secret` values to `SealedSecret` resources is only called during installation or an upgrade prior to 6.x. The utility functions there are currently not suitable for incrementally moving more into secrets.

As of now, the platform loses the `accessKeyId` value. I can look into how to mitigate that during the following week.
For fresh installs this is fine and certainly more consistent than the current state.

merll:
I resolved merge conflicts in this branch: https://github.com/linode/apl-core/tree/feat/obj-accesskeyid-from-obj-secrets-2.
Would you prefer to merge them or change the PR to that branch?

merll:
Regarding the migration, I realized this cannot be done without major complications. Inserting 'x-secret' into the values schema makes the value disappear before the migration even sees it. Neither can it be re-read from the filesystem, as the working copy is already modified.
This is pretty much an edge case that I would prefer to solve in release notes, given that the access key can be recovered also from the Cloud Manager / API as it is not a show-once value like the actual secret key. @j-zimnowoda what do you think?

Issue #3420, opened 2026-07-06
Internal ServiceEntry hosts (<svc>.<domainSuffix>) don't resolve for sidecar-less pods / init containers


aweingarten:
**Is your feature request related to a problem? Please describe.**

apl-core exposes several internal hosts (keycloak, harbor, argocd, auth, api, console, …) as Istio `ServiceEntry`s under `<svc>.<domainSuffix>`. These resolve fine for mesh/sidecar'd pods, but **sidecar-less pods and init containers can't resolve them** — there's no in-cluster DNS record backing the ServiceEntry host. apl-core's own components hit this: the oauth2-proxy `wait-for-keycloak` init container is sidecar-less by definition and polls `https://keycloak.<domainSuffix>/realms/otomi` at boot, but the name doesn't resolve, so it can't even reach the CA/TLS failure — it fails at DNS.

**Describe the solution you'd like**

Ship in-cluster DNS resolution for the internal ServiceEntry hosts apl-core defines, so sidecar-less/init workloads resolve them out of the box — e.g. a CoreDNS rewrite/hosts entry for `*.<domainSuffix>` (or real ClusterIP Services) generated alongside the ServiceEntries. apl-core is best positioned to know the full host set (it defines the ServiceEntries).

**Describe alternatives you've considered**

We maintain a CoreDNS `*.internal` rewrite downstream so sidecar-less/init consumers can resolve `<svc>.<domainSuffix>`. It works but couples us to apl-core's host set and domainSuffix templating, and every adopter re-discovers it as an opaque "init container loops forever" failure.

**Additional context**

This is the DNS-resolution counterpart to the CA-trust half of #3388 (the oauth2-proxy `wait-for-keycloak` init container also can't trust the custom-CA cert). The two are independent: even with the CA fixed, the host still has to resolve. Fixing both is what lets apl-core's own init containers reach Ready unaided.

PR #3467, opened 2026-07-28
fix(oauth2-proxy): poll keycloak over the backchannel in wait-for-keycloak

### LINKED ISSUE #3388: Charts hardcode values that can't be overridden via apl-values
# Is your feature request related to a problem? Please describe.

Several apl-core components bake in configuration that adopters cannot override through the supported apl-values / _rawValues path. When the hardcoded value is wrong for a given platform or deployment topology, there is no supported escape hatch — the only way to fix it is to mutate the rendered Kubernetes objects at admission time (Kyverno) or to keep dummy components alive. This turns routine configuration into invasive, fragile workarounds that re-break whenever apl-core changes its internal rendering. The specific cases we've hit:

- gitops-global is hardwired to clone the in-cluster Gitea and ignores otomi.git. We run a GitHub-external-repo-only topology with in-cluster Gitea disabled, but gitops-global hard-fails with gitea-http … no such host when Gitea is off. There is no knob to point it at otomi.git.
- oauth2-proxy's wait-for-keycloak init container has no Otomi CA and can't be configured. It's templated as vanilla curlimages/curl and runs curl https://keycloak.primary.internal against the in-cluster Otomi-signed wildcard cert, exits 60 (self-signed certificate in certificate chain), and loops forever. No chart value exposes the init image, a CA volume mount, or the curl args.
- Loki's schema object_store can't be set via values. The chart defaults to object_store: filesystem and doesn't derive it from loki.storage.type. Overriding loki.schemaConfig through _rawValues fails because apl-operator's value pipeline round-trips YAML through mixed libraries and corrupts the date-only from field (YAML 1.1 timestamp coercion), which breaks Loki's parser. So even the "supported" override path is unusable for this field.

(This is the same class of problem as the PVC-encryption issue, where gitea-valkey and oauth2-proxy redis-HA hardcode storageClass: linode-block-storage.)

# Describe the solution you'd like

Every value that can legitimately vary by platform or topology should be overridable through apl-values without resorting to admission-time mutation. Concretely:

- gitops-global should honor otomi.git (or expose an explicit repo-URL override) so external-repo-only deployments work with in-cluster Gitea disabled.
- oauth2-proxy's init container should expose its image and a CA-bundle mount (or accept the Otomi root CA by default) so its readiness probe trusts Otomi-issued certs.
- Loki's object_store / schemaConfig should be safely settable via values — which also requires fixing the value-pipeline date-field corruption so _rawValues overrides round-trip losslessly.

More broadly: where a chart value is hardcoded, prefer a defaulted-but-overridable value over a baked-in literal.

# Describe alternatives you've considered

We compensate downstream, and each workaround is invasive:

- gitops-global: keep apps.gitea.enabled=true as an unused dummy purely so the host resolves, and Terraform-patch the argocd namespace name=argocd label so apl-core's gitea NetworkPolicy matches (see apl-values/_shared/values.yaml and cluster-bootstrap/main.tf).
- oauth2-proxy init: a Kyverno ClusterPolicy injects -k into the curl invocation (kyverno-oauth2-proxy-wait-keycloak-trust.yaml).
- Loki object_store: a Kyverno ClusterPolicy rewrites the rendered ConfigMap object_store: filesystem → s3 (kyverno-loki-s3-object-store.yaml), because the values path corrupts the date field.

All three couple us to apl-core's internal rendering (object names, labels, ConfigMap shape) and the Kyverno ones are admission-timing-sensitive: if the webhook lags the resource creation, the fix is missed.

# Describe alternatives you've considered (cont.) / Additional context

- Files: instance-template/.../manifests/kyverno-oauth2-proxy-wait-keycloak-trust.yaml, .../kyverno-loki-s3-object-store.yaml, apl-values/_shared/values.yaml, cluster-bootstrap/main.tf.
- The Loki case is a two-part upstream ask: (a) expose object_store safely, and (b) fix the apl-operator value pipeline's YAML 1.1 date coercion, which is a latent corruption risk for any date-shaped _rawValues field, not just Loki.
- Can be filed as one umbrella issue ("hardcoded chart values should be overridable via apl-values") with the three cases as a checklist, or split per-component.

aweingarten:
## 📌 Summary

Fixes #3388 — the oauth2-proxy item (and, for this container, the resolution failure in #3420).

> **Note on scope:** #3388 lists three independent items; this PR implements the oauth2-proxy one. The other two (gitops-global ignoring `otomi.git`, and Loki's `object_store`/`schemaConfig` being unsettable through `_rawValues`) are untouched. Since `Fixes` will auto-close #3388 on merge, those two should be split into their own issue — happy to open it, or leave this as a plain reference instead if you'd rather keep #3388 open until all three land.

The `wait-for-keycloak` init container polls the **public** issuer URL with a vanilla `curlimages/curl`:

```
args: ["while [ $(curl -sw '%{http_code}' https://keycloak.<domainSuffix>/realms/otomi -o /dev/null) -ne 200 ]; ..."]
```

The oauth2-proxy pod is explicitly sidecar-less (`sidecar.istio.io/inject: "false"`), so that URL is reached without the mesh, and it hits one of two walls depending on the deployment:

- the host is an Istio ServiceEntry with no in-cluster DNS record backing it, so it may not resolve at all (#3420);
- when it does resolve, the Otomi-signed wildcard cert isn't trusted by a stock curl image, so curl exits 60 (`self-signed certificate in certificate chain`).

Either way the container loops forever and oauth2-proxy never starts. No chart value exposes the init image, a CA mount, or the curl args, so there's no supported way out.

Worth noting the asymmetry this creates: oauth2-proxy's own config already tolerates the untrusted CA — `ssl_insecure_skip_verify = {{ $v._derived.untrustedCA }}` — so **the gate is currently stricter than the app it gates**.

The fix points the poll at `_derived.oidcBaseUrlBackchannel` (`http://keycloak-keycloakx-http.keycloak:8080/realms/otomi`), which already exists for exactly this purpose and is used by kubernetes-gateways for its JWKS endpoint. Plain HTTP to a ClusterIP Service: nothing to resolve through the mesh, nothing to trust.

## 🔍 Reviewer Notes

- **The gate condition is unchanged.** Both URLs return 200 only once the `otomi` realm is being served; only the route changes. It does not verify the public ingress path — but the init container never usefully verified that either, since it fails on DNS/TLS before reaching Keycloak.
- I picked this over the alternative (mount the platform CA and pass `--cacert`, following the `custom-ca` Secret pattern in argocd-raw/gitea-raw). Mounting a CA fixes the TLS half but leaves the DNS half, and it adds a Secret + volume + mount where a value that already exists does the whole job. Happy to switch if you'd rather the gate stay on the public URL.
- The other two items in #3388 (gitops-global ignoring `otomi.git`; Loki `object_store`/`schemaConfig` not settable through `_rawValues`) are untouched here — they're independent and larger. Splitting that issue may be worthwhile.
- The init container image is still `curlimages/curl:latest` with no value to override it. That part of the #3388 ask stands; it seemed better as its own change.

Template-only change. `validate-templates`/`lint:hf` need `helmfile`, which I don't have locally — leaving those to CI.

## 🧹 Checklist

- [ ] Unit tests added/updated — n/a, this is a one-line template change covered by `validate-templates`


CasLubbers:
The public URL check here isn't just "is Keycloak up", its an end-to-end smoke test (DNS, cert, ingress) gating oauth2-proxy, which is the platform's SSO entrypoint. Switching to oidcBaseUrlBackchannel fixes the sidecar-less DNS/TLS failure, but it also means the gate now passes even when the public path is broken. That defeats the point of gating on it in the first place.

If the goal is just don't crashloop forever, maybe worth keeping the public URL but adding CA trust + in-mesh DNS path (mount platform CA, skip the ServiceEntry issue) instead of swapping to backchannel. Otherwise, we lose the one check that verifies real users can actually reach Keycloak.

Alternative: make this configurable instead of picking one. Add a value (default true = current public-url behavior) so people who want the real e2e check keep it, and people hitting the DNS/ServiceEntry issue can opt into backchannel. Still need the CA trust fix for the public path though, otherwise the flag is pointless for anyone who leaves it on.

aweingarten:
You are right, and I have reverted the URL swap. Routing the gate around the public path defeats what the gate is for.

Chasing your CA-trust suggestion turned up the actual root cause, which is narrower than what I originally wrote in the description:

- The pod sets `sidecar.istio.io/inject: "false"` — that is the **whole** oauth2-proxy pod, not just this init container. So the main container is in the same position: no sidecar, no mesh DNS, no sidecar-terminated TLS.
- The main container copes because it is handed `ssl_insecure_skip_verify = {{ $v._derived.untrustedCA }}` a few lines up in the same file. The init `curl` had no equivalent.
- `_derived.untrustedCA` is true when the issuer is `custom-ca` **or** letsencrypt-staging. `custom-ca` is the default issuer, so on an ordinary install that curl is verifying a platform-issued cert with no trust for it and loops forever at a TLS error, never reaching a status code at all.

So this was never really a DNS or ServiceEntry problem — my original description overreached there. It is one container missing the trust decision its sibling already makes.

The fix now keeps `_derived.oidcBaseUrl` and adds `-k` on **exactly the same condition** as `ssl_insecure_skip_verify`:

```
curl -s{{ if $v._derived.untrustedCA }}k{{ end }}w '%{http_code}' {{ $v._derived.oidcBaseUrl }}
```

That keeps your end-to-end property — DNS, ingress and the route are all still exercised, and the gate still fails when real users cannot reach Keycloak — while making the init container no more and no less permissive about certificates than the process it is gating for. It also means no new config knob, which I think is better than the toggle option you floated: a flag defaulting to the public URL would leave the default install still broken.

If you would rather have the CA mounted and verified properly instead of `-k`, I am happy to do that — but note it would then be *stricter* than the main container, which skips verification outright in this mode, so the init container could block on a cert that oauth2-proxy itself would have accepted. Matching the main container seemed like the more honest gate.

merll:
The unstrustedCA functionality has mostly been used for testing (e.g. using LetsEncrypt staging). Due to limitations of dependencies it could not be implemented consistently throughout the platform (e.g. Harbor). While there might be some demand to have better support for custom CAs, skipping verification for testing purposes is just fine. Adding this to the Keycloak polling is a good idea.

values/oauth2-proxy/oauth2-proxy.gotmpl
```diff
@@ -82,7 +82,16 @@ extraInitContainers:
...
+    # Deliberately polls the public issuer, not the backchannel: this gate is an end-to-end check
+    # that DNS, ingress and the certificate chain are all serving Keycloak, which is what oauth2-proxy
+    # needs to work for real users.
+    #
+    # `-k` mirrors the `ssl_insecure_skip_verify` the main container is already given on the same
+    # condition. Without it the curl fails cert verification forever against a platform-issued cert
+    # — and `custom-ca` is the default issuer, so that is the ordinary install, not an edge case.
+    # This container has no istio sidecar to terminate TLS for it (the whole pod sets
+    # sidecar.istio.io/inject=false), so it has to make the same trust decision itself.
+    args: ["while [ $(curl -s{{ if $v._derived.untrustedCA }}k{{ end }}w '%{http_code}' {{ $v._derived.oidcBaseUrl }} -o /dev/null) -ne 200 ]; do sleep 2; echo 'Waiting for Keycloak OIDC Issuer URL'; done"]
```
merll: Could we make this a template comment instead?
Keeping the rendered values short has several advantages, e.g. making ad-hoc editing easier.
```suggestion
    {{- /*
    Deliberately polls the public issuer, not the backchannel: this gate is an end-to-end check
    that DNS, ingress and the certificate chain are all serving Keycloak, which is what oauth2-proxy
    needs to work for real users.

    `-k` mirrors the `ssl_insecure_skip_verify` the main container is already given on the same
    condition. Without it the curl fails cert verification forever against a platform-issued cert
    — and `custom-ca` is the default issuer, so that is the ordinary install, not an edge case.
    This container has no istio sidecar to terminate TLS for it (the whole pod sets
    sidecar.istio.io/inject=false), so it has to make the same trust decision itself.
    */}}   
```

merll (approved):

CasLubbers (approved):

PR #3468, opened 2026-07-28
feat(chart): allow the platform namespaces to be created out-of-band

### LINKED ISSUE #3418: apl chart ships an annotation-less 00-namespace.yaml, so helm cannot adopt a pre-existing namespace
**Describe the bug:**

The `apl` chart renders its namespace via a bare `00-namespace.yaml` with no helm ownership metadata. When that namespace already exists — which is common, because bootstrap Terraform frequently must create it first to land CRDs, StorageClasses, or NetworkPolicies ahead of the operator — helm refuses to adopt it and the release collides.

**To Reproduce**

1. Pre-create the target namespace (e.g. via Terraform) before installing the `apl` chart.
2. `helm install apl/apl ...`
3. Install fails with an ownership conflict:
   ```
   Error: rendered manifests contain a resource that already exists.
   Unable to continue with install: Namespace "<ns>" ... exists and cannot be
   imported into the current release: invalid ownership metadata; ... missing key
   "app.kubernetes.io/managed-by"; missing key "meta.helm.sh/release-name" ...
   ```

**Expected behavior:**

Either add helm ownership metadata to the namespace manifest (`app.kubernetes.io/managed-by: Helm`, `meta.helm.sh/release-name`, `meta.helm.sh/release-namespace`) so a pre-existing namespace can be adopted, or gate namespace creation so integrators who create it out-of-band aren't blocked. Bootstrap ordering frequently requires the namespace to exist before the operator install.

**Cluster(s):** Linode / Akamai LKE-Enterprise.

**Versions:** apl-core `6.0.0` (`00-namespace.yaml` still annotation-less on the v6 line).

**Additional context:**

We manage the collision downstream in cluster-bootstrap Terraform; native helm-adoptable ownership metadata would remove the workaround entirely.

aweingarten:
## 📌 Summary

Fixes #3418.

The chart renders `apl-operator`, `otomi` and `apl-secrets` from a bare `00-namespace.yaml`. When any of them already exists, the install fails outright:

```
Error: rendered manifests contain a resource that already exists.
Unable to continue with install: Namespace "<ns>" ... exists and cannot be
imported into the current release: invalid ownership metadata; ... missing key
"app.kubernetes.io/managed-by"; missing key "meta.helm.sh/release-name" ...
```

That's a common position to be in: bootstrap tooling frequently has to create a namespace *before* the operator installs, to land CRDs, StorageClasses or NetworkPolicies first.

Adds `installation.createNamespaces` (default `true`, so nothing changes for existing installs). Set it to `false` and the chart leaves the namespaces alone.

## 🔍 Reviewer Notes

- **This is deliberately not "add helm ownership metadata to the manifest"**, which is what the issue originally asked for. Helm validates ownership against the **live** object, so annotating the template can't make an already-existing namespace adoptable — it would only label namespaces the chart created itself, which were never the problem. Gating creation is what actually unblocks the integrator. Worth correcting in the issue too.
- The values comment spells out the labels an out-of-band creator has to set (`kubernetes.io/metadata.name`, `name`), because the platform's NetworkPolicies select namespaces by the `name` label — dropping it fails in a way that's hard to trace back here.
- Placed under `installation.` alongside `skipOperatorDeployment`, which is the same shape of knob. `installation` is `additionalProperties: true` in values-schema, so the new key validates; I confirmed `helm lint` passes against the generated `values.schema.json` both with the flag set and unset.
- Verified `helm template` renders 3 Namespaces by default and 0 with `--set installation.createNamespaces=false`.

## 🧹 Checklist

- [ ] Unit tests added/updated — n/a, chart-template change; covered by `helm lint`/`helm template` in `lint:apl-chart`


chart/apl/values.yaml
```diff
@@ -136,3 +136,12 @@ installation:
   mode: standard
   # Skip deployment of the operator
   skipOperatorDeployment: false
+  ## Create the namespaces the platform needs (apl-operator, otomi, apl-secrets).
```
CasLubbers: This comment is way to verbose. This is more clear and concise:
`## Create namespaces (apl-operator, otomi, apl-secrets). Disable if creating them yourself.`

CasLubbers (changes requested):
The flag is good, if you could only change the comment.

chart/apl/values.yaml
aweingarten: Taken, with two short lines kept beside it. Now three lines instead of eight:

```
## Create namespaces (apl-operator, otomi, apl-secrets). Disable if creating them yourself.
## Label them `kubernetes.io/metadata.name: <name>` and `name: <name>` — NetworkPolicies match on `name`.
## Set this at install time; flipping it to false later makes helm delete the namespaces.
```

The label line is there because disabling the flag without those labels silently breaks the platform NetworkPolicies, which select namespaces by `name`. The third line is the footgun Copilot flagged below. Happy to drop either if you would rather have just your one line.

chart/apl/values.yaml
```diff
@@ -136,3 +136,12 @@ installation:
...
+  ## Set to false when they are created out-of-band — bootstrap tooling often has to create them
+  ## first to land CRDs, StorageClasses or NetworkPolicies ahead of the operator, and helm refuses
+  ## to adopt a namespace it does not already own ("invalid ownership metadata").
+  ## When disabling this, create the namespaces with the labels this chart would have set:
+  ##   kubernetes.io/metadata.name: <name>
+  ##   name: <name>
+  ## The platform's NetworkPolicies select namespaces by the `name` label.
+  createNamespaces: true
```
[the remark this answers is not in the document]
aweingarten: Agreed, and worth stating in the file rather than only in review — flipping this to false on an existing release makes helm see the Namespaces as removed from the chart and try to delete them, which cascades through `otomi` and `apl-secrets`. Added as a one-liner: `## Set this at install time; flipping it to false later makes helm delete the namespaces.`

aweingarten:
Re: your CHANGES_REQUESTED review — comment rewritten, down from eight lines to three:

```
## Create namespaces (apl-operator, otomi, apl-secrets). Disable if creating them yourself.
## Label them `kubernetes.io/metadata.name: <name>` and `name: <name>` — NetworkPolicies match on `name`.
## Set this at install time; flipping it to false later makes helm delete the namespaces.
```

Line 1 is yours verbatim. I kept two more because both are silent failures rather than prose:

- **The labels.** Disabling the flag and creating the namespaces without those labels leaves the platform NetworkPolicies matching nothing — they select namespaces by `name`, and nothing errors, it just quietly does not apply.
- **The one-way toggle.** Flipping this to false on an existing release makes helm see the Namespaces as removed from the chart and delete them, cascading through `otomi` and `apl-secrets`. Copilot flagged this too.

If you would rather have only your single line, say so and I will cut the other two — no argument from me, they are your charts.

No code change; the template conditional is unchanged and still renders 3 namespaces with the flag on and 0 with it off.

CasLubbers (approved):

Issue #3389, opened 2026-06-25
openbao + ESO as a feature for APL


aweingarten:
## Is your feature request related to a problem? Please describe.

apl-core ships Argo CD, cert-manager, Istio, Harbor, Loki, and other platform components, but it does not ship a secrets-management stack — neither a Vault/OpenBao backend nor the External Secrets Operator (ESO) that consumes it. apl-core 5.0.0 has no ESO chart and no OpenBao integration at all. Yet nearly every real deployment needs a way to inject secrets into platform and tenant workloads, so each adopter has to bolt on their own secrets stack and wire it into apl-core's bootstrap by hand.

We've built this downstream, and it's a large, fragile surface that every adopter independently re-pays:

- *Install + wire two components from scratch* — OpenBao and ESO as separate Argo Applications at hand-tuned sync-waves, pinned versions (ESO 0.10.7 for a CA-revalidation fix), and a second platform-support AppProject because apl-core only seeds a git-sourced platform-bootstrap project.
Bootstrap orchestration — OpenBao starts sealed; a GitHub Actions workflow (bootstrap-openbao.yml) has to unseal it and seed the ESO AppRole before any ExternalSecret can resolve. That human-in-the-loop step is the root of a long not-ready window on first boot.
- Cert lifecycle glue — OpenBao loads its TLS cert only at startup (no SIGHUP reload), so we run a custom openbao-cert-watcher Deployment that polls the cert and bounces the StatefulSet on rotation (pods come back sealed, needing re-unseal).
NetworkPolicy gaps — ESO's webhook needs apiserver ingress on 10250 under apl-core's default-deny (see the related NetworkPolicy issue).
- Argo tolerance hacks — because the bootstrap is legitimately not-ready for a while, we override Argo health checks (ESO not-ready → Progressing, not Degraded), inflate retry budgets, and run an argo-resync-nudger CronJob to recover failed syncs.

None of this is exotic — it's the baseline "how do secrets get into workloads" problem. Because apl-core leaves it out, every adopter reinvents the same ~6 moving parts, and gets to discover each failure mode (sealed-on-boot, cert-rotation breakage, webhook NP gaps, Argo wedging) the hard way.

## Describe the solution you'd like

apl-core should ship OpenBao + ESO as first-class, integrated platform components, the same way it ships cert-manager or Istio — installable via an apps.* toggle and wired into the platform out of the box. Concretely:

[...]

Issue #3387, opened 2026-06-25
apl-core Default NetworkPolicies


aweingarten:
# Is your feature request related to a problem? Please describe.

apl-core ships default-deny NetworkPolicies across the namespaces it manages, but does not ship the corresponding allow-rules that its own bundled components require to function. The result: with default-deny in effect, components apl-core itself deploys silently fail to converge, usually with cryptic, non-obvious symptoms (timeouts, FailedDiscoveryCheck, crashloops) that give no hint the cause is a missing NetworkPolicy. Every adopter running default-deny has to independently rediscover each gap and hand-author the missing rules.

We currently maintain ~20 such allow-rules downstream (in our llz-cluster-foundation chart and per-component values) just to make apl-core's own stack work behind its own default-deny posture. Concretely, the recurring gaps we've had to fill:

- ESO webhook ingress — the apiserver can't reach the External Secrets validating/mutating webhook (port 10250), so every ClusterSecretStore/ExternalSecret create/update mutation times out.
- cert-manager-webhook-linode aggregated APIService — the apiserver's discovery probe to the webhook (:443) is dropped, leaving the APIService Available=False (FailedDiscoveryCheck) and convergence hung. The webhook pod is labeled app=cert-manager-webhook-linode (no app.kubernetes.io/name), so apl-core's generic webhook NP doesn't even match it.
- CNPG operator → Harbor DB — the CloudNativePG operator can't poll instance status (8000) or metrics (9187) across namespaces, so the Harbor Postgres replica never starts and the stack stalls (Instance Status Extraction Error → ClusterIsNotReady → Degraded).
- Istio sidecar egress to istiod — sidecars can't reach istiod (15010/15012/15014) for xDS and mTLS cert signing. The iptables redirect still installs, so when Envoy dies it takes all pod egress with it (including kube-apiserver), surfacing as cryptic "connection refused" everywhere.
- LKE-Enterprise apiserver egress — on LKE-E the in-cluster kubernetes Service maps 443→targetPort 6443, and Cilium evaluates policy after kube-proxy-replacement DNAT, so rules allowing only 443 are silently dropped. Every NP that talks to the apiserver must also allow 6443 or all API calls hang.

These are not exotic edge cases — they're apl-core's own first-party components (ESO, cert-manager DNS-01 webhook, CNPG/Harbor, Istio) failing under apl-core's own default-deny NetworkPolicies, on a supported platform (LKE / LKE-Enterprise).

# Describe the solution you'd like

[...]

PR #3470, opened 2026-07-28
feat(netpols): ship the ESO webhook allow-rule

aweingarten:
## 📌 Summary

First of #3387 — and the specific rule @j-zimnowoda asked for in #3389 ("Could you provide an netpol spec that is missing to support ESO's webhook?").

apl-core ships default-deny NetworkPolicies for the namespaces it manages, but not the allow-rules its own bundled components need. Under default-deny those components fail with symptoms that give no hint a NetworkPolicy is the cause. External Secrets is the clearest case: the apiserver can't reach the validating/conversion webhook on **10250**, so every `ClusterSecretStore`/`ExternalSecret` create or update times out against the webhook and the secrets stack never converges.

This adds an opt-in policy for the ESO webhook, following the per-app pattern already in `apl-network-policies`:

- `netpols.externalSecrets`, driven by `apps.external-secrets.networkPolicies.enabled` (new, via the existing `appNetworkPolicyConfig` definition)
- ingress on 10250 **scoped by port, not by source** — see below
- ingress on 8080 from `po-prometheus`, so turning the policy on doesn't silently cut metrics scraping

## 🔍 Reviewer Notes

- **Why the webhook rule has no `from`.** The apiserver is off the pod network on a managed control plane (LKE/LKE-E) and has no address a `namespaceSelector` or `ipBlock` can pin down, so an ingress rule with `ports` only is the portable way to express "let the apiserver in". If you'd rather pin it to the LKE control-plane range, say which and I'll switch — I didn't want to bake an assumption about the address space into the chart.
- **Default is `false`, deliberately.** The policy is pod-scoped and `Ingress`-typed, so creating it makes the webhook pods deny any ingress not listed. That's only the intent on a cluster already running default-deny; on a cluster without it, enabling this *tightens* rather than loosens. `appNetworkPolicyConfig` already defaults `enabled` to false.
- **Selector verified against the rendered chart**, not assumed: `helm template charts/external-secrets` produces pod labels `app.kubernetes.io/name: external-secrets-webhook` and container ports 10250 (webhook), 8080 (metrics), 8081 (ready) in namespace `external-secrets`.
- **Probe port 8081 is deliberately not allowed.** Kubelet→pod probe traffic isn't subject to NetworkPolicy under Cilium on LKE, and the existing gitea/otomi-api policies don't allow it either — I followed the convention rather than diverge. Flagging it since it would matter under a CNI that does enforce host→pod.
- Verified `helm template` renders the policy with the flag on and nothing with it off, and `helm lint` passes both ways.

## Remaining #3387 gaps

Keeping this to one component so it's reviewable. The other rules from the issue are each their own PR, and I'd rather land them one at a time than in a batch:

1. **cert-manager-webhook-linode** aggregated APIService — also needs a selector fix: that pod is labeled `app=cert-manager-webhook-linode` with no `app.kubernetes.io/name`, so the chart's generic webhook policy (`charts/cert-manager/templates/networkpolicy-webhooks.yaml`) doesn't match it at all.
2. **CNPG operator → Harbor DB** (8000 instance status, 9187 metrics), cross-namespace.
3. **Istio sidecar egress to istiod** (15010/15012/15014) — the nasty one, since a dead Envoy takes all pod egress with it.
4. **LKE-E apiserver egress on 6443** — cross-cutting rather than a single rule: the in-cluster `kubernetes` Service maps 443→targetPort 6443 and Cilium evaluates policy after kube-proxy-replacement DNAT, so *every* policy that talks to the apiserver must allow 6443 or all API calls hang.

## 🧹 Checklist

- [ ] Unit tests added/updated — n/a, chart-template change; covered by `helm lint`/`helm template` and verified against the rendered ESO chart


aweingarten:
Addressed the one naming point from review: the policy is now `external-secrets-platform-policy`, matching `git-server-platform-policy` / `gitea-platform-policy` / `otomi-api-platform-policy` — it was the only one in the chart not following that pattern.

No other change. Flagging the one thing a reviewer should push back on if they disagree: the apiserver ingress rule on 10250 is scoped by port with no `from`, because on a managed control plane (LKE / LKE-E) the apiserver is off the pod network and has no address a `namespaceSelector` or `ipBlock` can pin down. That is deliberate and commented in place, but it is the loosest rule in the file.

merll (approved):

PR #3499, opened 2026-08-07
chore(chart-deps): update argocd to version 10.3.0

svcAPLBot:
This PR updates the dependency **argo-cd** to version **10.3.0**.


merll (changes requested):
Blocking until cause of reported issue (Operation State unknown in list view) is known.

PR #3525, opened 2026-08-11
chore: upgrade tools and tty binaries

merll:
## 🧹 Checklist


tools/Dockerfile
```diff
@@ -4,29 +4,29 @@ FROM ubuntu:26.04 AS builder
...
+ARG KUBECTL_VERSION=1.36.3
 # https://github.com/helm/helm/tags
-ARG HELM_VERSION=3.20.2
+ARG HELM_VERSION=4.2.3
 # https://github.com/databus23/helm-diff/releases
-ARG HELM_DIFF_VERSION=3.15.6
+ARG HELM_DIFF_VERSION=3.15.11
 # https://github.com/jkroepke/helm-secrets/releases
-ARG HELM_SECRETS_VERSION=4.7.6
+ARG HELM_SECRETS_VERSION=4.7.7
```
CasLubbers, line 13: Should we remove helm secrets? With the removal of sops this should not be needed anymore.

tools/Dockerfile
```diff
@@ -4,29 +4,29 @@ FROM ubuntu:26.04 AS builder
...
 # https://github.com/mozilla/sops/releases
 ARG SOPS_VERSION=3.12.2
```
CasLubbers, line 15: We can remove SOPS and AGE right?

tools/Dockerfile
```diff
@@ -4,29 +4,29 @@ FROM ubuntu:26.04 AS builder
...
 # https://github.com/FiloSottile/age/releases
 ARG AGE_VERSION=1.2.1
 # https://github.com/noqcks/gucci/releases
 ARG GUCCI_VERSION=1.9.0
```
CasLubbers, line 19: Do we still use gucci? That was something used for the sops secrets, but do not know what exactly

CasLubbers (changes requested):

tools/Dockerfile
merll: Unfortunately yes, in two places. Bootstrap still calls it, initializing the secret values. We should replace this with a `Password` resource from ESO as soon as possible. And the migrations call it, but under a condition which we never make use of. So we should remove it soon, but cannot do it in this PR.

CasLubbers (approved):
discussed that we will handle the deletion of binaries with a separate ticket.

PR #3542, opened 2026-08-14
docs: add ADRs for Dex as OIDC issuer and Keycloak migration path

CasLubbers:
## 📌 Summary

PRD's created from the ADR's:
- https://github.com/linode/apl-core/issues/3534
- https://github.com/linode/apl-core/issues/3535
- https://github.com/linode/apl-core/issues/3536
- https://github.com/linode/apl-core/issues/3537
- https://github.com/linode/apl-core/issues/3538
- https://github.com/linode/apl-core/issues/3539
- https://github.com/linode/apl-core/issues/3540
- https://github.com/linode/apl-core/issues/3541
## 🧹 Checklist


adr/2026-08-06-dex-as-issuer.md
```diff
@@ -0,0 +1,129 @@
...
+
+So configuration it is. The consequence is that adding a user, adding a team, or changing group membership requires Dex to restart, because Dex reads its configuration only at startup and never reloads it.
+
+That restart is made harmless rather than avoided: Dex runs two replicas with `maxUnavailable: 0`, and a checksum of the rendered configuration is annotated on the Deployment so a change rolls the pods automatically. Sessions live in Dex storage, not pod memory, so a roll does not log anybody out.
+
+We have asked upstream for group membership to be accepted over the gRPC API (dexidp/dex#4972). If that is granted, users could be managed through the API instead, which would remove the restart. Nothing else in this decision depends on it.
+
+### How users reach the configuration
+
+Users are created through the API and stored as SealedSecrets in the `apl-users` namespace. Under Keycloak they reach the issuer at runtime, because the Keycloak operator watches those Secrets. Dex has no operator, so the records must be in hand when the configuration is rendered.
```
merll: In Keycloak, various problems arose from this approach, too. The operator was both watching Secrets and a common ConfigMap, and had several issues synchronizing data correctly. We eventually decided for a plan to abandon the operator and instead have the API act as a direct proxy into Keycloak, but due to the perspective of deprecating Keycloak this was never followed up on.

adr/2026-08-06-dex-as-issuer.md
```diff
@@ -0,0 +1,129 @@
...
+A Dex password entry carries an explicit `userID`, and Dex derives the subject as `base64url(proto{user_id, connector_id})`. Setting `userID` to the APL user's existing UUID makes subjects repository-derived rather than Dex-allocated: they survive a cluster rebuild, and any user's subject is computable offline before a cluster is touched.
+
+The connector ID is part of every subject and must never be renamed.
+
+### Two user populations
+
+Bringing an external identity provider remains supported, and the two populations stay mutually exclusive, as they are under Keycloak:
+
+- **Users created in APL** originate in the values repository and are rendered into Dex's `staticPasswords` from their materialised form on the cluster, as described above. The values repository remains their system of record.
+- **Federated users** arrive through a Dex `oidc` connector configured from the existing `oidc.*` settings. They are never written to the values repository, and their team membership comes from group claims.
```
merll: Will this also be the migration path for Keycloak? Adding Keycloak as an OIDC connector but excluding it from direct APL management would leave an option for legacy installations to keep the user base, and migrate at their own choice (if ever). This would require them to exclusively create and update users through the Keycloak UI from that point on, but that would be acceptable.

adr/2026-08-06-dex-as-issuer.md
```diff
@@ -0,0 +1,129 @@
...
+
+A Dex password entry carries an explicit `userID`, and Dex derives the subject as `base64url(proto{user_id, connector_id})`. Setting `userID` to the APL user's existing UUID makes subjects repository-derived rather than Dex-allocated: they survive a cluster rebuild, and any user's subject is computable offline before a cluster is touched.
+
+The connector ID is part of every subject and must never be renamed.
+
+### Two user populations
+
+Bringing an external identity provider remains supported, and the two populations stay mutually exclusive, as they are under Keycloak:
+
+- **Users created in APL** originate in the values repository and are rendered into Dex's `staticPasswords` from their materialised form on the cluster, as described above. The values repository remains their system of record.
```
merll: Why is the only considered approach to render the configuration?

There were ideas before to not use apl-operator for purposes anymore, where it has to deal with mixes of resources and states of the cluster. Teams are somewhere on the border of this, but too many templates need the information which teams exist. With users however, reading resources, processing them, and writing them back to another resource seems quite complicated.
What if API were to read and write the configuration file resource directly? It would require a separate backup. Are there other reasons this was not considered?

adr/2026-08-06-dex-as-issuer.md
```diff
@@ -0,0 +1,129 @@
...
+- **Runs a pre-release Dex.** Auth sessions and RP-initiated logout are merged on Dex's master branch but unreleased (dexidp/dex#4560). Without sessions there is no single sign-on across applications that run their own OIDC flow, and no logout endpoint. The image is therefore pinned to a master digest with `DEX_SESSIONS_ENABLED=true`. **APL must not be released while this pin stands.**
+- **Single sign-on needs explicit configuration.** `sessions.ssoSharedWithDefault` defaults to `none`, which gives no SSO at all. It is set to `all`.
+- **Every claim must be requested.** Keycloak attached `groups` through a protocol mapper regardless of scope; Dex only emits a claim when its scope is requested (`server/tokens/issuer.go`). Consumers that previously asked for `openid` alone now need the full scope list, or they authenticate successfully with no authorization.
+- **Logout requires the ID token.** Dex refuses `post_logout_redirect_uri` without `id_token_hint`, and a browser redirect cannot carry one. Logout is therefore driven from oauth2-proxy's `backend_logout_url`, which substitutes `{id_token}`.
```
merll: Is that a problem, or the solution (i.e. OAuth2-Proxy)? To be fair this parameter is also recommended in Keycloak, but optional.

adr/2026-08-06-dex-as-issuer.md
merll: I read the options below, what I meant is a fourth option, somewhat a mix of option 2 and 3: Making Dex the OIDC provider on all installations, but on migrated ones leave Keycloak behind it. Cut off apl-keycloak-operator, creating new users through API exclusively in Dex.

adr/2026-08-06-dex-as-issuer.md
CasLubbers: The problem now is that we cannot fully use the Dex API for users as it is missing the Group attribute. That one only works in the configuration file. So the our apl-api cannot suffice as an proxy for Dex. 
But it is good to keep in mind for future reference that we do not want to create a dex-operator.

adr/2026-08-06-dex-as-issuer.md
CasLubbers: Yes that sounds as a good option. The only problem I have seen so far is that if we switch a Keycloak cluster to Dex as issuer. The users will lose their Harbor and Gitea account. 
The subject that Dex returns is different. Gitea and Harbor will therefore create new accounts automatically. Maybe further down the line we can fix this or have a better solution for it.

adr/2026-08-06-dex-as-issuer.md
CasLubbers: TBH I did not think of it. Could be a valid option. The apl-operator seemed the easiest one to me because it has the values structure and needs the configuration file on start up. So with some extra templating it can create users.

adr/2026-08-06-dex-as-issuer.md
CasLubbers: A solution, did not test the logout option in the PoC

adr/2026-08-06-dex-as-issuer.md
merll: Especially in Harbor I doubt individual users have significant amounts of data attached to their personal account. But in that case there is indeed little use in adding Dex as a layer on migrated installations.

adr/2026-08-06-dex-as-issuer.md
merll: What I am aiming at is that I would rather avoid adding more complexity to the operator. Considering our discussions of possibly migrating code, actually simplifying would help in the long run.

I do see the advantage of keeping everything in the values repository however. One similar idea: The API could write out the users as SealedSecret resources, and External Secrets Operator could select the resulting Secrets to produce the configuration file. That is all components we already have.

merll (approved):

PR #3570, opened 2026-08-21
feat(chart): align chart/ and charts/apl-operator

CasLubbers:
## 🧹 Checklist


charts/apl-operator/templates/deployment.yaml
```diff
@@ -26,19 +34,26 @@ spec:
       labels:
         {{- include "apl-operator.selectorLabels" . | nindent 8 }}
     spec:
-      {{- with .Values.imagePullSecrets }}
+      # Appended to the cluster resolvers, as a fallback while CoreDNS starts.
+      dnsConfig:
+        nameservers:
+          - 8.8.8.8
+          - 8.8.4.4
```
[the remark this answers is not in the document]
merll: Good suggestion, but since it is a new feature and should be consistently in the platform I would consider it out of scope here.

charts/apl-operator/templates/deployment.yaml
```diff
@@ -49,8 +64,12 @@ spec:
...
           {{- end }}
           env:
-            - name: CI
-              value: "true"
+            - name: VERBOSITY
+              value: '1'
+            - name: ENV_DIR
+              value: /home/app/stack/env
+            - name: VALUES_INPUT
+              value: /secret/values.yaml
```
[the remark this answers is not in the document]
merll: Not relevant, because we do not support a SOPS recovery, which would be a legacy version.

merll (approved):

chart/apl/templates/deployment.yaml
j-zimnowoda, line 1: Will that symlink resolve during the helm chart packaging process? In other words after releasing that chart can user helm install without any problems ?

CasLubbers: Will double check it today

CasLubbers: Validated it: helm lint and helm package both dereference it

PR #3592, opened 2026-08-25
refactor: replace generated platform secrets with eso generator

merll:
## 📌 Summary

This PR intends to fix an issue that sometimes occurs during installations, causing delays of several minutes. Likely this was due to the operator restarting due to an error condition, and stalling on the attempt to re-read generated secrets.

The use of generated secrets has been refactored so that External Secrets Operator is used for creating random passwords that are internal to the platform. The generated Secret resources are immutable (but could be cycled by just deleting them). The migration makes sure to copy the already-generated secrets into the new structure, for avoiding having to restart processes. Afterwards, SealedSecret resources are removed from the values repository.

Other (non-generated) 'x-secret' values from the values schema are kept for backwards compatibility with existing values input on installation.

There is one additional generated password (Valkey for Gitea). This causes Gitea to restart after a migration.

Gucci is no longer required and has therefore been removed from build files.

## 🧹 Checklist


src/cmd/migrate.ts
```diff
@@ -808,15 +745,264 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
...
+        api,
+        'otomi',
+        SEALED_SECRETS_NAMESPACE,
+        {
+          adminPassword: secrets.otomi.adminPassword,
+        },
+        false,
+        true,
+      )
+    }
```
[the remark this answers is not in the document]
merll: This has been split now between a generated secret `otomi`, but keeping the existing `otomi-secrets` where `globalPullSecret_password` is present.

src/cmd/migrate.ts
```diff
@@ -808,15 +745,264 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
   }
 }
 
+const migrateGeneratedSecrets = async (values: Record<string, any>) => {
+  const d = terminal('migrateGeneratedSecrets')
+  const BASE_APPS = ['argocd', 'keycloak', 'oauth2-proxy', 'oauth2-proxy-redis', 'otomi']
+  const OPTIONAL_APPS = ['gitea', 'harbor', 'loki', 'kubeflow-pipelines']
+  const ALL_APPS = [...BASE_APPS, ...OPTIONAL_APPS]
```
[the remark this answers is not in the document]
merll: We have used these functions before, and they can only usefully be tested on a cluster actually performing the migration. Unit tests on the other hand are not supposed to actually contact the cluster, and confirming our own assumptions in mock calls is not useful in this case.

src/cmd/migrate.ts
```diff
@@ -808,15 +745,264 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
...
+        false,
+        true,
+        'kubernetes.io/basic-auth',
+      )
+      await createUpdateGenericSecret(
+        api,
+        'gitea',
+        SEALED_SECRETS_NAMESPACE,
+        {
+          adminUsername: 'otomi-admin',
```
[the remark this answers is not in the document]
merll: Replaced with values

src/cmd/migrate.ts
```diff
@@ -808,15 +745,264 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
...
+        },
+        false,
+        true,
+      )
+      await createUpdateGenericSecret(
+        api,
+        'harbor-secret-key',
+        'harbor',
+        {
+          secretKey: generate({ ...GENERATE_OPTS, length: 16 }),
```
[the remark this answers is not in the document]
merll: This value should already exist. This has been replaced.

src/cmd/migrate.ts
```diff
@@ -808,15 +745,264 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
...
+        },
+        false,
+        true,
+      )
+      await createUpdateGenericSecret(
+        api,
+        'harbor-core-xsrf-secret',
+        'harbor',
+        {
+          CSRF_KEY: generate({ ...GENERATE_OPTS }),
```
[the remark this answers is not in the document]
merll: This value should already exist. This has been replaced.

src/cmd/migrate.ts
```diff
@@ -808,15 +745,264 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
...
+        },
+        false,
+        true,
+      )
+      await createUpdateGenericSecret(
+        api,
+        'harbor-registry-http',
+        'harbor',
+        {
+          REGISTRY_HTTP_SECRET: generate({ ...GENERATE_OPTS, length: 16 }),
```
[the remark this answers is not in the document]
merll: This value should already exist. This has been replaced.

src/cmd/migrate.ts
```diff
@@ -808,15 +745,264 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
...
+        },
+        false,
+        true,
+      )
+      await createUpdateGenericSecret(
+        api,
+        'harbor-core-secret',
+        'harbor',
+        {
+          secret: generate({ ...GENERATE_OPTS, length: 16 }),
```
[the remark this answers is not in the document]
merll: This value should already exist. This has been replaced.

src/cmd/migrate.ts
```diff
@@ -858,8 +972,7 @@ export const applyChanges = async (
         const prev = get(values, path)
         if (prev !== undefined) {
           // path worked and we found something, simple scenario, just replace directly
-          const ret = await replace(tmplStr, prev)
-          set(values, path, ret)
+          set(values, path, tmplStr)
         } else {
```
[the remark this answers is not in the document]
merll: An upgrade this far back is unlikely to work and not supported. Removed upgrades prior to 6.x to make this clear.

src/cmd/migrate.ts
```diff
@@ -808,15 +745,192 @@ const removeIngressNginxValues = async (values: Record<string, any>) => {
...
+          username: 'gitea',
+          password: secrets.gitea.postgresqlPassword,
+        },
+        'kubernetes.io/basic-auth',
+      )
+      await setSecret('gitea', SEALED_SECRETS_NAMESPACE, {
+        adminUsername: values?.apps?.gitea?.adminUsername || 'otomi-admin',
+        adminPassword: secrets.gitea.adminPassword,
+        valkeyPassword: generate(GENERATE_OPTS),
+      })
```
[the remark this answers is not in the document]
merll: I guess that was just missing in the test fixtures. The `adminPassword` was a value generated from the values schema before and therefore should exist invariably.

charts/external-secrets/crds/clustergenerator.yaml
j-zimnowoda, line 1: Can this CRD be added in separate PR?

values/raw/generated-secrets.gotmpl
```diff
@@ -0,0 +1,142 @@
...
+            generatorRef:
+              apiVersion: generators.external-secrets.io/v1alpha1
+              kind: ClusterGenerator
+              name: alpha-num-32
+          rewrite:
+            - regexp:
+                source: password
+                target: idpClientSecret
+{{- if $v.apps.gitea.enabled }}
+  - apiVersion: external-secrets.io/v1
```
j-zimnowoda, line 64: what is the reason behind not including item the  gitea-artifacts release?

values/raw/generated-secrets.gotmpl
```diff
@@ -0,0 +1,142 @@
+{{- $v := .Values }}
+
+resources:
```
j-zimnowoda, line 3: why these secrets cannot be added as app release artifacts ?

values-changes.yaml
```diff
@@ -507,3 +42,12 @@ changes:
   - version: 71
     customFunctions:
       - preservePvcStorageClassInRawValues
+  - version: 72
+    customFunctions:
+      - migrateGeneratedSecrets
+    deletions:
```
j-zimnowoda, line 48: These removed schema properties are not listed here:

apps.argocd.redisPassword
apps.gitea.adminPassword
apps.gitea.postgresqlPassword
apps.harbor.adminPassword
apps.keycloak.idp.clientSecret
apps.kubeflow-pipelines.rootPassword
apps.loki.adminPassword
apps.oauth2-proxy-redis.password
otomi.adminPassword

Was it intended ?

j-zimnowoda (dismissed):
The minimal install test case scenario fails while deploying `argocd-redis`: 

```
  Warning  Failed     1s (x5 over 25s)  kubelet            spec.containers{redis}: Error: secret "argocd-redis" not found
```

charts/external-secrets/crds/clustergenerator.yaml
merll: Yes, I can split it from this PR, including the chart upgrade script change.

values-changes.yaml
merll: This would have no effect, since the values are actually no longer stored there. They are already dropped at the time of the SOPS migration.

values/raw/generated-secrets.gotmpl
merll: These are cases where the generated Secret values should be kept in the ClusterStore, and therefore in the `apl-secrets` namespace. This can be either
* because they are shared between applications
* or because they cannot be used in one step (see Gitea above) because we set up applications this way.

values/raw/generated-secrets.gotmpl
merll: The Gitea config is shared between multiple Helmfile releases. For example the valkeyPassword is used as an env value in `gitea-valkey`, and rendered into a URL in `gitea`.

charts/external-secrets/crds/clustergenerator.yaml
merll: https://github.com/linode/apl-core/pull/3628

helmfile.d/helmfile-01.init.yaml.gotmpl
```diff
@@ -42,6 +42,14 @@ releases:
     installed: {{ $a | get "sealed-secrets.enabled" }}
     namespace: external-secrets
     <<: *raw
+  - name: generated-secrets
```
CasLubbers, line 45: Depending on which one gets merged first. I have a similar artifact in the Dex PR's

CasLubbers (approved):
Fresh install works. All apps tested expect for KFP and RabbitMQ

j-zimnowoda (approved):

PR #3604, opened 2026-08-29
chore(chart-deps): update external-secrets to version 2.10.0

svcAPLBot:
This PR updates the dependency **external-secrets** to version **2.10.0**.


CasLubbers (approved):
Tested on cluster and eso is shown as an app

PR #3632, opened 2026-09-09
feat: add loki network policy

merll:
## 🧹 Checklist


CasLubbers (approved):
Confirmed netpols are on cluster and Loki still works

